PR #5, opened 2024-03-18
Feat: Separate blocks by short indicator lines on the left of a block.

redur:
**Implementation of left line splitting of blocks**
This PR contains the code and parameter to split blocks based on small indicator lines on the lefthandside of the blocks. These lines are often used when space is tight, and come in conjunction with what we call 'splitting due to vertical distances'.
Overall, **F1 improved from 86.3% to 86.7%.** The changes are mainly driven due to a few documents where larger improvements could be made.

**Document level metric changes:**
The below table contains all documents that have a change in their F1 metric. 'left_splitting' indicates the new metrics, obtained with this PR:

There are mostly improvements, but a few cases come with a lowering in the score. Sometimes it's not so clear why (typically documents that have a low score anyways). Sometimes the worsening is due to a non recognized line split. This happens because we lowered `block_line_ratio`. It was needed to lower this parameter, as otherwise some left line splits would be recognized as complete lines. It matters, as we still look for splitting due to vertical spacing in the case of left line splits, which we don't do for complete horizontal lines. We can rethink this heuristic together.

|    | document_name    |   F1_left_splitting |   precision_left_splitting |   recall_left_splitting |   Number Elements_left_splitting |   Number wrong elements_left_splitting |        F1 |   precision |   recall |   Number Elements |   Number wrong elements |
|---:|:-----------------|--------------------:|---------------------------:|------------------------:|---------------------------------:|---------------------------------------:|----------:|------------:|---------:|------------------:|------------------------:|
|  0 | 680244005-bp.pdf |           0.285714  |                  0.315789  |                0.26087  |                               46 |                                     34 | 0.309524  |   0.342105  | 0.282609 |                46 |                      33 |
|  1 | 268124635-bp.pdf |           0.766667  |                  0.766667  |                0.766667 |                               30 |                                      7 | 0.833333  |   0.833333  | 0.833333 |                30 |                       5 |
|  2 | 267123080-bp.pdf |           0.888889  |                  0.857143  |                0.923077 |                               13 |                                      1 | 0.814815  |   0.785714  | 0.846154 |                13 |                       2 |
|  3 | 268124569-bp.pdf |           1         |                  1         |                1        |                               64 |                                      0 | 0.79646   |   0.918367  | 0.703125 |                64 |                      19 |
|  4 | 699248001-bp.pdf |           0.767857  |                  0.843137  |                0.704918 |                               61 |                                     18 | 0.621359  |   0.761905  | 0.52459  |                61 |                      29 |
|  5 | 268125447-bp.pdf |           0.363636  |                  0.4       |                0.333333 |                               18 |                                     12 | 0.424242  |   0.466667  | 0.388889 |                18 |                      11 |
|  6 | 268124125-bp.pdf |           0.0263158 |                  0.0178571 |                0.05     |                               20 |                                     19 | 0.0273973 |   0.0188679 | 0.05     |                20 |                      19 |
|  7 | 691251009-bp.pdf |           0.25      |                  0.333333  |                0.2      |                               20 |                                     16 | 0.266667  |   0.4       | 0.2      |                20 |                      16 |
|  8 | 268124307-bp.pdf |           0.862745  |                  0.846154  |                0.88     |                               25 |                                      3 | 0.923077  |   0.888889  | 0.96     |                25 |                       1 |
|  9 | 677250019-bp.pdf |           0.823529  |                  0.777778  |                0.875    |                                8 |                                      1 | 0.588235  |   0.555556  | 0.625    |                 8 |                       3 |
| 10 | 267123060-bp.pdf |           0.941176  |                  0.888889  |                1        |                                8 |                                      0 | 0.75      |   0.75      | 0.75     |                 8 |                       2 |
| 11 | 680244002-bp.pdf |           0.44      |                  0.458333  |                0.423077 |                               26 |                                     15 | 0.4       |   0.416667  | 0.384615 |                26 |                      16 |
| 12 | 268124599-bp.pdf |           0.5       |                  0.4       |                0.666667 |                                3 |                                      1 | 0.444444  |   0.333333  | 0.666667 |                 3 |                       1 |
| 13 | 267124070-bp.pdf |           0.816327  |                  0.869565  |                0.769231 |                               26 |                                      6 | 0.857143  |   0.913043  | 0.807692 |                26 |                       5 |

**Examples:** (Left before, right after)
267123060-bp.pdf
image
--> Dashed line recognised as a left line splitting.

677250019-bp.pdf
image
--> All splits due to lines are now correct. We expect one block too much and that's why the last block is split into two.

268124307-bp.pdf
image
Line not detected any more. This is due to lowering the `block_line_ratio` parameter. Worsening of result in this case.

699248001-bp.pdf
image
image
image
--> Overall, good improvement (+14%F1). But still untapped potential left. Not all blocks are split due to the lefthand line segments. This is mainly due to the fact, that the lines do not "cross" the block edge. And if they do, they're not horizontal, and thus cross the block "too high". --> For the future: explore the splitting condition and try to cover this case as well.

268124569-bp.pdf
image
image
--> perfect example, with the left line splitting we reach 100% recognition rate.

268124635-bp.pdf
image
--> not so clear what happens here. F1 drops. Seems that the lower `block_line_ratio` lead to not recognizing a line split.

### DIFF
--- a/config/matching_params.yml
+++ b/config/matching_params.yml
@@ -1,2 +1,3 @@
 # TODO: Any thoughts on a better name?
-block_line_ratio: 0.13
+block_line_ratio: 0.19
+left_line_length_threshold: 7
--- a/src/stratigraphy/util/find_description.py
+++ b/src/stratigraphy/util/find_description.py
@@ -21,8 +21,22 @@ def get_description_blocks(
     description_lines: list[TextLine],
     geometric_lines: list[Line],
     block_line_ratio: float,
+    left_line_length_threshold: float,
     target_layer_count: int = None,
 ) -> list[TextBlock]:
+    """Group the description lines into blocks based on the presence of geometric lines, the indentation of lines
+    and the vertical spacing between lines.
+
+    Args:
+        description_lines (list[TextLine]): The text lines to group into blocks.
+        geometric_lines (list[Line]): The geometric lines detected in the pdf page.
+        block_line_ratio (float): The relative part a line has to cross a block in order to induce a splitting.
+        left_line_length_threshold (float): The minimum length of a line segment on the left side of a block to split it.
+        target_layer_count (int, optional): Expected number of blocks. Defaults to None.
+
+    Returns:
+        list[TextBlock]: A list of blocks containing the description lines.
+    """
     distances = []
     for line_index in range(len(description_lines) - 1):
         line1rect = description_lines[line_index].rect
@@ -44,7 +58,17 @@ def get_description_blocks(
             ):
                 blocks.append(TextBlock(current_block_lines, is_terminated_by_line=True))
                 current_block_lines = []
-
+            elif _block_separated_by_lefthandside_line_segment(
+                last_line,
+                line,
+                TextBlock(current_block_lines),
+                geometric_lines,
+                length_threshold=left_line_length_threshold,
+            ):
+                blocks.append(
+                    TextBlock(current_block_lines, is_terminated_by_line=False)
+                )  # we consider a splitting due to a lefthandside line segment as weak.
+                current_block_lines = []
         current_block_lines.append(line)
     if len(current_block_lines):
         blocks.append(TextBlock(current_block_lines))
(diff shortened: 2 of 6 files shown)


config/matching_params.yml
```diff
@@ -1,2 +1,3 @@
 # TODO: Any thoughts on a better name?
-block_line_ratio: 0.13
\ No newline at end of file
+block_line_ratio: 0.19
```
redur, line 2: Result from some parameter tuning.

src/stratigraphy/util/find_description.py
```diff
@@ -44,7 +58,17 @@ def get_description_blocks(
             ):
                 blocks.append(TextBlock(current_block_lines, is_terminated_by_line=True))
                 current_block_lines = []
-
+            elif _block_separated_by_lefthandside_line_segment(
```
redur, line 61: The check for both, lines as well as the lefthandside lines, leads to some trouble in the recognition.

Sometimes, a block is split by a lefthandside line. The resulting block consists of a single textline, which may be short. Therefore, the new block is small in its width. The next lefthandside line may now be detected as a "full line". The block is split correctly, but then later, we do not apply the splitting due to vertical spacing.

This is a solvable problem. We could for example loop through the blocks twice.

src/stratigraphy/util/find_description.py
```diff
@@ -59,34 +83,73 @@ def get_description_blocks(
...
-
-            current_block_lines.append(line)
-        if len(current_block_lines):
-            blocks.append(TextBlock(current_block_lines))
-
+        # This case means that there are fewer blocks than the minimum number of blocks we expect.
+        # In this case we redo all the blocks from scratch.
+        blocks = _split_block_by_vertical_spacing(description_lines, threshold=threshold)
+
+    elif count_blocks_divided_by_line < min_block_count:
```
redur, line 92: That just means, if the blocks are due to lefthandside segments, then still apply vertical distance splitting. But keep the previously made blocks.

src/stratigraphy/util/find_description.py
```diff
@@ -102,3 +165,35 @@ def _block_separated_by_line(
         if is_line_long_enough and line_ends_block:
             return True
     return False
+
+
+def _block_separated_by_lefthandside_line_segment(
```
redur, line 170: I'd like to think here on how to split blocks as well where the lefthandside line does not cross the block. 

Example here:
image

stijnvermeeren-swisstopo (approved):
LGTM!

PR #6, opened 2024-03-19
Feat/indicator lines v2

redur:
**Description**
Consider separation of blocks due to lines separate from block separation due to indicator lines on the left.
Included more sophisticated logic for block separation due to indicator lines.

Overall performance in F1 drops from 86.7% to 86.4%.

**Document Level Metrics Change**
|---:|:-----------------|--------------------:|---------------------------------:|---------------------------------------:|---------:|------------------:|------------------------:|
|  0 | 680244005-bp.pdf |            0.309524 |                               46 |                                     33 | 0.285714 |                46 |                      34 |
|  1 | 268124635-bp.pdf |            0.833333 |                               30 |                                      5 | 0.766667 |                30 |                       7 |
|  2 | 681246002-bp.pdf |            0.857143 |                               10 |                                      1 | 1        |                10 |                       0 |
|  3 | 699248001-bp.pdf |            0.883333 |                               61 |                                      8 | 0.767857 |                61 |                      18 |
|  4 | 677259003-bp.pdf |            0.714286 |                                7 |                                      2 | 0.857143 |                 7 |                       1 |
|  5 | 691251009-bp.pdf |            0.228571 |                               20 |                                     16 | 0.25     |                20 |                      16 |
|  6 | 268124307-bp.pdf |            0.923077 |                               25 |                                      1 | 0.862745 |                25 |                       3 |
|  7 | 680244002-bp.pdf |            0.36     |                               26 |                                     17 | 0.44     |                26 |                      15 |
|  8 | 268124599-bp.pdf |            0.2      |                                3 |                                      2 | 0.5      |                 3 |                       1 |

**Examples:**
left before | right after

268124635-bp.pdf
image
Result improve. Two blocks correctly identified that were not before.

681246002-bp.pdf
image
Result decreases. One block is falsely split.

699248001-bp.pdf
image
image
image
Big improvement. From previously 18 wrong recognitions we went down to 8.

677259003-bp.pdf
image
Decrease in performance. One Block is falsely split.

268124307-bp.pdf
image
Performance improvment. We gain back what we lost due to increasing block_line_ratio.

680244002-bp.pdf
image
Performance decrease. Two additional blocks are not recognized in this case. Not so clear why and how.

268124599-bp.pdf	
image
One more block not recognized. Seems blocks are split too often in this case.

**Conclusion**
Overall, 10 mistakes less are made. However, the F1 score (as well as precision and recall) decreased. This is due to the fact, that for most documents the metrics decreased slightly, and for some documents with many blocks, we saw larger improvements. For me, fewer manual annotation are needed using the proposed changes herein, and that's why I deem them better. --> maybe we have to rethink our metrics.

### DIFF
--- a/src/stratigraphy/util/find_description.py
+++ b/src/stratigraphy/util/find_description.py
@@ -220,12 +220,14 @@ def _block_separated_by_lefthandside_line_segment(
 
         line_ends_block = last_line_y_coordinate < line_y_coordinate and line_y_coordinate < current_line_y_coordinate
 
-        alternative_condition = (
+        weak_condition = (
             (line.start.x - 5 < block.rect.x0 and line.end.x > block.rect.x0)
             and (np.abs(line.start.x - line.end.x) > length_threshold - 2)
             and (len(block.lines) > 1)
         )  # if block has at least three lines, we weaken the splitting condition.
+        # It is three lines because the statement means that block.lines has at least two elements.
+        # The third line is current_line
 
-        if line_ends_block and ((is_line_long_enough and line_cuts_lefthandside_of_block) or alternative_condition):
+        if line_ends_block and ((is_line_long_enough and line_cuts_lefthandside_of_block) or weak_condition):
             return True
     return False
--- a/src/stratigraphy/util/find_description.py
+++ b/src/stratigraphy/util/find_description.py
@@ -1,3 +1,5 @@
+from typing import Callable
+
 import fitz
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


src/stratigraphy/util/find_description.py
```diff
@@ -48,30 +50,30 @@ def get_description_blocks(
...
-                length_threshold=left_line_length_threshold,
-            ):
-                blocks.append(
-                    TextBlock(current_block_lines, is_terminated_by_line=False)
-                )  # we consider a splitting due to a lefthandside line segment as weak.
-                current_block_lines = []
-        current_block_lines.append(line)
-    if len(current_block_lines):
-        blocks.append(TextBlock(current_block_lines))
+                _block_separated_by_lefthandside_line_segment,
```
stijnvermeeren-swisstopo, line 69: Maybe pass this as e.g.
`lambda last_line, line, block: _block_separated_by_lefthandside_line_segment(last_line, line, block, geometric_lines, left_line_length_threshold)`? Then the method `_create_blocks_by_separator` itself can do with two fewer parameters.

src/stratigraphy/util/find_description.py
```diff
@@ -194,6 +219,15 @@ def _block_separated_by_lefthandside_line_segment(
         )  # for the block splitting, we only care about x-extension
 
         line_ends_block = last_line_y_coordinate < line_y_coordinate and line_y_coordinate < current_line_y_coordinate
-        if is_line_long_enough and line_ends_block and line_cuts_lefthandside_of_block:
+
+        weak_condition = (
+            (line.start.x - 5 < block.rect.x0 and line.end.x > block.rect.x0)
+            and (np.abs(line.start.x - line.end.x) > length_threshold - 2)
+            and (len(block.lines) > 1)
+        )  # if block has at least three lines, we weaken the splitting condition.
```
stijnvermeeren-swisstopo, line 227: Could you elaborate on the intuition behind this weaker condition? Why does it make sense to weaken the splitting condition if the block has at least three lines?

stijnvermeeren-swisstopo (changes requested):

src/stratigraphy/util/find_description.py
redur: Looking at the pdf's, and the errors we made, I have realised that it seldomly happens to have more than three lines in one block. If blocks are larger, then, more often than not, we missed a split. That lead me to the decision to experiment with weaker conditions for large blocks only.

Afterwards, I checked all the documents that changed due to this weaker condition, and the improvements are dominant. Especially in some cases there is a large improvement (i.e. many more cases are recognized) whereas a few documents do not benefit from this change.

src/stratigraphy/util/find_description.py
redur: I personally find the code clearer to read and understand the way it is currently. I agree that we'd need to pass fewer arguments in your case, which comes with the benefit of added usage.

If we are not to use this function very frequently in other places too, I would keep it. But happy to implement otherwise if you prefer.

stijnvermeeren-swisstopo (approved):
Happy to merge this now. I think I will still revisit this code a little bit afterwards, to see if it can be cleaned up more, but I'll do that after the linter has been merged as well.

PR #7, opened 2024-03-19
Use material_description_block to determine if line is long enough fo…

redur:
…r a split.

**Description**
Used material_description_rec to determine if a line is long enough to consider it as a split condition for a block.
While doing so, I once again optimized for the `block_line_ratio` parameter.

F1 increases from 86.4% - 86.6%.

**Document Level Changes**
The changes (almost) exclusively improve the scores on all documents.

|    | document_name    |   F1_new |   Number Elements_new |   Number wrong elements_new |        F1 |   Number Elements |   Number wrong elements |
|---:|:-----------------|---------:|----------------------:|----------------------------:|----------:|------------------:|------------------------:|
|  0 | 680244005-bp.pdf | 0.380952 |                    46 |                          30 | 0.309524  |                46 |                      33 |
|  1 | 677259003-bp.pdf | 0.857143 |                     7 |                           1 | 0.714286  |                 7 |                       2 |
|  2 | 268124125-bp.pdf | 0.025    |                    20 |                          19 | 0.0263158 |                20 |                      19 |
|  3 | 680244002-bp.pdf | 0.44     |                    26 |                          15 | 0.36      |                26 |                      17 |
|  4 | 268124391-bp.pdf | 0.901961 |                    23 |                           0 | 0.92      |                23 |                       0 |

Visual inspection show a (very small) qualitative improvement.

### DIFF
--- a/src/stratigraphy/util/find_description.py
+++ b/src/stratigraphy/util/find_description.py
@@ -22,6 +22,7 @@ def get_description_lines(lines: list[TextLine], material_description_rect: fitz
 def get_description_blocks(
     description_lines: list[TextLine],
     geometric_lines: list[Line],
+    material_description_rect: fitz.Rect,
     block_line_ratio: float,
     left_line_length_threshold: float,
     target_layer_count: int = None,
@@ -55,7 +56,7 @@ def get_description_blocks(
         description_lines,
         geometric_lines,
         _block_separated_by_line,
-        {"threshold": block_line_ratio},
+        {"threshold": block_line_ratio, "material_description_rect": material_description_rect},
         set_terminated_by_line_flag=True,
     )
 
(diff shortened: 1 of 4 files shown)


src/stratigraphy/util/find_description.py
```diff
@@ -183,7 +189,8 @@ def _block_separated_by_line(
         line_y_coordinate = (line.start.y + line.end.y) / 2
 
         is_line_long_enough = (
-            np.min([block.rect.x1, line_right_x]) - np.max([block.rect.x0, line_left_x]) > threshold * block.rect.width
+            np.min([material_description_rect.x1, line_right_x]) - np.max([material_description_rect.x0, line_left_x])
+            > threshold * block.rect.width
```
stijnvermeeren-swisstopo, line 193: Don't we rather want to compare with `threshold * material_description_rect.width` here? That would also make the `block` parameter to this method redundant.

stijnvermeeren-swisstopo (changes requested):

src/stratigraphy/util/find_description.py
redur: Yep, you are right!

redur: Your suggested change increased F1 to 86.7%. Thank you!

stijnvermeeren-swisstopo (approved):

PR #8, opened 2024-03-19
Refactor/ruff lint

redur:
Introduction of Ruff. A very fast code formatter and code linter. Settings are compatible with previous isort and black settings. But now, more rules are added.

The PR comes with refactoring of the code to be ruff compliant.

Code runs, metrics are not affected.

### DIFF
--- a/src/stratigraphy/benchmark/score.py
+++ b/src/stratigraphy/benchmark/score.py
@@ -1,10 +1,10 @@
+"""Evaluate the predictions against the ground truth."""
+
 import json
 import logging
 from pathlib import Path
-from typing import Dict, Tuple
 
 import pandas as pd
-
 from stratigraphy import DATAPATH
 from stratigraphy.benchmark.ground_truth import GroundTruth
 
@@ -12,6 +12,15 @@ logger = logging.getLogger(__name__)
 
 
 def f1(precision: float, recall: float) -> float:
+    """Calculate the F1 score.
+
+    Args:
+        precision (float): Precision.
+        recall (float): Recall.
+
+    Returns:
+        float: The F1 score.
+    """
     if precision + recall > 0:
         return 2 * precision * recall / (precision + recall)
     else:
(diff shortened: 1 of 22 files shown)


src/stratigraphy/benchmark/score.py
```diff
@@ -77,7 +87,8 @@ def evaluate_matching(predictions_path: Path, ground_truth_path: Path) -> tuple[
         overall_recall = sum(document_level_metrics["recall"]) / len(document_level_metrics["recall"])
         logging.info("Macro avg:")
         logging.info(
-            f"F1: {f1(overall_precision, overall_recall):.1%}, precision: {overall_precision:.1%}, recall: {overall_recall:.1%}"
+            f"F1: {f1(overall_precision, overall_recall):.1%},"
+            "precision: {overall_precision:.1%}, recall: {overall_recall:.1%}"
```
stijnvermeeren-swisstopo, line 91:
```suggestion
            f"precision: {overall_precision:.1%}, recall: {overall_recall:.1%}"
```

redur:
Merge conflicts resolved. Tested & metrics still remain the same compared to branch feat/indicator_lines_v2

stijnvermeeren-swisstopo (changes requested):
Please also update the pre-commit section in the README.

So for this pre-commit hook to run, it is not necessary to install ruff in the Conda environment? However, if I also want to run ruff in my IDE (PyCharm), then probably I do need to explicitly install it as a dependency, or how does that work?

redur:
> Please also update the pre-commit section in the README.
> 
> So for this pre-commit hook to run, it is not necessary to install ruff in the Conda environment? However, if I also want to run ruff in my IDE (PyCharm), then probably I do need to explicitly install it as a dependency, or how does that work?

You do not need to install ruff in your conda environment. I believe that the source code of ruff is downloaded from the github repo by pre-commit. And pre-commit calls that code inside our environment. Therefore we don't need to install any of the code. I may be mistaken on that one though. But for sure: No, you don't need to install ruff in order to use it with pre-commit.

Your next question: I am not familiar with how pycharm does the linting. I believe you would need to install ruff in our dev environment and select the right interpreter in pycharm for it to work. Maybe pycharm offers a similar functionality as pre-commit, and you can specify the source code directly. I don't know about that.

src/stratigraphy/benchmark/score.py
redur: Thanks.

stijnvermeeren-swisstopo (approved):

PR #9, opened 2024-03-20
ODR Regression for line merging.

redur:
Implementation of ODR regression for line merging. 

Results slightly improve.

The only document that changes is:
|---:|:-----------------|---------:|----------------------:|----------------------------:|---------:|------------------:|------------------------:|
|  0 | 699240002-bp.pdf | 0.468085 |                    27 |                          16 | 0.425532 |                27 |                      17 |

Vertical lines are still not dealt with, but the code at least throws a warning and does not do unintentional behavior.

### DIFF
--- a/src/stratigraphy/util/geometric_line_utilities.py
+++ b/src/stratigraphy/util/geometric_line_utilities.py
@@ -177,10 +177,12 @@ def merge_parallel_lines_neighbours(
             current_line, line, tol=tol
         ):
             merged_line = _merge_lines(current_line, line)
-            if merged_line is not None:  # no merge possible
+            if merged_line is not None:
+                # current_line and line were merged
                 current_line = merged_line
                 any_merges = True
                 continue
+        # current_line and line were not merged
         merged_lines.append(current_line)
         current_line = line
     merged_lines.append(current_line)
--- a/src/stratigraphy/util/geometric_line_utilities.py
+++ b/src/stratigraphy/util/geometric_line_utilities.py
@@ -177,14 +177,12 @@ def merge_parallel_lines_neighbours(
             current_line, line, tol=tol
         ):
             merged_line = _merge_lines(current_line, line)
-            if merged_line is None:  # no merge possible
-                merged_lines.append(line)
+            if merged_line is not None:  # no merge possible
+                current_line = merged_line
+                any_merges = True
                 continue
-            current_line = merged_line
-            any_merges = True
-        else:
-            merged_lines.append(current_line)
-            current_line = line
+        merged_lines.append(current_line)
+        current_line = line
     merged_lines.append(current_line)
     if any_merges:
         return merge_parallel_lines_neighbours(merged_lines, sorting_function=sorting_function, tol=tol)
@@ -264,22 +262,22 @@ def _odr_regression(x: ArrayLike, y: ArrayLike) -> tuple:
     return phi, r
 
 
-def _merge_lines(line1: Line, line2: Line) -> Line:
+def _merge_lines(line1: Line, line2: Line) -> Line | None:
     """Merge two lines into one.
 
-    The algorithm performes odr regression on the points of the two lines to find the best fit line.
+    The algorithm performs odr regression on the points of the two lines to find the best fit line.
     Then, it calculates the orthogonal projection of the four points onto the best fit line and takes the two points
     that are the furthest apart. These two points are then used to create the merged line.
 
-    Note: Strictly vertical lines cannot be handled. In that case the slope is nan. A warning is thrown
-    and the first line is returned.
+    Note: a few cases (e.g. the lines are sides of a perfect square) the solution is not well-defined. In such a case,
+    the method returns None.
 
     Args:
         line1 (Line): First line to merge.
         line2 (Line): Second line to merge.
 
     Returns:
-        Line: The merged line.
+        Line | None: The merged line.
     """
     x = np.array([line1.start.x, line1.end.x, line2.start.x, line2.end.x])
     y = np.array([line1.start.y, line1.end.y, line2.start.y, line2.end.y])
--- a/src/stratigraphy/util/geometric_line_utilities.py
+++ b/src/stratigraphy/util/geometric_line_utilities.py
@@ -236,7 +236,7 @@ def merge_parallel_lines(lines: list[Line], tol: int = 8, angle_threshold: float
 def _odr_regression(x: ArrayLike, y: ArrayLike) -> tuple:
     """Perform orthogonal distance regression on the given data.
 
-    Note: If the problem is ill defined (i.e. denominator == nominator == 0),
+    Note: If the problem is ill-defined (i.e. denominator == nominator == 0),
     then the function will return phi=np.nan and r=np.nan.
 
     Args:
@@ -258,6 +258,10 @@ def _odr_regression(x: ArrayLike, y: ArrayLike) -> tuple:
         return np.nan, np.nan
     phi = 0.5 * np.arctan2(nominator, denominator)  # np.arctan2 can deal with np.inf due to zero division.
     r = x_mean * cos(phi) + y_mean * sin(phi)
+
+    if r < 0:
+        r = -r
+        phi = phi + np.pi
 
     return phi, r
 
--- a/src/stratigraphy/util/geometric_line_utilities.py
+++ b/src/stratigraphy/util/geometric_line_utilities.py
@@ -176,7 +176,11 @@ def merge_parallel_lines_neighbours(
         if _are_parallel(current_line, line, angle_threshold=angle_threshold) and _are_close(
             current_line, line, tol=tol
         ):
-            current_line = _merge_lines(current_line, line)
+            merged_line = _merge_lines(current_line, line)
+            if merged_line is None:  # no merge possible
+                merged_lines.append(line)
+                continue
+            current_line = merged_line
             any_merges = True
         else:
             merged_lines.append(current_line)
@@ -216,6 +220,9 @@ def merge_parallel_lines(lines: list[Line], tol: int = 8, angle_threshold: float
                 line, merge_candidate, tol=tol
             ):
                 line = _merge_lines(line, merge_candidate)
+                if line is None:  # No merge possible
+                    merged_lines.append(line)
+                    continue
                 merged_lines.remove(merge_candidate)
                 merged_lines.append(line)
                 merged = True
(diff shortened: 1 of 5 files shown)


src/stratigraphy/util/geometric_line_utilities.py
```diff
@@ -223,36 +229,106 @@ def merge_parallel_lines(lines: list[Line], tol: int = 8, angle_threshold: float
...
+    """
+    v = np.array([1, slope])
+    projection = np.dot(np.outer(v, v) / np.inner(v, v), np.array([point.x, point.y - intercept])) + np.array(
+        [0, intercept]
+    )
+    if np.shape(projection) != (2,):
+        logger.critical(
+            f"Wrong shape for orthogonal projection. The correct shape is (2,), but it is {np.shape(projection)}."
+        )
+        raise ValueError
```
redur, line 327: Any nan's should be caught before. This is here to make sure there would be an appropriate action if this case is ever to happen.

src/stratigraphy/util/plot_utils.py
```diff
@@ -15,15 +20,20 @@ def _draw_lines(open_cv_img, lines, scale_factor=1):
...
+                np.dot(scale_factor, line.start.tuple),
+                np.dot(scale_factor, line.end.tuple),
+                color,
+                1,
+            )
+            cv2.circle(open_cv_img, np.dot(scale_factor, line.start.tuple), radius=1, color=(0, 0, 255), thickness=-1)
+            cv2.circle(open_cv_img, np.dot(scale_factor, line.end.tuple), radius=1, color=(0, 0, 255), thickness=-1)
+
+        except cv2.error as e:
+            logging.warning(f"Error drawing line. Exception: {e}. Skipping to draw the line.")
```
redur, line 35: Happened for nan's. E.g. for when we attempt to merge vertical lines.

No warning is thrown now, as we can deal with all the cornercases present in our data.

src/stratigraphy/util/geometric_line_utilities.py
```diff
@@ -223,36 +229,106 @@ def merge_parallel_lines(lines: list[Line], tol: int = 8, angle_threshold: float
...
 
 
+def odr_regression(x: ArrayLike, y: ArrayLike) -> tuple[float, float]:
+    """Perform orthogonal distance regression on the given data.
+
+    Solves the line equation y = mx + c for the best fit values of m and c.
+    Different to linear regression, ODR takes into account the error in both x and y.
+    As such, it can handle horizontal and vertical lines.
+
+    Note: For vertical lines, the slope is nan.
```
stijnvermeeren-swisstopo, line 239: Are you sure about this? I tried to reproduce it, but I actually got a slope of `0` even for points that are on a vertical line. E.g.

```
    x = np.array([1, 1, 1])
    y = np.array([0, 1, 2])
    slope, intercept = odr_regression(x, y)
    print(slope, intercept)
```

src/stratigraphy/util/geometric_line_utilities.py
```diff
@@ -223,36 +229,106 @@ def merge_parallel_lines(lines: list[Line], tol: int = 8, angle_threshold: float
...
+    if np.isnan(slope):
+        logger.warning("Merged line slope is nan. Returning line1 in absence of any better choice.")
+        return line1
+    projected_points = []
+    for point in [line1.start, line1.end, line2.start, line2.end]:
+        projection = _get_orthogonal_projection_to_line(point, slope, intercept)
+        projected_points.append(projection)
+    # Take the two points such that the distance between them is the largest.
+    # Since there are only 4 points involved, we can just calculate the distance between all points.
+    point_combinations = list(combinations(projected_points, 2))
```
stijnvermeeren-swisstopo, line 292: I don't really like the quadratic complexity here, though for us it's probably not a problem because we only ever call this with four points.
I guess a more efficient method would be: take the points with min and max value for the x coordinates, unless the points all have equal x coordinates, in that case take the points with the min and max values for the y coordinates.

src/stratigraphy/util/geometric_line_utilities.py
```diff
@@ -223,36 +229,106 @@ def merge_parallel_lines(lines: list[Line], tol: int = 8, angle_threshold: float
...
+    """Merge two lines into one.
+
+    The algorithm performes odr regression on the points of the two lines to find the best fit line.
+    Then, it calculates the orthogonal projection of the four points onto the best fit line and takes the two points
+    that are the furthest apart. These two points are then used to create the merged line.
 
-    Note: This method is not robust for near vertical lines. We assume
-    an ordering in the x-coordinates of the lines.
+    Note: Strictly vertical lines cannot be handled. In that case the slope is nan. A warning is thrown
+    and the first line is returned.
```
stijnvermeeren-swisstopo, line 271: Here's a derivation of a fitting line that also supports the vertical case: https://www.researchgate.net/publication/272179120_A_tutorial_on_the_total_least_squares_method_for_fitting_a_straight_line_and_a_plane
I did not find a Python implementation of this yet, though. Maybe you can find something? Or otherwise, it does not look too difficult to implement the actual formulas ourselves.

stijnvermeeren-swisstopo (changes requested):

src/stratigraphy/util/geometric_line_utilities.py
```diff
@@ -223,36 +229,106 @@ def merge_parallel_lines(lines: list[Line], tol: int = 8, angle_threshold: float
...
+    )
+    if np.shape(projection) != (2,):
+        logger.critical(
+            f"Wrong shape for orthogonal projection. The correct shape is (2,), but it is {np.shape(projection)}."
+        )
+        raise ValueError
+    if np.isnan(projection).any():
+        logger.critical("Projection is nan. Cannot calculate orthogonal projection.")
+        raise ValueError
+    return Point(int(np.round(projection[0], 0)), int(np.round(projection[1], 0)))
```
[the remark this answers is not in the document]
redur: It's because we operate on a pixel grid, and some downstream functionalities such as plotting might not work any more. 

I even thought about forcing the class object Point to take integers only.

I believe plotting with cv2 is the only thing that doesnt work. (Mainly as we don't use any other cv2 functionalities afterwards any more).

src/stratigraphy/util/geometric_line_utilities.py
redur: Let's write some unit tests for this function.

src/stratigraphy/util/geometric_line_utilities.py
redur: Since the number of points is always four, I deemed it constant in time. And it executes fast. I don't see any use case where we feed more than 4 points to this function.

redur:
Implemented regression version as you suggested from here: https://www.researchgate.net/publication/272179120_A_tutorial_on_the_total_least_squares_method_for_fitting_a_straight_line_and_a_plane.

The scores are identical to the previous odr implementation.


For now it is not entirely clear for me on how to deal with the case where nominator and denominator are 0 in the arctan condition. I believe the problem is ill defined in that case. E.g. four points on a square do not well define the best line. (could be horizontal or vertical). The denominator == nominator == 0 case does happen 4 times in our data set.

I think the best option would be to do nothing with the lines in that case.

redur:
I attempted to do "nothing" in case of an ill defined problem now.

src/stratigraphy/util/geometric_line_utilities.py
```diff
@@ -171,7 +176,11 @@ def merge_parallel_lines_neighbours(
         if _are_parallel(current_line, line, angle_threshold=angle_threshold) and _are_close(
             current_line, line, tol=tol
         ):
-            current_line = _merge_lines(current_line, line)
+            merged_line = _merge_lines(current_line, line)
```
redur, line 179: Any idea if this could be solved prettier?

redur:
The latest addition slightly decreased performance for a single case. However, this branch is anyways outdated with respect to the block_line_ratio param. I suggest we merge it as is and revisit the parameter at a later time.

stijnvermeeren-swisstopo:
Generally it looks pretty good, though sometimes I still see some issues with the merged lines. For example in 267124133-bp.pdf:
image
gets merged to
image

It seems that the many short lines progressively change the slope of the merged line, until it is very different from the original.

Maybe we need to take some weighted average in the ODR regression, where the points are weighted by the length of the line (i.e. long lines have more influence over the final result than short lines) and maybe also by the number of lines that were already merged together. Let me try something like that, to see if we get an improvement...

stijnvermeeren-swisstopo:
I implemented the idea mentioned above (differently weighted points for the regression) and it seems to give more satisfactory results for the line merging, and an F1-score improvement of 0.1%.
However, the implementation is not very clean yet, and not covered by unit tests yet either, so I'm putting it into a draft PR for now:
https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/12/files

In this PR, I've pushed a few minor changes (documentation corrections + minor refactoring).

tests/test_geometric_line_utilities.py
```diff
@@ -0,0 +1,99 @@
...
+from stratigraphy.util.dataclasses import Line, Point
+from stratigraphy.util.geometric_line_utilities import (
+    _get_orthogonal_projection_to_line,
+    _merge_lines,
+    _odr_regression,
+)
+
+
+# Remember, phi is orthogonal to the line we are to parameterize
+# The way phi is defined is a bit counterintuitive, but it seems consistent all along.
```
stijnvermeeren-swisstopo, line 14: Yes, I'm also still confused by this. I would have expected the sign to be the other way around... Needs more investigation.

stijnvermeeren-swisstopo: Actually, now it makes more sense, after ensuring that r>0, as also recommended in the paper https://www.researchgate.net/publication/272179120_A_tutorial_on_the_total_least_squares_method_for_fitting_a_straight_line_and_a_plane

stijnvermeeren-swisstopo (commented):
TODO for next week:
- Clean up / add more unit tests
- Decide if we want to introduce a weighted average into the regression model, or if we leave that for later

Apart from that, the PR looks pretty good to me :).

src/stratigraphy/util/geometric_line_utilities.py
redur: Thanks. The change makes sense. I'd just adjust the comment after `if merged_line is not None: # no merge possible`.  to `#  check if merge is possible`. Otherwise it's counterintuitive what happens in the if clause.

I can do the change :).

redur:
> TODO for next week:
> 
> * Clean up / add more unit tests
> * Decide if we want to introduce a weighted average into the regression model, or if we leave that for later
> 
> Apart from that, the PR looks pretty good to me :).

What exactly would you like to clean up?

stijnvermeeren-swisstopo (approved):
I fixed the code comment and added a few test cases.

If you are happy with those, then we can merge this PR.

I suggest that we don't wait for https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/12 , but rather review that one separately.

PR #10, opened 2024-03-21
introduce class for description block splitting

stijnvermeeren-swisstopo:
Object-oriented approach to the different methods for partitioning description lines into blocks.

### DIFF
--- a/src/stratigraphy/util/find_description.py
+++ b/src/stratigraphy/util/find_description.py
@@ -1,11 +1,13 @@
 """This module contains functions to find the description (blocks) of a material in a pdf page."""
 
-from collections.abc import Callable
-
 import fitz
-import numpy as np
 
 from stratigraphy.util.dataclasses import Line
+from stratigraphy.util.description_block_splitter import (
+    SplitDescriptionBlockByLeftHandSideSeparator,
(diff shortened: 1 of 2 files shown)


stijnvermeeren-swisstopo:
@redur was meinst du zu diese Code-Struktur?

src/stratigraphy/util/find_description.py
```diff
@@ -112,145 +110,8 @@ def get_description_blocks(
         # indicators are dropped. That's why we have to consider vertical spacing as well.
         _blocks = []
         for block in blocks:
-            _blocks.extend(_split_block_by_vertical_spacing(block.lines, threshold=threshold))
+            _blocks.extend(splitter.create_blocks(block.lines))
```
redur, line 113: Nice that we can reuse the same object here.

redur (approved):
Thank you Stijn. I believe you are right and this is overall more readable and better abstracted than it was before.

Did you check if executing before and after leads to the same results?

stijnvermeeren-swisstopo:
I verified that the scores are identical for all documents, between the current main branch and this branch. (Overall F1 score: 86.6%.)

PR #11, opened 2024-03-22
Feat/better depth value layers

redur:
**Description**
Updated the logic that decides when we should add a material description for the final, open ended depth interval.

Previously, we where looking at whether the last material description block is visually lower than the last depth entry. We now additionally also check if there are at least as many blocks than depth intervals.

The idea is, that we need to have enough blocks to assign any block to the last, open ended interval. Previously, the block recognition did not have very high accuracy. But with the improvements due to the line detection we gained accuracy in the initial block recognition before we enter transform groups.

**Document Level Metrics**

|---:|:-----------------|---------:|----------------------:|----------------------------:|---------:|------------------:|------------------------:|
|  0 | 267124110-bp.pdf | 1        |                     7 |                           0 | 0.933333 |                 7 |                       0 |
|  1 | 268125592-bp.pdf | 1        |                     9 |                           0 | 0.947368 |                 9 |                       0 |
|  2 | 683248005-bp.pdf | 0.764706 |                    14 |                           1 | 0.742857 |                14 |                       1 |
|  3 | 679249080-bp.pdf | 1        |                    12 |                           0 | 0.88     |                12 |                       1 |
|  4 | 267123078-bp.pdf | 1        |                    18 |                           0 | 0.918919 |                18 |                       1 |

In all those cases, we previously expected a final block where there wasn't supposed to be one. Thus, in transfrom_groups we mistakenly splitted one block into two.

**Results**
For all cases, we can only see improvements of the recognition in F1.
Compared to current main this results to an improvement of F1 from 86.6% to 86.8%.

### DIFF
--- a/src/stratigraphy/util/depthcolumn.py
+++ b/src/stratigraphy/util/depthcolumn.py
@@ -201,6 +201,13 @@ class BoundaryDepthColumn(DepthColumn):
         )
 
     def depth_intervals(self) -> list[BoundaryInterval]:
+        """Creates a list of depth intervals from the depth column entries.
+
+        The first and last depth intervals are open-ended.
+
+        Returns:
+            list[BoundaryInterval]: A list of depth intervals.
+        """
         depth_intervals = [BoundaryInterval(None, self.entries[0])]
         for i in range(len(self.entries) - 1):
             depth_intervals.append(BoundaryInterval(self.entries[i], self.entries[i + 1]))
@@ -302,6 +309,33 @@ class BoundaryDepthColumn(DepthColumn):
         return [BoundaryDepthColumn(segment) for segment in segments]
 
     def identify_groups(self, description_lines: list[TextLine], geometric_lines: list[Line], **params) -> list[dict]:
+        """Identifies groups of description blocks that correspond to depth intervals.
+
+        Note: includes a heuristic of whether there should be a group corresponding to a final depth interval
+        starting from the last depth entry without any end value.
+
+        Args:
+            description_lines (list[TextLine]): A list of text lines that are part of the description.
+            geometric_lines (list[Line]): A list of geometric lines that are part of the description.
+            params (dict): A dictionary of parameters used for line detection.
+
+        Returns:
+            list[dict]: A list of groups, where each group is a dictionary
+                        with the keys "depth_intervals" and "blocks".
+
+        Example:
+            [
+                {
+                    "depth_intervals": [BoundaryInterval(None, 0.1), BoundaryInterval(0.1, 0.3), ...],
+                    "blocks": [DescriptionBlock(...), DescriptionBlock(...), ...]
+                },
+                {
+                    "depth_intervals": [BoundaryInterval(0.1, 0.3)],
+                    "blocks": [DescriptionBlock(...), DescriptionBlock(...), ...]
+                },
+                ...
+            ]
+        """
         depth_intervals = self.depth_intervals()
 
         groups = []
@@ -315,6 +349,8 @@ class BoundaryDepthColumn(DepthColumn):
             left_line_length_threshold=params["left_line_length_threshold"],
             target_layer_count=len(depth_intervals),
         )
+
+        open_end_interval = len(all_blocks) >= len(depth_intervals)
 
         block_index = 0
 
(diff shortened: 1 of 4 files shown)


src/stratigraphy/util/depthcolumn.py
```diff
@@ -345,9 +388,11 @@ def identify_groups(
                 current_intervals = []
             else:
                 # only add "unlimited" final layer, if the description is visually below the final depth label
+                # and if there are at least as many blocks as intervals (intervals always include the open ended one)
                 if interval.end is not None or (
                     len(current_blocks)
                     and current_blocks[-1].rect.y1 > interval.start.rect.y1 + interval.start.rect.height / 2
+                    and open_end_interval
```
redur, line 395: That's where the change in the logic happens.

stijnvermeeren-swisstopo:
@redur Did you identify the documents in our dataset where the final "open ended" interval is actually required? How many of them are there?

redur:
> @redur Did you identify the documents in our dataset where the final "open ended" interval is actually required? How many of them are there?

Let me check.

redur:
I could not identify a single pdf for which we would need to detect an open ended interval. However, removing the open ended interval from the code slightly decreases F1. This is the case when a) the algorithm detects an open ended interval, and b) some depth entries are missing. The change is less than 0.1%.

I will remove the open ended interval, as I believe this makes more sense.

redur:
I made a suggestion regarding open ended intervals. Completely removing open ended intervals in `depth_intervals()` in class `BoundaryDepthColumn` leads to strange behavior and to many non detected blocks.

I must admit, I did not fully understand the logic here. Happy to discuss.

src/stratigraphy/util/depthcolumn.py
```diff
@@ -345,10 +388,8 @@ def identify_groups(
                 current_intervals = []
             else:
                 # only add "unlimited" final layer, if the description is visually below the final depth label
-                if interval.end is not None or (
-                    len(current_blocks)
-                    and current_blocks[-1].rect.y1 > interval.start.rect.y1 + interval.start.rect.height / 2
-                ):
+                # and if there are at least as many blocks as intervals (intervals always include the open ended one)
```
stijnvermeeren-swisstopo, line 391: Update comment

src/stratigraphy/util/depthcolumn.py
```diff
@@ -207,10 +209,19 @@ def strictly_contains(self, other: BoundaryDepthColumn):
...
+        Returns:
+            list[BoundaryInterval]: A list of depth intervals.
+        """
         depth_intervals = [BoundaryInterval(None, self.entries[0])]
         for i in range(len(self.entries) - 1):
             depth_intervals.append(BoundaryInterval(self.entries[i], self.entries[i + 1]))
-        depth_intervals.append(BoundaryInterval(self.entries[len(self.entries) - 1], None))
+        depth_intervals.append(
+            BoundaryInterval(self.entries[len(self.entries) - 1], None)
+        )  # even though no open ended intervals are allowed, they are still useful for matching
```
stijnvermeeren-swisstopo, line 224: Add: "especially for documents where the material description rectangle is too tall (and includes additional lines below the actual material descriptions)"

stijnvermeeren-swisstopo (changes requested):
I believe we need to open ended interval because of the condition
```            self.end is None or all_blocks[block_index].rect.y1 < self.end.rect.y1``` 
in `BoundaryInterval.matching_blocks()`.

I tried removing this interval, and replacing the `self.end is None` with a `self.is_final` flag, that is set to true for the final interval, but this does not work as well either. Sometimes, the material description rectangle goes a little too far down in the document, and in those case we're actually better off completely dropping these lines if we already have enough material descriptions (as happens with the current implementation), rather than forcing them to be also linked to the final depth interval.

So I think we can probably merge this, but we should think about how we can determine the boundaries of the material descriptions a little more exactly, and once we have that, we might be able to simplify this logic a little further.

**Comment by redur** [comment]
Review comments are resolved.

File: src/stratigraphy/util/depthcolumn.py [hunk 4]
```diff
@@ -345,10 +390,9 @@ def identify_groups(
                 current_intervals = []
             else:
                 # only add "unlimited" final layer, if the description is visually below the final depth label
```
stijnvermeeren-swisstopo, line 392: remove redundant comment line

stijnvermeeren-swisstopo (changes requested):

stijnvermeeren-swisstopo (approved):

PR #12, opened 2024-03-22
Feat/deming regression stijn

stijnvermeeren-swisstopo:


src/stratigraphy/util/dataclasses.py
```diff
@@ -16,6 +18,9 @@ class Point:
     def tuple(self) -> (float, float):
         return self.x, self.y
 
+    def distance_to(self, point: Point) -> float:
```
redur, line 21: Where not using this at all, are we? Happy to keep it anyways as it's a logic extension of this class.

src/stratigraphy/util/geometric_line_utilities.py
```diff
@@ -235,7 +235,7 @@ def merge_parallel_lines(lines: list[Line], tol: int = 8, angle_threshold: float
         return merged_lines
 
 
-def _odr_regression(x: ArrayLike, y: ArrayLike) -> tuple:
+def _odr_regression(x: ArrayLike, y: ArrayLike, weights: ArrayLike) -> tuple:
```
redur, line 238: Would be good to add a default value for weights. Something like:
`weights: ArrayLike = None`

And then something like
```
if weights is None:
    weights = np.ones((len(x), ))
```

tests/test_geometric_line_utilities.py
```diff
@@ -96,4 +96,7 @@ def merge_lines_case(request):  # noqa: D103
 def test_merge_lines(merge_lines_case):  # noqa: D103
     line1, line2, expected_merged_line = merge_lines_case
     merged_line = _merge_lines(line1, line2)
-    assert merged_line == expected_merged_line  # Adjust this line if Line objects can't be compared directly
+    assert (
+        pytest.approx(merged_line.start.tuple) == expected_merged_line.start.tuple
+        and pytest.approx(merged_line.end.tuple) == expected_merged_line.end.tuple
+    )  # Adjust this line if Line objects can't be compared directly
```
redur, line 102:
```suggestion
    )
```

src/stratigraphy/util/geometric_line_utilities.py
```diff
@@ -244,14 +244,15 @@ def _odr_regression(x: ArrayLike, y: ArrayLike) -> tuple:
...
         tuple: (phi, r), the best fit values for the line equation in normal form.
     """
-    x_mean = np.mean(x)
-    y_mean = np.mean(y)
-    nominator = -2 * np.sum((x - x_mean) * (y - y_mean))
-    denominator = np.sum((y - y_mean) ** 2 - (x - x_mean) ** 2)
+    x_mean = np.mean(np.dot(weights, x)) / np.sum(weights)
+    y_mean = np.mean(np.dot(weights, y)) / np.sum(weights)
+    nominator = -2 * np.sum(np.dot(weights, (x - x_mean) * (y - y_mean)))
+    denominator = np.sum(np.dot(weights, (y - y_mean) ** 2 - (x - x_mean) ** 2))
```
redur, line 255: Is this consistent?

I.e. if we are to use weights like [2, 2, 2, 2] would it lead to the same result as [1, 1, 1, 1]?

I am not sure if the weights would cancel.

I don't say it's wrong. I'm just confused. Let's discuss.

redur: For me, it would make sense to weight the differences `(x - x_mean)` and `(y - y_mean)`.

I.e. weight all the `(x_i - x_mean)` and `(y_i - y_mean)` with `weights_i`.

Thus, something like:

```
x_mean = np.mean(np.dot(weights, x)) / np.sum(weights)
y_mean = np.mean(np.dot(weights, y)) / np.sum(weights)
nominator = -2 * np.sum(np.dot(weights**2, (x - x_mean) * (y - y_mean)))
denominator = np.sum(np.dot(weights**2, (y - y_mean) ** 2 - (x - x_mean) ** 2))
```

Since the weights for x and y are the same, we can just square them.

redur (changes requested):
Hi Stijn,

Great idea with the weighting. I'm a bit uncertain how it should look like in practise.

Usually in regression, one takes the uncertainties of the error as a weight. But we're solving something different here. I believe your approach makes sense, but I don't have any mathematical argument for it. In essence, we're trying to scale the deviations `(x - x_mean)` as well as `(y - y_mean)`. Which I believe is sound.

Added some comments.

src/stratigraphy/util/geometric_line_utilities.py
redur: A better test would be something like: `[1, 1, 2, 2]` and `[2, 2, 4, 4]`.

stijnvermeeren-swisstopo: I have reverted this change, going back to using the simple weights, instead of the squared weights here.

I believe that this is mathematically more correct, though for our purposes, it does not change much. (At least in the case of perfectly parallel input lines, the resulting merged lines will be identical.)

Section "V.A." of https://www.researchgate.net/publication/272179120 explicitly gives formulas for weighted orthogonal distance regression, and these use the simple weights, not the squared weights. The mathematical derivation of these formulas also seems correct to me. Moreover, arguments resting on "dimensional analysis" can't really be used here, as the weights are unitless/dimensionless anyway, and stay the same even when taking a derivative with respect to r or phi (which actually occurs in the derivation of the formulas).

stijnvermeeren-swisstopo:
I reverted the squared weights (as explained in a different comment on this PR) and added some documentation. Could you check the PR again @redur ?

src/stratigraphy/util/geometric_line_utilities.py
stijnvermeeren-swisstopo: And yes, using weights like [2, 2, 2, 2] leads to the same result as [1, 1, 1, 1]. Multiplying weights by a constant factor cancels out everywhere.

PR #13, opened 2024-03-25
Doc: Extended project documentation in the README.

redur:
The readme now contains a better overview of the project with guidelines on how to run and interpret the data extraction pipeline.

Also comes with minor updates in docstrings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,9 +54,9 @@ Example: predictions.json
 {
     "685256002-bp.pdf": {  # file name
         "page_1": {
-            "layers": [
+            "layers": [  # a layer corresponds to a material layer in the borehole profile
                 {
-                    "material_description": {
+                    "material_description": {  # all information about the complete description of the material of the layer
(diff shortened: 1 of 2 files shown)


redur:
> This will require updates after #14 .

Updates are now incorporated.

stijnvermeeren-swisstopo (commented):
I pushed a few minor changes. Happy to merge it like this.

It's slightly weird that we have comments with `#` within the JSON example, while JSON actually does not support comments like that, but since the example is not intended for copy-pasting, I guess it's fine. I don't immediately have a better solution for adding "documentation annotations" to the JSON example.

redur:
> I pushed a few minor changes. Happy to merge it like this.
> 
> It's slightly weird that we have comments with `#` within the JSON example, while JSON actually does not support comments like that, but since the example is not intended for copy-pasting, I guess it's fine. I don't immediately have a better solution for adding "documentation annotations" to the JSON example.

I don't have a better solution either.

PR #14, opened 2024-03-26
Refactor: Decouple inference from evaluation.

redur:
**Content**
Previously, inference (i.e. perform_matching) was dependent on ground_truth information. Furthermore, various validation steps in form of drawing onto pdf pages was done in perform matching. This commit moves all evaluation steps to evaluate_matching.

The file predictions.json is adjusted and now contains much more information. The information is sufficient to reconstruct all drawings in the pdf pages, without the need of the python objects.

In particular, predictions.json now contains:
  - information about material_description and depth_column pairs (bounding boxes & values)
  - more information about the identified material description blocks (bounding boxes as well as lines)

Also comes with some refactoring of the main.py script.

Scores are not affected.

**Discussion**
Let me know what you think about the interface of predictions.json. I believe we'll require something like this in the future anyways.

Most classes for our stratigraphy module now have a `.to_json() `method. We could think of a `create_from_json()` init method, but that would require a bit more thought. I would only attempt this, if the frontend would make use of our classes - or if there are further downstream pipelines that could benefit from the initial objects (e.g. classification of material description texts).

With this PR, some class methods become obsolete. They are kept for the moment until we have a final decision on how to define the interfaces.

Happy to receive your thoughts on this.

### DIFF
--- a/src/stratigraphy/benchmark/score.py
+++ b/src/stratigraphy/benchmark/score.py
@@ -7,8 +7,30 @@ from pathlib import Path
 import pandas as pd
 from stratigraphy import DATAPATH
 from stratigraphy.benchmark.ground_truth import GroundTruth
+from stratigraphy.util.draw import draw_predictions
 
 logger = logging.getLogger(__name__)
+
+
+def _get_from_all_pages(predictions: dict) -> dict:
+    """Get all predictions from all pages.
+
+    Args:
+        predictions (dict): The predictions.
+
+    Returns:
+        dict: The predictions from all pages.
+    """
+    all_predictions = {}
+    layers = []
+    depths_materials_column_pairs = []
+    for page in predictions:
+        layers.extend(predictions[page]["layers"])
+        depths_materials_column_pairs.extend(predictions[page]["depths_materials_column_pairs"])
+    all_predictions["layers"] = layers
+    if len(depths_materials_column_pairs):
+        all_predictions["depths_materials_column_pairs"] = depths_materials_column_pairs
+    return all_predictions
 
 
 def f1(precision: float, recall: float) -> float:
@@ -27,7 +49,9 @@ def f1(precision: float, recall: float) -> float:
         return 0
 
 
-def evaluate_matching(predictions_path: Path, ground_truth_path: Path) -> tuple[dict, pd.DataFrame]:
+def evaluate_matching(
+    predictions_path: Path, ground_truth_path: Path, directory: Path, out_directory: Path
+) -> tuple[dict, pd.DataFrame]:
     """Calculate F1, precision and recall for the predictions.
 
     Calculate F1, precision and recall for the individual documents as well as overall.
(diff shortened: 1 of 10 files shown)


src/stratigraphy/benchmark/score.py
```diff
@@ -27,7 +49,9 @@ def f1(precision: float, recall: float) -> float:
         return 0
 
 
-def evaluate_matching(predictions_path: Path, ground_truth_path: Path) -> tuple[dict, pd.DataFrame]:
+def evaluate_matching(
+    predictions_path: Path, ground_truth_path: Path, directory: Path, out_directory: Path
```
stijnvermeeren-swisstopo, line 53: these new params are missing at the bottom of this file

src/stratigraphy/benchmark/score.py
```diff
@@ -45,6 +71,8 @@ def evaluate_matching(predictions_path: Path, ground_truth_path: Path) -> tuple[
     with open(predictions_path) as in_file:
         predictions = json.load(in_file)
 
+    draw_predictions(predictions, ground_truth, directory, out_directory)
```
stijnvermeeren-swisstopo, line 74: We could take this one step further, and instead of calling `ground_truth.is_correct()` for every prediction twice (once when drawing and once when computing the scores), we could compute the truth value for every prediction in advance, and then reuse it for both drawing and computing the scores.

stijnvermeeren-swisstopo (changes requested):

src/stratigraphy/benchmark/score.py
redur: Addressed. Thank you.

src/stratigraphy/util/draw.py
```diff
@@ -54,10 +177,31 @@ def draw_layer(
...
                     line_anchor * page.derotation_matrix,
                     fitz.Point(block_rect.x0, (block_rect.y0 + block_rect.y1) / 2) * page.derotation_matrix,
                     color=fitz.utils.getColor(color),
                 )
+
+
+def _get_line_anchor(interval):
+    if interval["start"] and interval["end"]:
+        return fitz.Point(
+            interval["start"]["rect"][2], (interval["start"]["rect"][1] + interval["end"]["rect"][3]) / 2
```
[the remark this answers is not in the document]
redur: Addressed. Thank you.

stijnvermeeren-swisstopo (approved):
@redur I pushed a few minor fixes:
- Fix logic in `GroundTruthForFile`, so that `is_correct()` also returns `False` (instead of `None`) if all ground truth entries have already been matched
- Remove computation for metrics "excluding worst 5"; I don't need them anymore, and I guess you don't either?
- Fix code for the case that `precision == 0`.

I'm not a huge fan of the nested dicts that we are working with in the evaluation script. I would generally prefer more restrictively typed data structures. However, I don't see a quick and simple way of implementing that without adding a significant amount of boilerplate code, so I'm ok with keeping it like this for now.

PR #15, opened 2024-03-26
Doc: Update documentation regarding package manager.

redur:


stijnvermeeren-swisstopo (approved):
I switched to Miniforge, everything seems to work fine.

PR #18, opened 2024-03-28
Refactor/indicator line weak condition

stijnvermeeren-swisstopo:
- Decrease the `merging_tolerance` value from 5 to 3, to avoid longer lines being merged across different adjacent letters (see screenshot below: the line along "du" and the one along "mi" happens with the old value, but no longer with the new value)
- Remove default values for parameters in the individual methods, to avoid having (potentially inconsistent) default values across many different files and methods, and to ensure that the values from the yml files are always used and passed along.
- Simplify logic related to short indicator lines
  - Remove dependency on `current_block`, use the coordinates of `last_line` and `current_line` as a reference instead
  - Remove special case for blocks with at least three text lines, as I don't see this bringing any advantage anymore

image

Overall F1 score improves with +0.1% (86.6% -> 86.7%).

Individual files with changed F1 scores:
- 681246002-bp.pdf (86% -> 100%)
- 268124125-bp.pdf (2.6% -> 2.7%)
- 268124635-bp.pdf (83% -> 77%; bounding box from OCR is not accurate, and therefore the indicator line is not recognised, as it does not intersect the left-hand-side of the text line)
- 699248001-bp.pdf (85% -> 74%, only very short horizontal indicator lines, would work better if we would also be able to consider connected diagonal line segments)
- 268124599-bp.pdf (22% -> 57%)
- 267123060-bp.pdf (94% -> 75%; dashed lined no longer recognized as a single merged line)
- 691251009-bp.pdf (23% -> 25%)
- 268124391-bp.pdf (92% -> 94%)
- 268125447-bp.pdf (36% -> 42%)
- 268124307-bp.pdf (92% -> 94%)

### DIFF
--- a/config/line_detection_params.yml
+++ b/config/line_detection_params.yml
@@ -14,6 +14,6 @@ vertical_lines_threshold: 0.6
 
 line_merging_params:
   clustering_threshold: 0.08
-  merging_tolerance: 5
+  merging_tolerance: 3
   angle_threshold: 5  # maximum difference in degrees two line segments can have in order to be merged.
   use_clustering: false  # No measureable performance improvement for "true"
--- a/config/line_detection_params.yml
+++ b/config/line_detection_params.yml
@@ -14,6 +14,6 @@ vertical_lines_threshold: 0.6
 
 line_merging_params:
   clustering_threshold: 0.08
-  merging_tolerance: 3
+  merging_tolerance: 3.  # maximum distance between two lines that are merged
   angle_threshold: 5  # maximum difference in degrees two line segments can have in order to be merged.
   use_clustering: false  # No measureable performance improvement for "true"
--- a/src/stratigraphy/util/description_block_splitter.py
+++ b/src/stratigraphy/util/description_block_splitter.py
@@ -98,19 +98,17 @@ class SplitDescriptionBlockByLine(DescriptionBlockSplitter):
 class SplitDescriptionBlockByLeftHandSideSeparator(DescriptionBlockSplitter):
     """Creates blocks based on shorter lines at the left-hand side of the material description text."""
 
-    def __init__(self, length_threshold: float, geometric_lines: list[Line], material_description_rect: fitz.Rect):
+    def __init__(self, length_threshold: float, geometric_lines: list[Line]):
         """Create a new SplitDescriptionBlockByLine instance.
 
         Args:
             length_threshold (int): The minimum length of a line segment on the left side of a block to split it.
             geometric_lines (list[Line]): The geometric lines detected in the pdf page.
-            material_description_rect (fitz.Rect): The bounding box for all material descriptions.
         """
         super().__init__()
         self.length_threshold = length_threshold
         self.set_terminated_by_line_flag = False
         self.geometric_lines = geometric_lines
-        self.material_description_rect = material_description_rect
 
     def separator_condition(self, last_line: TextLine, current_line: TextLine) -> bool:
         """Check if a block is separated by a line segment on the left side of the block.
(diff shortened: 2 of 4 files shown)


config/line_detection_params.yml
```diff
@@ -14,6 +14,6 @@ vertical_lines_threshold: 0.6
 
 line_merging_params:
   clustering_threshold: 0.08
-  merging_tolerance: 5
+  merging_tolerance: 3
```
redur, line 17:
```suggestion
  merging_tolerance: 3.  # maximum distance between two lines that are merged
```

redur: Maybe it's good to add a comment here.

src/stratigraphy/util/description_block_splitter.py
```diff
@@ -128,25 +126,16 @@ def separator_condition(self, last_line: TextLine, current_line: TextLine, curre
         for line in self.geometric_lines:
             line_y_coordinate = (line.start.y + line.end.y) / 2
 
-            line_cuts_lefthandside_of_block = line.start.x < current_block.rect.x0 < line.end.x
+            line_cuts_lefthandside_of_text = (line.start.x < last_line.rect.x0 < line.end.x) and (
+                line.start.x < current_line.rect.x0 < line.end.x
+            )
```
redur, line 131: Basically, you are saying that the indicator line should cut in between two lines. Therefore we look at where those two lines start. But it is not so important where the full block starts. Is that correct?

I don't see any strong reason for one or the other. As long as the scores improve or remain the same.

I imagine there is some difference between block.rect.x0 and last_line.rect.x0.

src/stratigraphy/util/description_block_splitter.py
```diff
@@ -128,25 +126,16 @@ def separator_condition(self, last_line: TextLine, current_line: TextLine, curre
...
-                and (np.abs(line.start.x - line.end.x) > self.length_threshold - 2)
-                and (len(current_block.lines) > 1)
-            )  # if block has at least three lines, we weaken the splitting condition.
-            # It is three lines because the statement means that block.lines has at least two elements.
-            # The third line is current_line
-            # The reason for the splitting is, that if we know that a block is long, the probability for
-            # a missed splitting is higher. Therefore, we weaken the condition. The weakened condition
-            # leads to an improved score as per 21.03.2024.
-
-            if line_ends_block and ((is_line_long_enough and line_cuts_lefthandside_of_block) or weak_condition):
```
redur, line 149: Do you have any intuition why this is not needed any more? Is it because the bounding boxes of the lines start further right compared to the bounding boxes of the blocks?

src/stratigraphy/util/geometric_line_utilities.py
```diff
@@ -230,7 +232,7 @@ def merge_parallel_lines(lines: list[Line], tol: int = 8, angle_threshold: float
         if not merged:
             merged_lines.append(line)
     if merged_any:
-        return merge_parallel_lines(merged_lines, tol=tol)
+        return merge_parallel_lines(merged_lines, tol=tol, angle_threshold=angle_threshold)
```
redur, line 235: Ui, that's a good catch. You are right, I should be careful with default values.

redur (approved):

redur:
I approved the PR. My comments are more of informational nature.

src/stratigraphy/util/description_block_splitter.py
stijnvermeeren-swisstopo: Yes, correct. The difference is small, though maybe using the more "local" context is slightly more reliable, as we prevent a single incorrectly identified text line further away from potentially influencing the condition.

src/stratigraphy/util/description_block_splitter.py
stijnvermeeren-swisstopo: As discussed on Teams: I don't have a clear intuition, but I did not have a clean intuition either for why this "weak condition" was needed in the first place. A plausible explanation is that in some cases, the bounding box of the material descriptions is incorrectly identified. Now we don't use this bounding box anymore, but rather the more local context of the individual text lines just before and after. Maybe it is this change that removes the need for this additional "weak condition".

PR #20, opened 2024-04-02
Feat/fix material descr rect

redur:
Sometimes the material_descr_rect contains lines/blocks that are no material descriptions.

This PR attempts at removing those lines. Either by directly adjusting the condition to add lines to a material_descr_rect or by checking if a block contains a legend.

Improves F1 from 87.7% to 88.0%. Improvements are present for all documents.

**Document Level Metrics**
|    | document_name    |    F1_new |   Number Elements_new |   Number wrong elements_new |        F1 |   Number Elements |   Number wrong elements |
|---:|:-----------------|----------:|----------------------:|----------------------------:|----------:|------------------:|------------------------:|
|  0 | 683248004-bp.pdf | 0.878049  |                    21 |                           5 | 0.780488  |                21 |                       9 |
|  1 | 268125447-bp.pdf | 0.6875    |                    18 |                          10 | 0.5       |                18 |                      18 |
|  2 | 268124125-bp.pdf | 0.0266667 |                    20 |                          73 | 0.0243902 |                20 |                      80 |
|  3 | 683248005-bp.pdf | 0.896552  |                    14 |                           3 | 0.764706  |                14 |                       8 |
|  4 | 268124307-bp.pdf | 0.96      |                    25 |                           2 | 0.941176  |                25 |                       3 |
|  5 | 268124391-bp.pdf | 1         |                    23 |                           0 | 0.938776  |                23 |                       3 |

**Discussion**
This solution is not really nice. The chosen expressions are obtained by looking at the documents. I believe this is already some overfitting. Also we are still faced with doing the same exercise for other languages.

I would prefer to perform some supervised learning exercise on blocks or lines, and classify whether a block/line is a material description. But that poses some problems such as labeling (which we can easily perform), creating test, eval, train data sets, and overfitting. --> i.e. we'd need more data and have to label some of them.

Also, we'd still need to include some logic to find potential material description blocks in the first place.

Let's discuss in the daily tomorrow.


redur:
Reimplementing the legend detection and merging main results to a score improvement from 87.9% to 88.2%.

stijnvermeeren-swisstopo (approved):
I like the legend detection. Relatively simple logic, that seems too work accurately for a range of different inputs.

I also agree with your assessment regarding the excluding_expressions implementation. It does not feel very robust yet, and will likely not generalize very well to new data (unless we continually keep fine-tuning the list of expressions). We can keep it like this for now, but I would revisit this logic as soon as we have more data to work with (to check how well it generalizes, and how easily we can support additional languages), or as soon as we have a more comprehensive line merging output (to see if we can use visual features to reliably identify the upper limit of the material descriptions; e.g. a wide line above the highest depth value).

PR #21, opened 2024-04-03
Feat/configure pipeline with arguments

redur:
Created a CLI interface to run our main script.

We now have the option to define input and output directories.
We can specify single files for the pipeline to be run on.
We can specify what drawings are performed (bounding boxes for predictions and lines).


src/stratigraphy/main.py
```diff
@@ -37,474 +25,67 @@
...
+    depth intervals. The input directory should contain pdf files with boreholes data. The algorithm can deal
+    with borehole profiles of multiple pages.
+
+    Args:\n
+        input_directory (Path): The directory containing the pdf files.\n
+        ground_truth_path (Path): The path to the ground truth file json file.\n
+        out_directory (Path): The directory to store the evaluation results.\n
+        predictions_path (Path): The path to the predictions file.\n
+        skip_draw_predictions (bool, optional): Whether to skip drawing predictions on pdf pages. Defaults to False.\n
+        draw_lines (bool, optional): Whether to draw lines on pdf pages. Defaults to False.\n
```
stijnvermeeren-swisstopo, line 87: Isn't there a way to generate the docstrings correctly without these explicit "\n"?

src/stratigraphy/main.py
```diff
@@ -513,28 +94,40 @@ def perform_matching(directory: Path, **params: dict) -> dict:
...
     temp_directory.mkdir(parents=True, exist_ok=True)
 
+    # if a file is specified instead of an input directory, copy the file to a temporary directory and work with that.
+    if input_directory.is_file():
+        if (temp_directory / "single_file").is_dir():
+            shutil.rmtree(temp_directory / "single_file")
+
+        Path.mkdir(temp_directory / "single_file")
+        shutil.copy(input_directory, temp_directory / "single_file")
+        input_directory = temp_directory / "single_file"
```
stijnvermeeren-swisstopo, line 110: I'm not a big fan of this temporary directory. Wouldn't it be possible to instead support passing a single file as well as a directory to the `perform_matching` and `evaluate_matching` methods.

stijnvermeeren-swisstopo (changes requested):
A few ideas/comments.

Also check the additional proposal/comments in https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/22

src/stratigraphy/main.py
redur: The problem is related to how click provides the docstring to stdout when prompting `boreholes-extract-materials --help`. Without the `\n` it's all on one line which is not very readable.

redur: But let me dig a bit.

pyproject.toml
```diff
@@ -7,6 +7,9 @@ requires-python = ">=3.10"
 dependencies = [
 ]
 
+[project.scripts]
+boreholes-extract-materials = "stratigraphy.main:start_pipeline"
```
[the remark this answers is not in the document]
redur: I am open to rename the command. Actually, I am not super happy with it.

I initially thought of sub commands. Some structure like:

`boreholes-extract materials`
`boreholes-extract metadata`, etc.

And then maybe `boreholes-extract all`

In my logic a material is assigned to a depth interval, and thus it's materials that we extract. Or maybe better layers?

What do you think?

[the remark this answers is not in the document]
stijnvermeeren-swisstopo: Yes, I prefer borehole-extract-layers. We can use that for now.

src/stratigraphy/main.py
redur: I thought of refactoring the code such that we loop over all files inside of the `start_pipeline` function. Then we can easily catch the case where we provide a file instead of a folder.

Also, the function calls would be nicer, as we specifically pass objects to the functions instead of paths. I believe that would make the code more robust, as we only had to check once if the files comply with what's expected. (which we're not doing at all at the moment).

What do you think about that suggestion?

stijnvermeeren-swisstopo: Sounds good, feel free to go ahead with that refactoring.

src/stratigraphy/main.py
```diff
@@ -37,474 +25,68 @@
...
-def perform_matching(directory: Path, **params: dict) -> dict:
-    """Perform the matching of text blocks with depth intervals.
-
+@click.command()
+@click.option(
+    "-i",
+    "--input_directory",
+    type=click.Path(exists=True, path_type=Path),
+    default=DATAPATH / "Benchmark",
+    help="Path to the input directory.",
```
stijnvermeeren-swisstopo, line 34: Update documentation to clarify that this can also be the path of a single PDF file.

stijnvermeeren-swisstopo (changes requested):
Looks good apart from these minor details.

stijnvermeeren-swisstopo (approved):

PR #22, opened 2024-04-03
Feat/configure pipeline with arguments stijn

stijnvermeeren-swisstopo:
I was thinking more in lines of this implementation, where we would not need to execute the line detection twice when we want the visualisation, but would rather reuse the results of the line detection that is executed anyway.

However, I'm not fully happy with the way we are mixing up extraction logic with debugging visualisations, nor with the need for passing additional parameters to several methods.

@redur what do you think. Do you see a better way of reusing the line detection results, whilst maintaining a clearer "separation of concerns" in the code?


redur:
I like the idea of performing line detection only once. But I would probably take the line detection out of process_page and not mix the extraction with the visualization part.

What do you think?

I would even go a step further and not loop over files in the directory inside `perform_matching` but inside `start_pipeline`. All following function calls shall take a doc or page object.

stijnvermeeren-swisstopo:
@redur Sounds good. Do you want to have a go at implementing that?

PR #23, opened 2024-04-03
Refactor extraction pipeline.2

redur:
Here the discussed proposition.

I think we can extend the thought even further and include the draw_predictions earlier. That would require that we load the ground truth earlier as well. I keep this for the future as the current implementation currently already serves for discussion.

### DIFF
--- a/src/stratigraphy/line_detection.py
+++ b/src/stratigraphy/line_detection.py
@@ -87,7 +87,7 @@ def extract_lines(page: fitz.Page, line_detection_params: dict) -> list[Line]:
     return lines
 
 
-def draw_lines_on_pdfs(filename: str, page: fitz.Page, geometric_lines: list[Line]):
+def draw_lines_on_page(filename: str, page: fitz.Page, geometric_lines: list[Line]):
     """Draw lines on pdf pages and stores them as artifacts in mlflow.
 
     Note: now the function draw_lines_on_pdfs may not even be needed any more.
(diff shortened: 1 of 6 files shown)


src/stratigraphy/main.py
```diff
@@ -110,8 +111,34 @@ def start_pipeline(
...
+                        )
+
+                        predictions[filename][f"page_{page_number}"] = {
+                            "layers": layer_predictions,
+                            "depths_materials_column_pairs": depths_materials_column_pairs,
+                        }
+
+                        if draw_lines:
+                            logger.info("Drawing lines on pdf pages.")
+                            draw_lines_on_pdfs(filename, page, geometric_lines)
```
redur, line 140: maybe rename to: `draw_lines_on_page`

src/stratigraphy/line_detection.py
```diff
@@ -88,26 +87,19 @@ def extract_lines(page: fitz.Page, line_detection_params: dict) -> list[Line]:
     return lines
 
 
-def draw_lines_on_pdfs(input_directory: Path, line_detection_params: dict):
+def draw_lines_on_pdfs(filename: str, page: fitz.Page, geometric_lines: list[Line]):
     """Draw lines on pdf pages and stores them as artifacts in mlflow.
 
+    Note: now the function draw_lines_on_pdfs may not even be needed any more.
```
redur, line 93: Not sure it's worth having a separate function that essentially only has two lines of code. The MLFLow warning can be thrown somewhere else as well.

redur:
Now single files work as well without the temporary directory.

stijnvermeeren-swisstopo (changes requested):

redur:
Thank you. Comments are addressed.

stijnvermeeren-swisstopo (approved):

PR #29, opened 2024-04-05
Feat/add unit tests

redur:
Implementation of (some) unit tests for the stratigraphy package.

Comes with updated pytest action that returns coverage with a GitHub bot.
Unit tests for the modules
- geometric_line_utitlities
- dataclasses
- texblock
- interval
- find_description

This is a starting point for me. I believe these cove some important functionalities.

Let's start a discussion about what other module should be tested. And whether current tests are sufficient for the tested modules.

### DIFF
--- a/tests/test_find_depth_columns.py
+++ b/tests/test_find_depth_columns.py
@@ -4,15 +4,15 @@ import fitz
 import pytest
 from stratigraphy.util.depthcolumnentry import DepthColumnEntry
 from stratigraphy.util.find_depth_columns import depth_column_entries, find_depth_columns, find_layer_depth_columns
-from stratigraphy.util.line import DepthInterval, TextLine
+from stratigraphy.util.line import TextLine, TextWord
 
 
 def test_depth_column_entries():  # noqa: D103
     all_words = [
-        TextLine([DepthInterval(fitz.Rect(0, 0, 5, 1), "10.00m")]),
-        TextLine([DepthInterval(fitz.Rect(0, 2, 5, 3), "20.0m")]),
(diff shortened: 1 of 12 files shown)


redur:
Evt. könnte man `depthcolumn.py` noch testen.

tests/test_find_depth_columns.py
```diff
@@ -0,0 +1,78 @@
...
+    entries = [
+        DepthColumnEntry(fitz.Rect(0, 0, 5, 1), 10.0),
+        DepthColumnEntry(fitz.Rect(0, 2, 5, 3), 20.0),
+        DepthColumnEntry(fitz.Rect(0, 4, 5, 5), 30.0),
+        DepthColumnEntry(fitz.Rect(0, 6, 5, 7), 40.0),
+        DepthColumnEntry(fitz.Rect(0, 8, 5, 9), 50.0),
+    ]
+
+    columns = find_depth_columns(entries, all_words_find_depth_column)
+    assert len(columns) == 0, "There should be 0 columns as the above is a perfect arithmetic progression"
```
redur, line 59: I am not so sure about that.

stijnvermeeren-swisstopo (approved):
I added a couple of test cases for interval.py and textblock.py, that have increased the coverage a little bit.

We could certainly go much further, because there is a lot of functionality that would still benefit from tests, especially the rather complex logic in depthcolumn.py that you already mentioned. However, I think that we should not get too bogged down with writing unit tests. Instead, we can work on adding more unit tests as we are working on improving the logic of the the respective files/classes.

PR #32, opened 2024-04-09
Feat/include new profiles

redur:
Added new set of borehole profiles, and with it, multilingual support.

- Adjusted GroundTruth to work with new format. --> old format will not work any more. There is a functionality in the notebook that converts old ground_truth.json into the new one.
- Added language detection.
- Translated including and excluding expression into french.
- Also comes with the EDA based on a different branch. Probably should be dropped here. Kept for the conversion script and the generation of test/train/eval splits. Can be dropped after the review.

### DIFF
--- a/src/stratigraphy/util/language_detection.py
+++ b/src/stratigraphy/util/language_detection.py
@@ -1,0 +1,43 @@
+"""This module contains functionalities for language detection of a document."""
+
+import fitz
+from langdetect import detect
+from langdetect.lang_detect_exception import LangDetectException
+
+
+def extract_text_from_document(doc: fitz.Document) -> str:
+    """Extracts and processes text from a document.
+
+    Args:
+        doc (fitz.Document): The document to extract text from.
+
+    Returns:
+        str: The extracted and processed text.
+    """
+    text = ""
+    for page in doc:
+        text += page.get_text()
+    text = text.replace("\n", " ")
+
+    # remove all numbers and special characters from text
+    return "".join(e for e in text if (e.isalnum() or e.isspace()) and not e.isdigit())
+
+
+def detect_language_of_document(doc: fitz.Document) -> str:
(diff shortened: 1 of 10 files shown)


src/stratigraphy/util/language_detection.py
```diff
@@ -0,0 +1,43 @@
...
+    Returns:
+        str: The extracted and processed text.
+    """
+    text = ""
+    for page in doc:
+        text += page.get_text()
+    text = text.replace("\n", " ")
+
+    # remove all numbers and special characters from text
+    return "".join(e for e in text if (e.isalnum() or e.isspace()) and not e.isdigit())
```
stijnvermeeren-swisstopo, line 23: Does this keep accented letter such as öäüéàè that might be important for detecting German/French, or only ASCII?

stijnvermeeren-swisstopo (commented):
I would indeed remove the notebook and not merge it into the main code base.

What's the easiest way for us to share our train/validation/test splits? Do you recommend that I execute the notebook myself, using the exact same random seed, or is there a more convenient way?

src/stratigraphy/util/language_detection.py
redur: Good question. I'll have to check.

redur: öäüéàè and the like are kept.

stijnvermeeren-swisstopo (approved):
LGTM.

For now, it's good to have the convert_ground_script here for reproducibility, but in future we can probable remove that script again, as we won't need to deal with the old format at all anymore?

redur:
> LGTM.
> 
> For now, it's good to have the convert_ground_script here for reproducibility, but in future we can probable remove that script again, as we won't need to deal with the old format at all anymore?

I agree with this statement.

PR #37, opened 2024-04-26
Feat/prepare for label studio

redur:
Adjusts the code-base such that it can be used in combination with the [label-studio-ml-backend repository](https://github.com/redur/label-studio-ml-backend).

**Contains**
- a script to convert pdf files to png files --> this is needed as we can only import images to label-studio
- Make swissgeol-boreholes-dataextraction installable (setup.py, add dependencies, adjust pyproject.toml)
- Fix some path assignments such that it can be used in a docker container
- Adjust predictions objects to include information required by label-studio

### DIFF
--- a/src/stratigraphy/util/interval.py
+++ b/src/stratigraphy/util/interval.py
@@ -59,10 +59,6 @@ class AnnotatedInterval:
     end: float
     background_rect: fitz.Rect
 
-    @property
-    def line_anchor(self) -> fitz.Point:
-        return fitz.Point(0, 0)
-
 
 class BoundaryInterval(Interval):
     """Class for boundary intervals.
--- a/src/stratigraphy/util/interval.py
+++ b/src/stratigraphy/util/interval.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import abc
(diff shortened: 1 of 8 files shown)


src/stratigraphy/util/interval.py
```diff
@@ -50,6 +51,19 @@ def to_json(self):
...
+class AnnotatedInterval:
+    """Class for annotated intervals."""
+
+    start: float
+    end: float
+    background_rect: fitz.Rect
+
+    @property
+    def line_anchor(self) -> fitz.Point:
+        return fitz.Point(0, 0)
```
stijnvermeeren-swisstopo, line 64: What's the point of this?

redur: That's a good question. I need to check myself. I believe the property was required, but we don't need it for AnnotatedInterval. I would tend to drop it and see if we really need line_anchor for the `AnnotatedInterval` class.

redur: Method is removed. Thank you for spotting it.

stijnvermeeren-swisstopo (approved):

PR #38, opened 2024-04-26
Feat/exact line merging

redur:
This branch contains a first version of a quadtree implementation for efficient line merging.

The algorithm works on most pdf's, but for some there is an error that I do not yet understand. (Basically, a point can't be added to the tree structure).

I will continue working on this after I returned.

### DIFF
--- a/src/stratigraphy/util/geometric_line_utilities.py
+++ b/src/stratigraphy/util/geometric_line_utilities.py
@@ -6,35 +6,13 @@ from itertools import combinations
 from math import atan, cos, pi, sin
 
 import numpy as np
-import quads
 from numpy.typing import ArrayLike
 
-from stratigraphy.util.dataclasses import IndexedLines, Line, Point
+from stratigraphy.util.dataclasses import Line, Point
+from stratigraphy.util.linesquadtree import LinesQuadTree
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
-
-
-def deduplicate_lines(lines: list[Line]) -> list[Line]:
-    """Deduplicate lines by merging lines that are close to each other.
-
-    Args:
-        lines (list[Line]): The lines to deduplicate.
-
-    Returns:
-        list[Line]: The deduplicated lines.
-    """
-    deduplicated_lines = []
-    for line in lines:
-        if not _check_if_present(deduplicated_lines, line):
-            deduplicated_lines.append(line)
-    return deduplicated_lines
-
-
-def _check_if_present(lines: list[Line], line: Line) -> bool:
-    return any(
-        value.start.distance_to(line.start) < 0.1 and value.end.distance_to(line.end) < 0.1 for value in lines
-    )  # we are on a pixel grid and 0.1 is a reasonable threshold
 
 
 def drop_vertical_lines(lines: list[Line], threshold: float = 0.1) -> ArrayLike:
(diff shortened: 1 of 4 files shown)


stijnvermeeren-swisstopo:
@redur I allowed myself to try this out (installing quads manually, since it's not yet in the Conda environment file).

Initially, I ran into "max recursion depth exceeded" errors as well. They seemed to pop up rather randomly, not always on the same files. My maximum recursion depth was set to 3000, which should normally be enough for our purposes.

I'm not quite sure yet what caused this excessive recursion, but I think it might have had to do with the fact that `merge_parallel_lines_quadtree` got called from scratch after every individual merge. I've now replaced this with an approach that uses a queue of lines (to which the merged lines are added, so that they are also considered). Now I don't run into any recursion depth errors anymore.
I've also added some code to deal with the case when `None` is returned when merging two lines.

The pipeline now runs successfully, and only takes a little longer than before (ca. 10 minutes compared to ca. 8 minutes).

However, I think that there is still some bug somewhere, as a visual inspection of the lines in the PNG artifacts shows, that there are still quite a lot of lines that appear to fulfill the merge conditions, but are nevertheless drawn separately.

redur:
> @redur I allowed myself to try this out (installing quads manually, since it's not yet in the Conda environment file).
> 
> Initially, I ran into "max recursion depth exceeded" errors as well. They seemed to pop up rather randomly, not always on the same files. My maximum recursion depth was set to 3000, which should normally be enough for our purposes.
> 
> I'm not quite sure yet what caused this excessive recursion, but I think it might have had to do with the fact that `merge_parallel_lines_quadtree` got called from scratch after every individual merge. I've now replaced this with an approach that uses a queue of lines (to which the merged lines are added, so that they are also considered). Now I don't run into any recursion depth errors anymore. I've also added some code to deal with the case when `None` is returned when merging two lines.
> 
> The pipeline now runs successfully, and only takes a little longer than before (ca. 10 minutes compared to ca. 8 minutes).
> 
> However, I think that there is still some bug somewhere, as a visual inspection of the lines in the PNG artifacts shows, that there are still quite a lot of lines that appear to fulfill the merge conditions, but are nevertheless drawn separately.

Thank you for working on this. I will have a look too.

redur:
Line merging using quad trees is now improved.

- Recursion depth had to be put to 10'000. The problem is building of the quad tree. I am not 100% sure why the tree gets so deep.
- Visual inspection now shows no obvious cases where the merge condition is met, but lines are not merged.


Furthermore, shall we remove "all other" merging functionalities to keep the code lean?

redur:
The quadtree implementation of the line merging now works as intended.

The scores on the Zurich dataset slightly decreases. I checked all cases and with the exception of one, the "new" result is not worse than the "old". The change in the scores are often in cases where lot's of mistakes happen and now the splits are slightly different. The majority of the decrease from 88.2% to 87.9% is one single case where one split is not performed any more.

redur:
Update: I pinged the maintainer of quads if he would support a contribution to remove points from a quad tree.

src/stratigraphy/util/geometric_line_utilities.py
```diff
@@ -381,3 +225,101 @@ def _are_parallel(line1: Line, line2: Line, angle_threshold: float) -> bool:
...
+            line = indexed_lines.hashmap[line_index]
+            # Get all points in the quadtree that are close to the line
+            min_x = min(line.start.x, line.end.x)
+            max_x = max(line.start.x, line.end.x)
+            min_y = min(line.start.y, line.end.y)
+            max_y = max(line.start.y, line.end.y)
+            bb = quads.BoundingBox(min_x - tol, min_y - tol, max_x + tol, max_y + tol)
+
+            points = qtree.within_bb(bb)
+            merge_candidates_idx = get_merge_candidates_from_points(points, indexed_lines)
```
stijnvermeeren-swisstopo, line 274: I'm still struggling to understand why the recursive calls of this method are necessary. I indeed see that we are still merging more lines after the first call of this method. However, I do not understand why.

Indeed, if I replace these lines with the follow (less efficient, but intuitively equivalent) code, then all the merges already happen in the first iteration.
```
            merge_candidates_idx = []
            for idx, merge_line in indexed_lines.hashmap.items():
                if (
                    min_x - tol < merge_line.end.x <= max_x + tol and min_y - tol < merge_line.end.y <= max_y + tol
                ) or (
                    min_x - tol < merge_line.start.x <= max_x + tol and min_y - tol < merge_line.start.y <= max_y + tol
                ):
                    merge_candidates_idx.append(idx)
```
In other words, there seem to be valid merge candidates, that are not immediately found by the `qtree.within_bb` method, at least not until the next iteration, after we've reconstructed the quad tree.

Is this an issue with the quads library, or with our usage of quad trees? I don't fully understand it yet. Still needs more debugging...

stijnvermeeren-swisstopo: So indeed, the current quads implementation runs into problems when you try to use it with points that have the same coordinates but potentially different data. The problems start as soon as there are more than 4 points with the same coordinates, as this is the default capacity value (https://github.com/toastdriven/quads/blob/master/quads.py#L279). You can increase the capacity value when constructing the quad tree, but you never know for sure how large the value has to be, in order not to run into errors, and also this could make the quad tree less efficient.

However, in general, it should be no problem to write a quad tree implementation that supports our use case. Either you check the capacity based on unique coordinates, rather than based on total number of points. Or you make the data associated with each point a set (containing all the indexes of lines associated with that point), to which elements can be added or removed dynamically. The latter approach might actually be possible with the current quads library, I'll give it a try.

stijnvermeeren-swisstopo: The idea from the previous comment seems to work :). See https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/43
The implementation still needs to be documented + cleaned up a little, but I think I prefer it over the current workaround/recursion.

stijnvermeeren-swisstopo (changes requested):
I created a PR on top of this one, that addresses all the issues that I've identified with this PR. Also, I've removed the inefficient checks for duplicate lines, so the runtime is now about the same as before. I'm now happy with my proposed implementation (with the exception of the lack of unit tests). Please review that PR. When you're happy with my proposal as well, then we could merge both PRs. -> https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/43

PR #40, opened 2024-05-06
Improve Depth Intervals

redur:
In this PR we:
- Added visualization whether depth interval is correct.
- Improved the depth_column recognition by weakening the noise_count criterion.

**Result**
Depth_interval_accuracy improves from 42% to 62.3% while keeping F1 score at 49.3%.

The F1 score for the Zurich dataset decreased to 87.1% from 87.9%.

For future implementation we might want to reconsider how we split the material description blocks with regards to the criterion that we expect the same number of blocks as depth intervals.

Leaving this PR for review now.

### DIFF
--- a/src/stratigraphy/util/find_depth_columns.py
+++ b/src/stratigraphy/util/find_depth_columns.py
@@ -124,12 +124,15 @@ def find_layer_depth_columns(entries: list[DepthColumnEntry], all_words: list[Te
     ]
 
 
-def find_depth_columns(entries: list[DepthColumnEntry], all_words: list[TextLine]) -> list[BoundaryDepthColumn]:
+def find_depth_columns(
+    entries: list[DepthColumnEntry], all_words: list[TextLine], depth_column_params: dict
+) -> list[BoundaryDepthColumn]:
     """Construct all possible BoundaryDepthColumn objects from the given DepthColumnEntry objects.
 
     Args:
         entries (list[DepthColumnEntry]): All found depth column entries in the page.
         all_words (list[TextLine]): All words in the page.
+        depth_column_params (dict): Parameters for the BoundaryDepthColumn objects.
 
     Returns:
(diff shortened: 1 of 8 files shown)


redur:
Latest commit improves overall depth interval accuracy from 42% to 51%.

redur:
Latest commit improves depth_interval_accuracy **from 51% to 57.6%** while F1 score drops from **50.0% to 49.8%** on the "new" dataset.

redur:
Latest commit improves depth_interval_accuracy to **62.3%** while keeping F1 score at **49.3%**.

Leaving this PR for review now.

redur:
Check docstrings for parameter.

src/stratigraphy/util/find_depth_columns.py
```diff
@@ -24,22 +24,21 @@ def depth_column_entries(all_words: list[TextLine], include_splits: bool) -> lis
 
     def value_as_float(string_value: str) -> float:  # noqa: D103
         # OCR sometimes tends to miss the decimal comma
-        parsed_text = re.sub(r"^([0-9]+)([0-9]{2})", r"\1.\2", string_value)
+        parsed_text = re.sub(r"^-?([0-9]+)([0-9]{2})", r"\1.\2", string_value)
```
redur, line 27: This change includes numerical values that start with a `-`.

stijnvermeeren-swisstopo (approved):
Looks good to me. Thanks as well for the improved documentation and configurability.

I'm a little worried about two examples from the Zürich dataset (267123120-bp and 699248001-bp; screenshots attached to the ticket on Jira), where a completely incorrect depth column is now accepted, due to the more relaxed "noise checks". However, I think that a "noise check" might not the best approach to fix such cases anyway, but that instead, a different type of check might be more robust. For that reason, I'm ok with accepting these as edge cases for the time being, and creating a follow-up ticket to address them later.

PR #41, opened 2024-05-11
Feat/coordinate extraction

redur:
Simple Regex Extractor for coordinates in borehole profiles.

Work in progress...

Note: As of now, latitude and longitude may be confused by the extractor.

**TODO:**
- [x] Extract coordinates in the CH1903+/LV95 form as well
- [ ]  ~~Convert coordinates from CH1903/LV03 to CH1903+/LV95~~ 
- [x] Write tests
- [x] Make sure longitude and latitude are correctly identified
- [x] Write some documentation in form of comments
- [x] Update Readme and add info about what is extracted.

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -14,7 +14,6 @@ dependencies:
   - scikit-learn==1.4.0
   - langdetect==1.0.9
   - boto3==1.34.82
-  - fuzzywuzzy==0.18.0
   - regex==2024.5.10
   - pip
 # dev dependencies
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -14,6 +14,7 @@ dependencies:
   - scikit-learn==1.4.0
   - langdetect==1.0.9
   - boto3==1.34.82
+  - fuzzywuzzy==0.18.0
   - pip
 # dev dependencies
   - matplotlib==3.8.0
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -15,6 +15,7 @@ dependencies:
   - langdetect==1.0.9
   - boto3==1.34.82
   - fuzzywuzzy==0.18.0
+  - regex==2024.5.10
   - pip
 # dev dependencies
   - matplotlib==3.8.0
--- a/src/stratigraphy/util/coordinate_extraction.py
+++ b/src/stratigraphy/util/coordinate_extraction.py
@@ -21,6 +21,8 @@ class CoordinateEntry:
     def __post_init__(self):
         if self.coordinate_value is not None:
             string_representation = str(self.coordinate_value)
+            if len(string_representation) < 6:  # leading zeros are lost in the int conversion
+                string_representation = "0" * (6 - len(string_representation)) + string_representation
             self.first_entry = string_representation[:3]
             self.second_entry = string_representation[3:]
 
@@ -61,12 +63,11 @@ class Coordinate(metaclass=abc.ABCMeta):
     def from_json(input: dict):
         east = input["E"]
         north = input["N"]
-        print(f"{east=} {north=}")
-        if east > 2e6 and east > 1e7:
+        if east > 2e6 and east < 1e7:
             return LV95Coordinate(
                 CoordinateEntry(coordinate_value=east - 2e6), CoordinateEntry(coordinate_value=north - 1e6)
             )
-        elif east > 1e5 and east < 1e6:
+        elif east < 1e6:
             return LV03Coordinate(CoordinateEntry(coordinate_value=east), CoordinateEntry(coordinate_value=north))
         else:
             raise ValueError("Invalid coordinates format")
@@ -87,8 +88,8 @@ class LV95Coordinate(Coordinate):
 
     def to_json(self):
         return {
-            "E": self.east.coordinate_value + 2e7,
-            "N": self.north.coordinate_value + 1e7,
+            "E": self.east.coordinate_value + 2e6,
+            "N": self.north.coordinate_value + 1e6,
         }
 
 
--- a/src/stratigraphy/util/coordinate_extraction.py
+++ b/src/stratigraphy/util/coordinate_extraction.py
@@ -1,0 +1,86 @@
+"""This module contains the CoordinateExtractor class."""
+
+import re
+
+import fitz
+import regex
+
+
+class CoordinateExtractor:
+    """Extracts coordinates from a PDF document."""
+
+    def __init__(self, document: fitz.Document):
+        """Initializes the CoordinateExtractor object.
+
+        Args:
+            document (fitz.Document): A PDF document.
+        """
+        self.doc = document
+        self.coordinate_keys = ["Koordinaten", "Koordinate", "coordinates", "coordinate"]
+
+    def find_coordinate_key(self, text: str, allowed_operations: int = 3) -> str:
+        """Finds the location of a coordinate key in a string of text.
+
+        Args:
+            text (str): Arbitrary string of text.
+            allowed_operations (int, optional): The maximum number of allowed operations to consider a key contained
+                                                in text. Defaults to 3.
+
+        Returns:
+            str: The coordinate key found in the text.
+        """
+        matches = []
+        for key in self.coordinate_keys:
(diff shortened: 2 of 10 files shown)


environment-dev.yml
```diff
@@ -14,6 +14,8 @@ dependencies:
   - scikit-learn==1.4.0
   - langdetect==1.0.9
   - boto3==1.34.82
+  - fuzzywuzzy==0.18.0
```
stijnvermeeren-swisstopo, line 17: where is this used?

src/stratigraphy/util/draw.py
```diff
@@ -61,6 +63,16 @@ def draw_predictions(predictions: list[FilePredictions], directory: Path, out_di
...
 
 
+def draw_coordinates(page: fitz.Page, coordinates: list[float]) -> None:
+    """Draw the coordinates on a pdf page.
+
+    Args:
+        page (fitz.Page): The page to draw on.
+        coordinates (list[float]): List of coordinates to draw.
+    """
+    page.add_freetext_annot(fitz.Rect([5, 5, 100, 250]), f"Coordinates: {coordinates}")
```
stijnvermeeren-swisstopo, line 73: I'm not a huge fan of this way of including the coordinates on the PNG image. If they are printed on a dark background, then you can hardly read them. And this does not give any information about where on the page the coordinates where found, which is kind-of the whole point of creating this PNG visualisation in the first place. It's not a major issue, we can also leave it like this for now, but I think we should think about how to better visualise the coordinates (and other extracted metadata) in the future.

stijnvermeeren-swisstopo (changes requested):
The results look pretty good for a first implementation!

A few thoughts:
- The logic with the regexes is fairly complex, and would probably warrant more unit tests.
- Do we want to add benchmarking code for verifying the correctness of the extracted coordinates here, or do we do that in a separate PR/ticket?
- In the example below (A202.png), the extractor "hallucinates" some coordinates from numbers that are not coordinates at all. Is there an easy way to fix this, or shall we create a follow-up ticket for that?

image

stijnvermeeren-swisstopo:
Another interesting case, where a phone number is interpreted as coordinates... (A57.pdf)

image

stijnvermeeren-swisstopo:
V69.png: coordinates are correctly recognized, but the first three digits are ignored, instead the digits after the decimal point are incorrectly included (not sure why the coordinates are listed here with such an exaggerated accuracy).

image

src/stratigraphy/util/coordinate_extraction.py
```diff
@@ -0,0 +1,271 @@
...
+            try:
+                # get the substring that contains the coordinate information
+                coord_substring = self.get_coordinate_substring(text)
+                coordinate_string = self.get_coordinates_text(coord_substring)[0]
+            except IndexError:
+                # if that doesnt work, try to directly detect coordinates in the text
+                try:
+                    coordinate_string = self.get_coordinates_text(text)[0]
+                except IndexError:
+                    return None
```
redur, line 235: raise info / warning

src/stratigraphy/util/coordinate_extraction.py
```diff
@@ -0,0 +1,271 @@
...
+            coordinate_string = y_coordinate_string[0] + " / " + x_coordinate_string[0]
+        except IndexError:  # no coordinates found
+            try:
+                # get the substring that contains the coordinate information
+                coord_substring = self.get_coordinate_substring(text)
+                coordinate_string = self.get_coordinates_text(coord_substring)[0]
+            except IndexError:
+                # if that doesnt work, try to directly detect coordinates in the text
+                try:
+                    coordinate_string = self.get_coordinates_text(text)[0]
```
redur, line 233: Maybe we can make this case stricter.

environment-dev.yml
redur: Should be removed.

src/stratigraphy/util/coordinate_extraction.py
```diff
@@ -0,0 +1,271 @@
...
+    def find_coordinate_key(self, text: str, allowed_operations: int = 3) -> str:
+        """Finds the location of a coordinate key in a string of text.
+
+        This is is useful to reduce the text within which the coordinates are searched. If the text is too large
+        false positive (found coordinates that are no coordinates) are more likely.
+
+        Args:
+            text (str): Arbitrary string of text.
+            allowed_operations (int, optional): The maximum number of allowed operations to consider a key contained
+                                                in text. Defaults to 3.
```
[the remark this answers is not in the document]
redur: As discussed. Let's address this in a follow up once we have a way to evaluate coordinate extraction.

src/stratigraphy/util/draw.py
redur: As discussed. Let's keep it for now.

src/stratigraphy/util/coordinate_extraction.py
```diff
@@ -0,0 +1,271 @@
...
+        return (
+            f"longitude: 2.{self.longitude.first_entry}.{self.longitude.second_entry}, "
+            f"latitude: 1.{self.latitude.first_entry}.{self.latitude.second_entry}"
+        )
+
+    def to_json(self):
+        return {
+            "longitude": f"2.{self.longitude.first_entry}.{self.longitude.second_entry}",
+            "latitude": f"1.{self.latitude.first_entry}.{self.latitude.second_entry}",
+        }
```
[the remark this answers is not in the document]
redur: It's now with numbers.

src/stratigraphy/util/coordinate_extraction.py
```diff
@@ -0,0 +1,283 @@
...
+    def to_json(self):
+        pass
+
+    @staticmethod
+    def from_json(input: dict):
+        east = input["E"]
+        north = input["N"]
+        if east > 2e6 and east < 1e7:
+            return LV95Coordinate(
+                CoordinateEntry(coordinate_value=east - 2e6), CoordinateEntry(coordinate_value=north - 1e6)
```
stijnvermeeren-swisstopo, line 70: Why subtract 2e6 / 1e6 here, instead of just leaving the 7-digit coordinate values?

src/stratigraphy/util/coordinate_extraction.py
```diff
@@ -0,0 +1,283 @@
...
+
+COORDINATE_ENTRY_REGEX = "(?:[12][\.\s']{0,2})?\d{3}[\.\s']{0,2}\d{3}\.?\d?"
+
+
+@dataclass
+class CoordinateEntry:
+    """Dataclass to represent a coordinate entry."""
+
+    first_entry: str | None = None
+    second_entry: str | None = None
```
stijnvermeeren-swisstopo, line 20: Why do we need these at all? If it's just for string formatting purposes, then we should rather use [format specifications](https://docs.python.org/3/library/string.html#format-specification-mini-language) (use `,` and then replace with `'`) or some string operations (slicing and `"'".join()` ) instead.

src/stratigraphy/util/coordinate_extraction.py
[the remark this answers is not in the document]
stijnvermeeren-swisstopo: Let's at least add a comment to the code about this, though.

stijnvermeeren-swisstopo (changes requested):

src/stratigraphy/util/coordinate_extraction.py
redur: If we switch to purely int based coordinates I see 7 digits numbers favoured.

src/stratigraphy/util/coordinate_extraction.py
redur: We usually extract two strings from the text and I wondered where we should process the strings. I decided it could be prettier to do it inside the class, than in the extraction code. That's why I kept the first and second entry, but introduced the coordinate value. The formatting was a nice side effect. Also, it was neat to deal with both coordinate formats the same way.

Either way, maybe the code gets a bit simpler if we slice and join in the extraction part.

stijnvermeeren-swisstopo (approved):

PR #44, opened 2024-05-16
Feat/coordinate evaluation

redur:
Added evaluation of coordinates. 
Addressed some pending items in PR #46.
- Change the representation of CoordinateEntry.
- Added some comments.

### DIFF
--- a/src/stratigraphy/util/coordinate_extraction.py
+++ b/src/stratigraphy/util/coordinate_extraction.py
@@ -9,7 +9,7 @@ import regex
 
 logger = logging.getLogger(__name__)
 
-COORDINATE_ENTRY_REGEX = "(?:[12][\.\s']{0,2})?\d{3}[\.\s']{0,2}\d{3}\.?\d?"
+COORDINATE_ENTRY_REGEX = r"(?:[12][\.\s']{0,2})?\d{3}[\.\s']{0,2}\d{3}\.?\d?"
 
 
 @dataclass
@@ -74,7 +74,7 @@ class LV95Coordinate(Coordinate):
     north: CoordinateEntry
 
     def __repr__(self):
-        return f"E: 2'{self.east}, " f"N: 1'{self.north}"
+        return f"E: {self.east}, " f"N: {self.north}"
 
     def to_json(self):
         return {
@@ -129,7 +129,7 @@ class CoordinateExtractor:
         """
         matches = []
         for key in self.coordinate_keys:
-            match = regex.search(r"\b(" + key + "){e<" + str(allowed_errors) + "}\s", text, flags=regex.IGNORECASE)
+            match = regex.search(r"\b(" + key + "){e<" + str(allowed_errors) + r"}\s", text, flags=regex.IGNORECASE)
             if match:
                 matches.append((match.group(), sum(match.fuzzy_counts)))
 
@@ -197,7 +197,7 @@ class CoordinateExtractor:
             list: A list of matched coordinates.
         """
         return regex.findall(
-            r"[XY]?[=:\s]{0,2}" + COORDINATE_ENTRY_REGEX + ".*[XY]?[=:\s]{0,2}" + COORDINATE_ENTRY_REGEX,
+            r"[XY]?[=:\s]{0,2}" + COORDINATE_ENTRY_REGEX + r".*[XY]?[=:\s]{0,2}" + COORDINATE_ENTRY_REGEX,
             text,
         )
 
--- a/src/stratigraphy/util/coordinate_extraction.py
+++ b/src/stratigraphy/util/coordinate_extraction.py
@@ -113,16 +113,21 @@ class CoordinateExtractor:
         self.coordinate_keys = ["Koordinaten", "Koordinate", "coordinates", "coordinate", "coordonnés", "coordonnes"]
         # TODO: extend coordinate keys with other languages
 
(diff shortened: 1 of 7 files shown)


src/stratigraphy/util/coordinate_extraction.py
```diff
@@ -9,33 +9,25 @@
 
 logger = logging.getLogger(__name__)
 
-COORDINATE_ENTRY_REGEX = "(?:[12][\.\s']{0,2})?\d{3}[\.\s']{0,2}\d{3}\.?\d?"
+COORDINATE_ENTRY_REGEX = r"(?:[12][\.\s']{0,2})?\d{3}[\.\s']{0,2}\d{3}\.?\d?"
```
redur, line 12: Some refactoring to not have too much duplicated code.

src/stratigraphy/util/coordinate_extraction.py
```diff
@@ -66,13 +58,12 @@ def from_json(input: dict):
...
         if east > 2e6 and east < 1e7:
-            return LV95Coordinate(
-                CoordinateEntry(coordinate_value=east - 2e6), CoordinateEntry(coordinate_value=north - 1e6)
-            )
+            return LV95Coordinate(CoordinateEntry(coordinate_value=east), CoordinateEntry(coordinate_value=north))
         elif east < 1e6:
             return LV03Coordinate(CoordinateEntry(coordinate_value=east), CoordinateEntry(coordinate_value=north))
         else:
-            raise ValueError("Invalid coordinates format")
+            logger.warning(f"Invalid coordinates format. Got E: {east}, N: {north}")
```
redur, line 65: This happens once in the whole dataset.

src/stratigraphy/util/coordinate_extraction.py
```diff
@@ -128,23 +113,23 @@ def __init__(self, document: fitz.Document):
...
 
         This is is useful to reduce the text within which the coordinates are searched. If the text is too large
         false positive (found coordinates that are no coordinates) are more likely.
 
         Args:
             text (str): Arbitrary string of text.
-            allowed_operations (int, optional): The maximum number of allowed operations to consider a key contained
-                                                in text. Defaults to 3.
+            allowed_errors (int, optional): The maximum number of errors to consider a key contained
+                                            in text. An error is a character that is different. Defaults to 3.
```
redur, line 125: Is this sufficient for you? There is not soo much documentation on that.

The best is to search for fuzzy matching with regex. There are some stackoverflow threads talking about it, but the official documentation is not very helpful.

src/stratigraphy/util/predictions.py
```diff
@@ -274,6 +286,43 @@ def evaluate(self, ground_truth_layers: list):
...
+            ):
+                ground_truth_east = int(metadata_ground_truth["coordinates"]["E"]) - 2e6
+                ground_truth_west = int(metadata_ground_truth["coordinates"]["N"]) - 1e6
+            else:
+                ground_truth_east = int(metadata_ground_truth["coordinates"]["E"])
+                ground_truth_west = int(metadata_ground_truth["coordinates"]["N"])
+
+            if (int(self.metadata.coordinates.east.coordinate_value) == ground_truth_east) and (
+                int(self.metadata.coordinates.north.coordinate_value) == ground_truth_west
+            ):
```
stijnvermeeren-swisstopo, line 321: Do we get better results if we allow the absolute value of the difference to be <= 2? Like this, we might catch some cases where conversions between LV03 and LV95 were already applied in the ground truth.

src/stratigraphy/util/predictions.py
```diff
@@ -274,6 +286,43 @@ def evaluate(self, ground_truth_layers: list):
...
+    def evaluate_metadata(self, metadata_ground_truth: dict):
+        """Evaluate the metadata of the file against the ground truth.
+
+        Note: For now coordinates is the only metadata extracted and evaluated for.
+
+        Args:
+            metadata_ground_truth (dict): The ground truth for the file.
+        """
+        if self.metadata.coordinates is None:
+            self.metadata_is_correct["coordinates"] = False
```
stijnvermeeren-swisstopo, line 298: Maybe in future we should relax this a little bit, and say that it's ok when no coordinates are extracted, as long as there are no coordinates in the ground truth either.
Currently this will not make a difference, as there are always coordinates in the ground truth. However, then are not always contained in the PDFs. So maybe in future, we will want to set the ground truth to None when the coordinates cannot be extracted from the PDF at all.

src/stratigraphy/util/coordinate_extraction.py
stijnvermeeren-swisstopo: But these "errors" are the same as for Levenshtein distance, is that correct? In that case, it would be clearer to write something like: "The maximum Levenshtein distance from a true coordinate key to a phrase in the text, for that phrase to be accepted as a valid coordinate key. Defaults to 3 (guestimation; no optimisation done yet)."

stijnvermeeren-swisstopo (changes requested):
Yes, this indeed solved the issues that I still had with the other PR. :)

A few minor questions regarding this one...

src/stratigraphy/util/predictions.py
redur: Deal with this in a follow up ticket.

stijnvermeeren-swisstopo (approved):

PR #45, opened 2024-05-21
Feat: Material Block recognition for Geneva Layout

redur:
Includes functionalities to recognize material description blocks for the Geneva borehole profile layout. The code detects so called layer index columns which indicate the presence of a material description rect as well as their position marks how to split the material description rect into material blocks. Note: Some visualization code is part of this commit which should be dropped in a follow up commit.

### DIFF
--- a/config/matching_params.yml
+++ b/config/matching_params.yml
@@ -46,21 +46,24 @@ material_description:
   fr:
     including_expressions:
       - sol
-      - végétal
+      - végétal  # remove accents generally; ocr might be wrong
       - dallage
       - terre
       - bitume
       - bitumineux
-      - grave d'infrastructure
+      - grave d'infrastructure  # what happens if we remove this?
       - sable
       - limon
       - gravier
       - asphalte
-      - humus
+      - humus  # hummus maybe?
       - brun
       - gris
+      - grise
       - mou
       - dur
+      - dure
+      - ferme
       - racine
       - revetement
       - pierre
@@ -68,9 +71,19 @@ material_description:
       - beton
       - craie
       - marne
-      - materiau de base
+      - materiau
       - matrice sableuse
-      - enrobé
+      - enrobé  # accent --> check what happens if it's removed
+      - terrain
+      - remblais
+      - remblai
+      - molasse
+      - phase
+      - formations
+      - limoneuse
+      - argileuse
+      - argileux
+      - mousse
     excluding_expressions:
       - monsieur
       - fin
--- a/src/stratigraphy/extract.py
+++ b/src/stratigraphy/extract.py
@@ -50,7 +50,7 @@ def process_page(page: fitz.Page, geometric_lines, language: str, **params: dict
     for x0, y0, x1, y1, word, block_no, line_no, _word_no in fitz.utils.get_text(page, "words"):
         rect = fitz.Rect(x0, y0, x1, y1) * page.rotation_matrix
         text_word = TextWord(rect, word)
-        words.append(TextLine([text_word]))
+        words.append(text_word)
         key = f"{block_no}_{line_no}"
         if key not in words_by_line:
             words_by_line[key] = []
@@ -105,7 +105,7 @@ def process_page(page: fitz.Page, geometric_lines, language: str, **params: dict
             if material_description_rect:
                 layer_identifier_pairs.append((layer_identifier_column, material_description_rect))
 
-        # Obtain the best pair. In contrast do depth columns, there only ever is one layer index column per page.
+        # Obtain the best pair. In contrast to depth columns, there only ever is one layer index column per page.
         if layer_identifier_pairs:
             layer_identifier_pairs.sort(key=lambda pair: score_column_match(pair[0], pair[1]))
             layer_identifier_column, material_description_rect = layer_identifier_pairs[-1]
--- a/src/stratigraphy/extract.py
+++ b/src/stratigraphy/extract.py
@@ -11,11 +11,14 @@ from stratigraphy.util.dataclasses import Line
 from stratigraphy.util.depthcolumn import DepthColumn
 from stratigraphy.util.find_description import (
     get_description_blocks,
-    get_description_blocks_from_layer_index,
+    get_description_blocks_from_layer_identifier,
     get_description_lines,
 )
 from stratigraphy.util.interval import BoundaryInterval, Interval
-from stratigraphy.util.layer_index_column import find_layer_index_column, find_layer_index_column_entries
+from stratigraphy.util.layer_identifier_column import (
+    find_layer_identifier_column,
+    find_layer_identifier_column_entries,
+)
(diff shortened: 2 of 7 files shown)


config/matching_params.yml
```diff
@@ -46,31 +46,44 @@ material_description:
...
+      - molasse
+      - phase
+      - formations
+      - limoneuse
+      - argileuse
+      - argileux
+      - mousse
     excluding_expressions:
       - monsieur
       - fin
```
redur, line 89: I checked our expressions with a native French speaker who also has experience with NLP with the French language. We looked at borehole profiles together and extended the list. Also, he suggested to remove all accents because OCR frequently makes errors there.

src/stratigraphy/extract.py
```diff
@@ -81,14 +87,63 @@ def process_page(page: fitz.Page, geometric_lines, language: str, **params: dict
...
+                    page, layer_index_column.rect() * page.derotation_matrix, color=fitz.utils.getColor("blue")
+                )
+            for block in blocks:
+                fitz.utils.draw_rect(page, block.rect * page.derotation_matrix, color=fitz.utils.getColor("red"))
+            fitz.utils.draw_rect(
+                page, material_description_rect * page.derotation_matrix, color=fitz.utils.getColor("blue")
+            )
+            page.parent.save(DATAPATH / "_temp" / "output.pdf", garbage=4, deflate=True, clean=True)
+
+            return predictions, json_filtered_pairs
```
redur, line 138: For now, I just exit here if there's a layer index column.

Some thoughts:
I believe it should be possible to consider a layer index column a special case of a depth column, and continue the "normal" way. I believe this will be "cleaner" to move on.

src/stratigraphy/util/find_description.py
```diff
@@ -24,16 +24,75 @@ def get_description_lines(lines: list[TextLine], material_description_rect: fitz
...
+        )
+
+        matched_block = matching_blocks(description_lines, line_index, next_layer_index)
+        line_index += sum([len(block.lines) for block in matched_block])
+        blocks.extend(matched_block)
+
+    return blocks
+
+
+def matching_blocks(all_lines: list[TextLine], line_index: int, next_layer_index: TextLine | None) -> list[TextBlock]:
```
redur, line 66: Could become a method of `LayerIndexColumn`.

src/stratigraphy/util/layer_index_column.py
```diff
@@ -0,0 +1,162 @@
...
+        """
+        return (
+            rect.x0 <= self.rect().x0
+            and self.rect().x1 <= rect.x1
+            and rect.y0 <= self.rect().y0
+            and self.rect().y1 <= rect.y1
+        )
+
+    def noise_count(self, words):
+        return 0
```
stijnvermeeren-swisstopo, line 96: method not used?

src/stratigraphy/util/find_description.py
stijnvermeeren-swisstopo: If `next_layer_index` is not just an index but of type `TextLine | None` , then probably `next_layer` is a better name?

src/stratigraphy/util/find_description.py
```diff
@@ -24,16 +24,75 @@ def get_description_lines(lines: list[TextLine], material_description_rect: fitz
...
+        layer_index_entries (list[TextLine]): The layer index entries.
+        description_lines (list[TextLine]): All lines constituting the material description.
+
+    Returns:
+        list[TextBlock]: The blocks of the material description.
+    """
+    blocks = []
+    line_index = 0
+    for layer_index_idx, _layer_index in enumerate(layer_index_entries):
+        # don't allow a layer above depth 0
```
stijnvermeeren-swisstopo, line 53: redundant comment?

src/stratigraphy/util/layer_index_column.py
```diff
@@ -0,0 +1,162 @@
...
+    This regular expression will match strings like "1)", "2)", "a)", "b)", "1a4)", "6de)", etc.
+
+    Args:
+        all_words (ist[TextLine]): The words to search for layer index columns.
+
+    Returns:
+        list: The layer index column entries.
+    """
+    entries = []
+    for line in sorted(all_words, key=lambda line: line.rect.y0):
```
stijnvermeeren-swisstopo, line 116: Better name it "word" rather than line, even though it is of type TextLine (maybe this class should be renamed as well)?

src/stratigraphy/util/layer_index_column.py
```diff
@@ -0,0 +1,162 @@
...
+        regex = re.compile(r"\b[\da-z-]+\)")
```
stijnvermeeren-swisstopo, line 117: Could we avoid false positives by requiring that this word is at the start of a line and/or there are not other words immediately to the left of it?

src/stratigraphy/extract.py
stijnvermeeren-swisstopo: Agreed

stijnvermeeren-swisstopo (changes requested):
I think that this is already a pretty decent first implementation, though obviously there is room for some simplification of the code (sharing more code between the "index column" and "depth column" implementations.

I would propose to merge this (after fixing the minor comments) and then to improve the code in a next iteration.

src/stratigraphy/util/find_description.py
redur: I did not create a class object `LayerIndexEntry` similar to the entries of depthcolumns. The entries of the LayerIndex are simply TextLine objects.

What I mean here is the next layer index "entry". Therefore I called it next_layer_index. Does that make sense?

Moving forward and considering LayerIndexColumn as a special case of a DepthColum we might want to create the "LayerIndexEntry" object as well, and then this will be clearer. What do you think?

src/stratigraphy/util/layer_index_column.py
redur: You are right!

src/stratigraphy/util/layer_index_column.py
redur: Are lines sometimes more than just a word? If not, then I am in favour of renaming the TextLine object.

And agree with naming it word over line.

src/stratigraphy/util/layer_index_column.py
redur: You would like to check other "lines"/"words" as well? The `\b` marks a word boundary and the query only matches if the sequence is at the start of a word.

stijnvermeeren-swisstopo: Yes, there are quite a few examples such as "(ca. 10 cm)" where "cm)" would be matched currently. I would like to avoid some of those. But that can be an idea for a future improvement. Maybe just add a code comment about it for now.

stijnvermeeren-swisstopo (approved):

PR #46, opened 2024-05-21
Feat/extend geneva layouts

redur:
- Refactored process_page such that LayerIdentifierColumn is considered a special case of a DepthColumn.
- Included depth interval extraction for Deriaz layouts. --> increased depth interval accuracy by a few percent.
- Some additional minor changes.

### DIFF
--- a/src/stratigraphy/util/find_depth_columns.py
+++ b/src/stratigraphy/util/find_depth_columns.py
@@ -6,7 +6,7 @@ import fitz
 
 from stratigraphy.util.depthcolumn import BoundaryDepthColumn, LayerDepthColumn
 from stratigraphy.util.depthcolumnentry import DepthColumnEntry, LayerDepthColumnEntry
-from stratigraphy.util.line import TextWord
+from stratigraphy.util.line import TextWord, TextLine
 
 
 def depth_column_entries(all_words: list[TextWord], include_splits: bool) -> list[DepthColumnEntry]:
@@ -21,12 +21,6 @@ def depth_column_entries(all_words: list[TextWord], include_splits: bool) -> lis
     Returns:
         list[DepthColumnEntry]: The extracted depth column entries.
     """
-
-    def value_as_float(string_value: str) -> float:  # noqa: D103
-        # OCR sometimes tends to miss the decimal comma
-        parsed_text = re.sub(r"^-?([0-9]+)([0-9]{2})", r"\1.\2", string_value)
-        return abs(float(parsed_text))
-
     entries = []
     for word in sorted(all_words, key=lambda word: word.rect.y0):
         try:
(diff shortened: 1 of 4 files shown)


src/stratigraphy/util/line.py
```diff
@@ -24,6 +24,12 @@ def __init__(self, rect: fitz.Rect, text: str):
     def __repr__(self) -> str:
         return f"TextWord({self.rect}, {self.text})"
 
+    def to_json(self):
+        return {
+            "text": self.text,
+            "rect": [self.rect.x0, self.rect.y0, self.rect.x1, self.rect.y1],
+        }
```
stijnvermeeren-swisstopo, line 31: is this used at all?

src/stratigraphy/util/layer_identifier_column.py
```diff
@@ -16,7 +40,7 @@ def __init__(self, entries: list[TextWord]):
         Args:
             entries (list[TextWord]): The entries corresponding to the layer indices.
         """
-        self.entries = entries
+        self.entries = [LayerIdentifierEntry(entry.rect, entry.text) for entry in entries]
```
stijnvermeeren-swisstopo, line 43: With this change, the types no longer match in the `add_entry` method.

src/stratigraphy/util/layer_identifier_column.py
```diff
@@ -16,7 +40,7 @@ def __init__(self, entries: list[TextWord]):
         Args:
             entries (list[TextWord]): The entries corresponding to the layer indices.
```
stijnvermeeren-swisstopo, line 41: Maybe rename this argument to "words", to clarify that at this point, we don't have actual "LayerIdentifierEntry" objects yet.

src/stratigraphy/util/find_depth_columns.py
```diff
@@ -37,27 +31,51 @@ def value_as_float(string_value: str) -> float:  # noqa: D103
...
+
+
+def extract_layer_depth_interval_entries(
+    input_string: str, line: TextWord | TextLine, require_start_of_string: bool = True
+) -> list[DepthColumnEntry]:
+    """Extracts two DepthColumnEntry obejcts from a string to represent a layer depth interval.
+
+    Args:
+        input_string (str): The string to extract the depth interval from.
+        line (TextLine): The line object containing the rectangle of the string.
```
stijnvermeeren-swisstopo, line 53: Types in the docstring don't match. Also, the value of `input_string` immediately gets overridden. And it can be confusing that `line` can be both a `TextLine` as well as a `TextWord`. 

Wouldn't it be sufficient to accept params `text: str` and `rect: fitz.Rect` instead of `input_string` and `line`.

Or alternatively, if we really want to have params that can be both `TextLine` as well as `TextWord`, then it would be cleaner to make them children of a common abstract parent class.

src/stratigraphy/util/layer_identifier_column.py
```diff
@@ -92,6 +116,48 @@ def is_contained(self, rect: fitz.Rect) -> bool:
...
+        Args:
+            block (TextBlock): The block to calculate the depth interval for.
+
+        Returns:
+            LayerDepthColumnEntry: The depth interval.
+        """
+        depth_entries = []
+        for line in block.lines:
+            try:
+                new_entries = extract_layer_depth_interval_entries(line.text, line, require_start_of_string=False)
```
stijnvermeeren-swisstopo, line 136: Can you explain why `require_start_of_string=False` is necessary here?

src/stratigraphy/util/find_depth_columns.py
```diff
@@ -37,27 +31,51 @@ def value_as_float(string_value: str) -> float:  # noqa: D103
                 value = value_as_float(match.group(1))
                 entries.append(DepthColumnEntry(word.rect, value))
             elif include_splits:
-                # support for e.g. "1.10-1.60m" extracted as a single word
```
stijnvermeeren-swisstopo, line 40: keep the comment

stijnvermeeren-swisstopo (changes requested):

src/stratigraphy/util/line.py
redur: Yes it is. When we save the depth column to the json, probably for the LayerIdentifierColumn. In any case, after merging your changes the code didn't run any more since TextWord did not have a .to_json method.

src/stratigraphy/util/layer_identifier_column.py
redur: There are example where the depth value is not at the start of a textline. 

Something like: "Remblais Heterogene: 0.00 - 0.5m"

redur: Add a comment to the code.

src/stratigraphy/util/line.py
redur: You are right. It was only used since add_entry added words to the LayerIdentifierColumn and not LayerIdentifierEntry. Now we manage without the to_json method.

src/stratigraphy/util/layer_identifier_column.py
```diff
@@ -92,6 +116,50 @@ def is_contained(self, rect: fitz.Rect) -> bool:
...
+            except ValueError:
+                pass
+
+        if depth_entries:
+            # Merge the sub layers into one depth interval.
+            start_idx = np.argmin([entry[0].value for entry in depth_entries])
+            end_idx = np.argmax([entry[1].value for entry in depth_entries])
+
+            start = depth_entries[start_idx][0]
+            end = depth_entries[end_idx][1]
```
stijnvermeeren-swisstopo, line 150: isn't this equivalent to
```
            start = min([entry[0].value for entry in depth_entries])
            end = max([entry[1].value for entry in depth_entries])
```
?

src/stratigraphy/util/find_depth_columns.py
```diff
@@ -38,26 +32,51 @@ def value_as_float(string_value: str) -> float:  # noqa: D103
...
 
+def value_as_float(string_value: str) -> float:  # noqa: D103
+    # OCR sometimes tends to miss the decimal comma
+    parsed_text = re.sub(r"^-?([0-9]+)([0-9]{2})", r"\1.\2", string_value)
+    return abs(float(parsed_text))
+
+
+def extract_layer_depth_interval_entries(
+    text: str, rect: fitz.Rect, require_start_of_string: bool = True
+) -> list[DepthColumnEntry]:
```
stijnvermeeren-swisstopo, line 49: Wouldn't it be more natural to return `LayerDepthColumnEntry | None` here, as we always either return an empty list, or a list with two entries (a start and an end) which is exactly what LayerDepthColumnEntry  represents.

stijnvermeeren-swisstopo (changes requested):

src/stratigraphy/util/layer_identifier_column.py
redur: Yes that's correct. The complexity stems from an older implementation with two lists.

PR #47, opened 2024-05-22
only check for layer identifiers in the first word of each line

stijnvermeeren-swisstopo:


redur:
I can see that this adds value, but unfortunately it is not fault proof either. 

See e.g.: 
- 5793.pdf (IdentifierColumn is in the description text).
- 7317.pdf (here there are only few entries not recognized)


I still propose that we merge your suggestion, as it seems to make the code more robust overall.

**5793**
image

**7317**
image

redur (approved):

stijnvermeeren-swisstopo:
Thanks @redur for these good observations! This seems to be a new layout type from a company "SOLAB F. Perrin S.A.", that is somewhere in between the Deriaz layouts and the "LayerDepthColumn" implementation. I've created a follow-up ticket for this: https://jira.swisstopo.ch/browse/LGVISIUM-40

PR #48, opened 2024-05-24
Feat/improve coordinate extraction

redur:
- Added additional metrics for coordinate extraction evaluation.
- Adjust extraction logic to ensure X & Y values are close together.
- Changed evaluation to allow small deviations due to approximate conversions between LV03 and LV95 coordinate systems.
- Added additional check for validity of extracted coordinates.

**Outcome:**
- (Almost) all cornercases are covered with the extraction improvement. Coordinate accuracy as well as coordinate precision increased. (i.e. listed cornercases in the jira ticket, and similar such cases are covered)
- There does not seem any need to use the pdf-coordinate (i.e. location of the text in the pdf) to determine the right coordinates.


redur:
Remaining mistakes:
3425.pdf. —> see below
A1150.pdf. —> OCR Issue
A1163.pdf —> typo on pdf
A1254.pdf. —> OCR issue
A185.pdf. —> strange (unexpected) coordinate format
A8425.pdf. —> OCR issue
A8432.pdf —> OCR issue


GroundTruth Wrong
A1141.pdf. —> typo in ground truth
A361.pdf —> deviation of 4
A494.pdf —> typo in ground truth
A7068.pdf —> deviation of 28 in ground truth
A7069.pdf —> deviation of 23
B542.pdf —> deviation of 62

3425.pdf --> The coordinates are written on a underlined dotted line. The dots from the line are recognized as "." in between the numbers. Weakening recognition statement leads to additional false positives (even though there is "X" and "Y" in the matched string)

I suggest we correct the ground truth. But do not allow larger deviations from the correct value.

stijnvermeeren-swisstopo:
This looks like a nice accuracy improvement!

The sentence "Adjust extraction logic to ensure X & Y values are close together." is not really accurate, if I understand correctly, as we are not checking that the are close together, but rather that they each fall within a specific range.

I'm not yet fully sure about the decision not to consider the location of the text in the PDF at all. You are right that with the additional constraints implemented here, consdering at the PDF coordinates does not really seem necessary to achieve a satisfactory accuracy on our data. On the other hand, I feel like this would still be a "cheap" way to give us some extra certainty about the extracted coordinates. Also, we might want to support other countries / coordinate systems in the future, where the coordinate values might be less constrained, and considering the PDF coordinates might become essential.
I'm not fully decided yet on whether I would like to implement this now, or keep this as an idea for the future. I'll review the results more after the weekend, and then make a decision.

src/stratigraphy/util/predictions.py
```diff
@@ -316,8 +317,8 @@ def evaluate_metadata(self, metadata_ground_truth: dict):
                 ground_truth_east = int(metadata_ground_truth["coordinates"]["E"])
                 ground_truth_west = int(metadata_ground_truth["coordinates"]["N"])
```
stijnvermeeren-swisstopo, line 318: `north`, nicht `west`

stijnvermeeren-swisstopo (changes requested):

redur:
> n the other hand, I feel like this would still be a "cheap" way to give us some extra certainty about the extracted coordinate

Thanks for the review! 
With "X and Y" close together I mean that we don't allow too much text (characters) in between X and Y values. There is an implicit "closeness" condition as the OCR engine will put text that is close in the pdf also close in the "textspace". Of course this is not fully bullet proof; but it works very well on all the different formats that we have in our data set.

Regarding "other" countries. Here I see the problem, that we have specific classes for our Coordinates with custom logic. E.g. adding the 1 and 2, switching the coordinates etc. I fear that another country would have to create their own coordinate class. Or we manage to parametrize the logic in the yaml file (which I believe would be hard).

Either way, probably it would be best to:
- Create a file that only contains coordinate classes
- Add documentation on how to add your own coordinate class
- A means to use the coordinate class, without the need to adjust the source code (maybe we leave this for the future and open an issue).

redur:
Improve comment about "X & Y are close"

redur:
Collect ideas regarding refactoring in a ticket.

PR #50, opened 2024-05-28
Feat/improve coordinate extraction review

stijnvermeeren-swisstopo:
* Also match non-ASCII single quotes in coordinates (necessary for digitally-born PDFs in Zurich dataset)
* Simplify/unify logic for matching coordinates in text and extracting coordinate values
* Iterate over all potential coordinate pairs that were found, and return the first valid one, instead of only checking the first one

### DIFF
--- a/src/stratigraphy/util/coordinate_extraction.py
+++ b/src/stratigraphy/util/coordinate_extraction.py
@@ -259,25 +259,25 @@ class CoordinateExtractor:
             text += page.get_text()
         text = text.replace("\n", " ")
 
-        # try to get the text by including X and Y
-        try:
-            x_values = [
-                int("".join(groups)) for groups in regex.findall(r"X[=:\s]{0,3}" + COORDINATE_ENTRY_REGEX, text)
-            ]
-            y_values = [
-                int("".join(groups)) for groups in regex.findall(r"Y[=:\s]{0,3}" + COORDINATE_ENTRY_REGEX, text)
-            ]
+        # Try to get the text by including explicit 'X' and 'Y' labels.
+        # In this case, we can allow for some whitespace in between the numbers.
+        # In some older borehole profile the OCR may recognize whitespace between two digits.
+        x_values = [int("".join(groups)) for groups in regex.findall(r"X[=:\s]{0,3}" + COORDINATE_ENTRY_REGEX, text)]
+        y_values = [int("".join(groups)) for groups in regex.findall(r"Y[=:\s]{0,3}" + COORDINATE_ENTRY_REGEX, text)]
+        # We are only checking the 1st x-value with the 1st y-value, the 2nd x-value with the 2nd y-value, etc.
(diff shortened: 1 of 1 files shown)


src/stratigraphy/util/coordinate_extraction.py
```diff
@@ -262,72 +261,40 @@ def extract_coordinates(self) -> list:
...
-                y_coordinate_string[0].replace(" ", "") + " / " + x_coordinate_string[0].replace(" ", "")
-            )
+            x_values = [
+                int("".join(groups)) for groups in regex.findall(r"X[=:\s]{0,3}" + COORDINATE_ENTRY_REGEX, text)
+            ]
+            y_values = [
+                int("".join(groups)) for groups in regex.findall(r"Y[=:\s]{0,3}" + COORDINATE_ENTRY_REGEX, text)
+            ]
+
+            coordinate_values = [(x_values[0], y_values[0])]
```
redur, line 271: Essentially, we're still just taking the first match. Did I get that correctly?

redur (commented):
I have just one question regarding the different matches on a pdf. 

Say our REGEX query has two matches. Will we consider both? You mentioned something in the daily that made me think we would look at all of them. But I don't see it in the code.

src/stratigraphy/util/coordinate_extraction.py
stijnvermeeren-swisstopo: In this case, yes, that's correct. Only lower down (line 286), when we use the results from the get_coordinate_pairs method, we now iterate through all potential matches, until a suitable one is found.

I guess that we could do the same here as well, though I did not come across an example where it would make a difference.

redur: If we already iterate through the coordinate values further down, I would propose we do it here too.

Basically 
```suggestion
            coordinate_values = [(x_values[i], y_values[i]) for i in range(min([len(x_values), len(y_values)])) ]
```

What do you think? It should not change anything as we'll check the first entry first for valid coordinates.

stijnvermeeren-swisstopo: Can we rely on the y_values and x_values corresponding with each other? Or should we really be checking all combinations of them (e.g. also the 2nd x_value and the 3rd y_value)?

redur: We can't be sure. But we have seen that the OCR engine extracts text in the right order (or at least text that is close in pdf space is close in text space). Therefore I would argue, that it makes sense to take the implicit ordering given by the OCR engine.

However, there could be cornercases where this is not true. But since we demand two coordinate values explicitly preceeded by 'X' and 'Y' the possibilities are slim for something going wrong.

My comment is rather, if we allow for multiple values further down, why don't we do this here? I don't think there are many cases where this ever makes a difference.

PR #52, opened 2024-05-31
Feat/improve is valid

redur:
**Improve is_valid criterion.**

The noise check within the is_valid criterion is now adjusted to check
short depth columns (i.e. few entries) more strictly than longer depth columns (i.e. more entries).
This is achieved by applying a quadratic behavior onto the number of entries.

F1 improvement on Zurich dataset by 0.6%.
Similar F1 on geoquat dataset.

### DIFF
--- a/src/stratigraphy/util/find_depth_columns.py
+++ b/src/stratigraphy/util/find_depth_columns.py
@@ -26,6 +26,8 @@ def depth_column_entries(all_words: list[TextWord], include_splits: bool) -> lis
         try:
             input_string = word.text.strip().replace(",", ".")
             regex = re.compile(r"^-?\.?([0-9]+(\.[0-9]+)?)[müMN\\.]*$")
+            # numbers such as '.40' are not supported. The reason is that sometimes the OCR
+            # recognizes a '-' as a '.' and we just ommit the leading '.' to avoid this issue.
             match = regex.match(input_string)
             if match:
                 value = value_as_float(match.group(1))
--- a/src/stratigraphy/util/find_depth_columns.py
+++ b/src/stratigraphy/util/find_depth_columns.py
@@ -25,7 +25,7 @@ def depth_column_entries(all_words: list[TextWord], include_splits: bool) -> lis
     for word in sorted(all_words, key=lambda word: word.rect.y0):
         try:
             input_string = word.text.strip().replace(",", ".")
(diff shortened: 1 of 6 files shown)


src/stratigraphy/util/find_depth_columns.py
```diff
@@ -25,7 +25,7 @@ def depth_column_entries(all_words: list[TextWord], include_splits: bool) -> lis
     for word in sorted(all_words, key=lambda word: word.rect.y0):
         try:
             input_string = word.text.strip().replace(",", ".")
-            regex = re.compile(r"^-?([0-9]+(\.[0-9]+)?)[müMN\\.]*$")
+            regex = re.compile(r"^-?\.?([0-9]+(\.[0-9]+)?)[müMN\\.]*$")
```
redur, line 28: We now support numbers such as `.40` that sometimes occur.

src/stratigraphy/util/language_detection.py
```diff
@@ -36,8 +36,8 @@ def detect_language_of_document(doc: fitz.Document) -> str:
     try:
         language = detect(text)
     except LangDetectException:
-        language = "de"
+        language = "de"  # TODO: default language should be read from config
```
redur, line 39: This is bothering me for the moment. Right now you need to adjust the code to extend the extraction to other languages. This should be doable from the config files. I believe there are other places where language is hard-coded in form of keywords. (e.g. coordinate extraction)

 I will open an issue for it.

src/stratigraphy/util/find_depth_columns.py
stijnvermeeren-swisstopo: The input `.40` is now extracted with value `40`, not as `0.40`. Is that really what we want?

I would also suggest adding this as a test case in `test_find_depth_columns.py`.

stijnvermeeren-swisstopo (changes requested):

src/stratigraphy/util/find_depth_columns.py
redur: I checked again, and actually it's rather that sometimes a '-' is recognized as a '.' in older borehole profiles that have this "handwritten style". Then the behavior is totally desired.

But I also found an occurrence of '.80'. See here: A531.pdf

For our dataset, it is for now better to use the current behaviour.

stijnvermeeren-swisstopo (approved):

PR #54, opened 2024-05-31
Feat/remove duplicated dependency declaration

redur:
All dependencies are now directly managed in the pyproject.toml file.

Please consider the following:
Since we're developing a python package that is intended to use in combination with other python code and packages, we should keep the dependency restrictions as weak as possible. On the production side on the other hand, we want to fix all dependencies to fix the behavior of the production service.

We now made the shift from a fixed environment to declaring dependencies for the swissgeol-boreholes-dataextraction package. Therefore I tried to lift as many restrictions as possible. I believe we could further lift restrictions. I.e. do we really need python>3.10?
As we're the only users for now it is not an issue, but something to consider in the future. We also need to be careful, every time we notice some dependency issues, we have to adapt our restrictions in pyproject.toml.

Another note: I kept the restrictions for the dev dependencies rather fixed (I don't call them dev any more, but rather split them into test,lint,experiment_tracking & visualize. I haven't wrapped my head around if it really makes sense to have that many.
Test can be used for the testing CI.
Lint can be used for the lint CI.
Experiment tracking should be installed for local development (alongside with test and lint).
Visualize was something we used in a notebook, there's no need for a general user to install visualize. We could even think of removing this extra dependency alltogether.


pyproject.toml
```diff
@@ -12,18 +12,37 @@ dependencies = [
...
-    "PyYAML==6.0.1",
-    "langdetect==1.0.9",
+    "pathlib",
+    "python-dotenv",
+    "scikit-learn>=1.4.0",
+    "click>=8.0.0",
+    "PyYAML>=6.0.1",
+    "langdetect>=1.0.9",
+    "regex",
+    "amazon-textract-textractor",
```
redur, line 22: remove textractor

pyproject.toml
```diff
@@ -12,18 +12,37 @@ dependencies = [
...
     "backoff",
-    "PyMuPDF==1.23.26",
+    "PyMuPDF>=1.23.26",
     "opencv-python-headless",
-    "quads==1.1.0",
+    "quads>=1.1.0",
 ]
 
+[project.optional-dependencies]
```
redur, line 28: For the optional dependencies, we don't need to be that strict. I.e. we don't need to weaken the version as this is very much related to development and not for production purposes.

redur:
Since we look at the pyproject.toml file in this PR, I open the discussion about the package name.
Would it make sense to rename the package to` boreholes-dataextraction`? Or maybe `swisstopo-boreholes-dataextraction`? We could also think about shorter names.

Further, do we plan to publish the package on pypi at some point?

pyproject.toml
```diff
@@ -12,18 +12,36 @@ dependencies = [
...
+test = [
+    "pytest==8.1.1",
+    "pytest-cov==5.0.0"
+]
+lint = [
+    "pre-commit==3.6.2",
+]
+experiment_tracking = [
```
stijnvermeeren-swisstopo, line 36: I had to change to this `experiment-tracking` (with hyphen instead of underscore) to make it work. (It seems that Python automatically normalizes these names to use hyphens instead of underscores.)

README.md
```diff
@@ -13,34 +13,22 @@ The coordinate types LV95 as well as the older LV03 are supported. More informat
...
-    If you are developing and testing the code, you might want to use the dev environment instead:
+1. **Activate the virtual environment**
 
-    `conda activate boreholes-dev`
+    Activate your virtual environment and install the package. On unix systems this is
+    ```bash
+    python -m venv env
+    source env/bin/activate
+    pip install -e .[all]
+    ```
```
stijnvermeeren-swisstopo, line 31: I think it makes more sense to include these commands under the "installation" section, and here only say: activate the virtual environment if it is not active yet, using `source env/bin/activate`?

stijnvermeeren-swisstopo (changes requested):

stijnvermeeren-swisstopo:
> Since we look at the pyproject.toml file in this PR, I open the discussion about the package name. Would it make sense to rename the package to` boreholes-dataextraction`? Or maybe `swisstopo-boreholes-dataextraction`? We could also think about shorter names.
> 
> Further, do we plan to publish the package on pypi at some point?

1. We could use `swissgeol-boreholes-dataextraction`. Also the script command (currently "boreholes-extract-layers") should probably be generalised, as we no longer only extract layers.
2. No immediate plans yet, but it could be a logical step in the future. Would this choice have any influence on the development at this point in time?

redur:
> > Since we look at the pyproject.toml file in this PR, I open the discussion about the package name. Would it make sense to rename the package to` boreholes-dataextraction`? Or maybe `swisstopo-boreholes-dataextraction`? We could also think about shorter names.
> > Further, do we plan to publish the package on pypi at some point?
> 
> 1. We could use `swissgeol-boreholes-dataextraction`. Also the script command (currently "boreholes-extract-layers") should probably be generalised, as we no longer only extract layers.
> 2. No immediate plans yet, but it could be a logical step in the future. Would this choice have any influence on the development at this point in time?

1. For me it would make sense to divide the commands a bit further. Something like `boreholes-extract-all`, `boreholes-extract-layers`and `boreholes-extract-metadata`. This would go in line with the discussion we had about the architecture, and the possibility to redo certain steps given some constraints. I am fine with the name `swissgeol-boreholes-dataextraction`.
2. We are preparing for that step already using pyproject.toml. However, at the moment the package is meant to be used using the source code as we provide the config file with the source code. I feel we need a solution for that problem anyways.

redur:
Comments are addressed. Let me know if it's all good now.

stijnvermeeren-swisstopo (approved):

Issue #53, opened 2024-05-31
Available languages are hard coded.


redur:
Languages are hard coded in the code in multiple places.

1. In language_detection.py
2. In coordinate_extraction.py
3. other?

To resolve this issue, read in all language related information from the config files.

PR #56, opened 2024-06-03
Parametrize language in language detection and coordinate extraction.

redur:
PR to parametrize language in the config files.

I believe there should be nothing remaining in the code that specifies a language now. The language is completely abstracted in the matching_params.yml config file.


redur:
#53

PR #58, opened 2024-06-05
Use TextLine objects instead of the full plain text in coordinate extraction

stijnvermeeren-swisstopo:
- Separate the creation of TextLine object for a given PDF page into a separate method, that is used by both layer extraction and coordinate extraction.
- Instead of finding the coordinate key in the plain document text, find it in the TextLine objects, so that we also have the information of where it is located on the PDF page
- Instead of looking for coordinates within 100 characters of the coordinate key (in the defaut "reading order"), look for coordinates to the right and/or immediately below the coordinate key. This should be a little more robust / less dependent on they way (in which order) text is defined in the document.
- Instead of concatenating the text of all pages at the beginning of the coordinate extraction, we iterate over the pages and return on the first page where a valid coordinate was found. (This also will allow us to know on which page the coordinate was found, which is necessary for a future visualisation.)
- Improve logging a little bit.

The extraction of the coordinates values themselves still happens from plain text, not from the TextLine objects. We could improve on this in a future implementation, to use the TextLine objects everywhere. The main advantage of this would be, that we would also know where on the PDF page the coordinates were found, and we could use this information for the visualisation.

There are no changes in the KPIs for both the Zurich dataset and the Geoquat validation dataset.

### DIFF
--- a/src/stratigraphy/extract.py
+++ b/src/stratigraphy/extract.py
@@ -19,7 +19,7 @@ from stratigraphy.util.layer_identifier_column import (
     find_layer_identifier_column,
     find_layer_identifier_column_entries,
 )
-from stratigraphy.util.line import TextLine, TextWord
+from stratigraphy.util.line import TextLine
 from stratigraphy.util.textblock import TextBlock, block_distance
 from stratigraphy.util.util import (
     parse_and_remove_empty_predictions,
@@ -30,13 +30,13 @@ from stratigraphy.util.util import (
 logger = logging.getLogger(__name__)
 
 
-def process_page(page: fitz.Page, geometric_lines, language: str, **params: dict) -> list[dict]:
+def process_page(lines: list[TextLine], geometric_lines, language: str, **params: dict) -> list[dict]:
     """Process a single page of a pdf.
 
     Finds all descriptions and depth intervals on the page and matches them.
 
     Args:
-        page (fitz.Page): The page to process.
+        lines (list[TextLine]): all the text lines on the page.
         geometric_lines (list[Line]): The geometric lines of the page.
         language (str): The language of the page.
         **params (dict): Additional parameters for the matching pipeline.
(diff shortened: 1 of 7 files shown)


src/stratigraphy/util/coordinate_extraction.py
```diff
@@ -247,47 +250,51 @@ def extract_coordinates(self) -> Coordinate | None:
...
-            else:
-                coordinate = LV03Coordinate(
-                    CoordinateEntry(east),
-                    CoordinateEntry(north),
-                )
-            if coordinate.is_valid():
-                return coordinate
-
-        logger.warning(f"Could not extract valid coordinates from {coordinate_values}")
+            text = page.get_text()
```
redur, line 254: This comment relates to the the docstring of the function:

Suggestion:
```Extracts the coordinates from a borehole profile.

Processes the borehole profile page by page and tries to find the coordinates in the respective text of the page.
```

src/stratigraphy/util/extract_text.py
redur, line 1: I am not sure whether it makes sense to create a new python file for this. Could we thematically group it to line.py as this is the code that creates the objects defined in line.py?

src/stratigraphy/extract.py
redur, line 1: I am fine with that change. I am just thinking if there is any use case where we'd need the page / doc object inside process page. But should that be the case we can always add the page object again.

redur:
This comes in handy with the changes I am thinking about for label studio with regards to coordinate extraction.

redur (approved):
Approved with some open questions.

src/stratigraphy/extract.py
stijnvermeeren-swisstopo: Good point, but at this point in time, I don't see a need for it.

Also, it feels more consistent now with the `geometric_lines`, which are derived from the page object in a similar way, but were already passed as a separate parameter to the `process_page` method.

src/stratigraphy/util/extract_text.py
stijnvermeeren-swisstopo: In general I'm more in favor of more files with fewer lines of code per file, and maybe group files into subdirectories/packages/modules when we start having too many files in one place.
But if you feel strongly about it, we could also put these in one file, they are certainly closely related.

redur: Follow up ticket regarding this is created on Jira.

PR #60, opened 2024-06-10
Feat/fixes for label studio conversion

redur:
Some fixes that are required to perform the conversion from LabelStudio annotations to ground truth.


stijnvermeeren-swisstopo (approved):
I did not test it with label studio, but everything looks reasonable.

PR #61, opened 2024-06-10
Coordinates with position 2

stijnvermeeren-swisstopo:
- Keep track of the bounding box and page number for the extracted coordinates, and include this information in the predictions JSON file
- Draw the bounding box of the extracted coordinates (if any) in the visualisation
- Minor refactorings
- Make unit test covering coordinate extraction a bit more complete
- Also match lowercase "x" and "y" labels
- Make `Coordinate.from_values` more flexible in accepting swapped LV95 coordinate values


redur (approved):
Looks good to me.
I did not run your code here, but everything makes sense. 
Good work, thank you!

PR #64, opened 2024-06-18
Update conversion from label studio to ground truth.

redur:
Include metadata (i.e. coordinates) in the instantiation of prediction objects from label studio and the subsequent conversion to ground truth.


stijnvermeeren-swisstopo (approved):
I tested it, and it seems to work well.

I kept all the code the same, but I moved all the method that are directly related to Label Studio from the `predictions.py` file to the `label_studio_annotation_to_ground_truth.py` script. This seems to be a cleaner separation of concerns. Like this, the `predictions.py` code is no longer concerned with the specifics structure of the Label Studio JSON export, nor with the specific string formatting of the coordinate and depth interval values used in Label Studio.

PR #65, opened 2024-06-20
Feat/improve depth entry recognition

redur:
Tune some parameters to improve depth entry recognition.
Introduce new checks:
- check if depths are strictly increasing.
- check if correcting a common OCR mistake makes a depth column valid.

Overall, more depth entries are now recognized. For the zurich dataset, even the layer detection is improved. For the geoquad dataset, the layer detection rate is approximately the same.

### DIFF
--- a/src/stratigraphy/benchmark/score.py
+++ b/src/stratigraphy/benchmark/score.py
@@ -147,7 +147,7 @@ def evaluate_borehole_extraction(predictions: dict, number_of_truth_values: dict
     coordinate_metrics, coordinate_document_level_metrics = evaluate_metadata(predictions)
     metrics = {**layer_metrics, **coordinate_metrics}
     document_level_metrics = pd.merge(
-        layer_document_level_metrics, coordinate_document_level_metrics, on="document_name"
+        layer_document_level_metrics, coordinate_document_level_metrics, on="document_name", how="outer"
     )
     return metrics, document_level_metrics
 
--- a/src/stratigraphy/util/depthcolumn.py
+++ b/src/stratigraphy/util/depthcolumn.py
@@ -11,7 +11,7 @@ from stratigraphy.util.dataclasses import Line
 from stratigraphy.util.depthcolumnentry import DepthColumnEntry, LayerDepthColumnEntry
 from stratigraphy.util.find_description import get_description_blocks
 from stratigraphy.util.interval import BoundaryInterval, Interval, LayerInterval
-from stratigraphy.util.line import TextLine
+from stratigraphy.util.line import TextLine, TextWord
 
 
 class DepthColumn(metaclass=abc.ABCMeta):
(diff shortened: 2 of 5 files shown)


src/stratigraphy/benchmark/score.py
```diff
@@ -147,7 +147,7 @@ def evaluate_borehole_extraction(predictions: dict, number_of_truth_values: dict
     coordinate_metrics, coordinate_document_level_metrics = evaluate_metadata(predictions)
     metrics = {**layer_metrics, **coordinate_metrics}
     document_level_metrics = pd.merge(
-        layer_document_level_metrics, coordinate_document_level_metrics, on="document_name"
+        layer_document_level_metrics, coordinate_document_level_metrics, on="document_name", how="outer"
```
redur, line 150: This is just fixing a bug where we only kept rows with coordinates in the document_level_metrics

src/stratigraphy/util/depthcolumn.py
```diff
@@ -262,14 +262,14 @@ def depth_intervals(self) -> list[BoundaryInterval]:
         return depth_intervals
 
     def significant_arithmetic_progression(self) -> bool:
-        if len(self.entries) < 6:
+        if len(self.entries) < 7:
```
redur, line 265: This simply improved the scores.

src/stratigraphy/util/depthcolumn.py
```diff
@@ -321,9 +321,16 @@ def is_valid(self, all_words: list[TextLine]) -> bool:
...
             return False
+        # Check if the entries are strictly increasing.
+        if not all(i.value < j.value for i, j in zip(self.entries, self.entries[1:], strict=False)):
+            return False
 
         corr_coef = self.pearson_correlation_coef()
-        return corr_coef and corr_coef > 0.99
+
+        return (
+            corr_coef and corr_coef > np.min([1.0382 - len(self.entries) * 0.01, 0.9985]) and corr_coef > 0.95
```
redur, line 331: Admittably, this is probably overfitted. Will be interesting to see the real scores ones we have a test set.

src/stratigraphy/util/depthcolumn.py
stijnvermeeren-swisstopo: Keeping the segment length at 6 and simplifying the threshold to `abs(scale_pearson_correlation_coef) >= 0.9999` regardless of segment length seems to work even better. (Implemented in my commit.)

stijnvermeeren-swisstopo (approved):
Thanks for the work!
The code for correcting OCR mistakes does not make a big difference yet, but offers more potential for further improvements in the future.

With the additional changes that I've implemented (which I believe make the logic slightly simpler as well), I now get the following numbers (overall F1 score):

Zurich dataset:
- main branch: 87.16%
- Renato's version: 87.91%
- Stijn's version: 87.90%

Geoquat validation dataset:
- main branch: 49.60%
- Renato's version: 49.43%
- Stijn's version: 49.53%

While I believe that some further optimization of the parameters involved might be possible, the improvements will be marginal, and probably not worth the effort. Therefore, I would propose to merge the PR like this.

PR #66, opened 2024-06-24
Improve depth entry recognition for cases such as '.80'.

redur:
And here the PR to include '.80' and the like as well.

### DIFF
--- a/src/stratigraphy/util/boundarydepthcolumnvalidator.py
+++ b/src/stratigraphy/util/boundarydepthcolumnvalidator.py
@@ -1,6 +1,7 @@
 """This module contains logic to validate BoundaryDepthColumn instances."""
 
 import dataclasses
+import re
 
 from stratigraphy.util.depthcolumn import BoundaryDepthColumn
 from stratigraphy.util.depthcolumnentry import DepthColumnEntry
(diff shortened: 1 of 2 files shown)


stijnvermeeren-swisstopo:
I made two changes:
- Apply individual corrections to a single depth value, instead of trying all possible corrections at once. For example, from a value of "40", we would like to try the possibilities "0.40" as well as "10", but with the original implementation, we would only try "0.10". This currently does not make a difference for our datasets, but I think it makes more sense to apply corrections individually, especially as we will likely add more possible corrections in the future.
- Allow for more than one depth value to be corrected in a depth column. This improves the results for 680244005-bp.pdf (where two corrections of the 4->1 type are needed), while for 16005.pdf the multiple corrections lead to a column of 3 different values to be incorrectly recognized as a depth column (though it does not make a difference for the F1 score, as this document scores 0% anyway). Maybe allowing a certain number of corrections that is proportional to the total number of entries in the depth column, would be most robust? However, with so few documents where this makes a difference at all, I don't really think that it's worthwhile implementing such a rule with a additional parameter that needs to be tuned. @redur what do you think, should we stick to allowing one correction per column for now, or shall we drop that limit?

redur:
> I made two changes:
> 
> * Apply individual corrections to a single depth value, instead of trying all possible corrections at once. For example, from a value of "40", we would like to try the possibilities "0.40" as well as "10", whereas with the original implementation, we would only try "0.10". This currently does not make a difference for our datasets, but I think it makes more sense to apply corrections individually, especially as we will likely add more possible corrections in the future.
> * Allow for more than one depth value to be corrected in a depth column. This improves the results for 680244005-bp.pdf (where two corrections of the 4->1 type are needed), while for 16005.pdf the multiple corrections lead to a column of 3 different values to be incorrectly recognized as a depth column (though it does not make a difference for the F1 score, as this document scores 0% anyway). Maybe allowing a certain number of corrections that is proportional to the total number of entries in the depth column, would be most robust? However, with so few documents where this makes a difference at all, I don't really think that it's worthwhile implementing such a rule with a additional parameter that needs to be tuned. @redur what do you think, should we stick to allowing one correction per column for now, or shall we drop that limit?

Thank you for the improvements!

I agree to your first point. This makes a lot of sense.

Regarding your second point, I initially did not want to do this as I had the opinion that "corrections" should be applied scarcely to minimize the risk of "correcting" something that was correct.

stijnvermeeren-swisstopo (approved):
Documentation was added as discussed.

PR #69, opened 2024-07-29
Close #LGVISIUM-51: Pip install Install compatibilty with zsh

dcleres:
Edited the README to make it possible to install the python packages from a zsh terminal


stijnvermeeren-swisstopo:
I tested this with bash, and it works as well.

However, doing a little googling, it seems that `pip install -e '.[all]'` is the more commonly suggested solution, and I would find that a bit easier to read as well compared to the backslashes. Is there any reason why you did not do it like that?
* https://github.com/mu-editor/mu/issues/852
* https://stackoverflow.com/questions/30539798/zsh-no-matches-found-requestssecurity
* https://www.stevemar.net/pip-and-zsh/

dcleres:
Thank you for having a look. I do not have any preference, to be honest. I have also seen both, and I agree that the backslashes do improve the readability.

stijnvermeeren-swisstopo (approved):

PR #72, opened 2024-08-21
Close #LGVISIUM-54: Groundwater extraction

dcleres:
The following code allows you to extract the groundwater level information from the files in the Zurich dataset and add this information to the metadata of the borehole profile. 

The current performance is: 

```
'groundwater_depth_accuracy': 0.24375
'groundwater_depth_precision': 0.5571428571428572
'groundwater_depth_recall': 0.302325581395348
 'groundwater_depth_f1': 0.39195979899497485
'groundwater_depth_tp': 39
'groundwater_depth_fp': 31
'groundwater_depth_fn': 90
```

During the work on this issue, it was also detected that some files have wrong labels. Therefore, the performance might actually be better, than the metrics are suggesting. 

The chosen approach was to use basic keyword matching to identify the location of the groundwater information and then to parse the text around this field. 

The current limitations of this code are: 
- When the groundwater is displayed as a logo only without the keyword, the groundwater level cannot be assessed
- When the text extraction did not find the keyword or the groundwater level, then it is also likely that the extraction is wrong


src/stratigraphy/groundwater/utility.py
```diff
@@ -0,0 +1,126 @@
...
+import regex
+
+
+def extract_date(text: str) -> str | None:
+    """Extract the date from a string."""
+    date = None
+    date_match = regex.search(r"(\d{2}\.\d{2}\.\d{4})", text)
+    if date_match:
+        date = date_match.group(1)
+    else:
```
dcleres, line 12: I am not super happy about this construct. I think it became like this every time I added a new rule.

stijnvermeeren-swisstopo: How about something like

```
patterns = [
    r"(\d{2}\.\d{2}\.\d{4})", # some comment
    ...
]
for pattern in patterns:
    date_match = regex.search(pattern, text)
    if date_match:
        return date_match.group(1)
```

That would get rid of most of the code duplication and the excessively nested if-clauses.

src/stratigraphy/benchmark/score.py
```diff
@@ -193,13 +215,99 @@ def evaluate_metadata(predictions: dict) -> tuple[dict, pd.DataFrame]:
...
+            document_level_metrics_groundwater_information["groundwater_information"].append(0)
+
+    try:
+        groundwater_precision = groundwater_tp / (groundwater_tp + groundwater_fp)
+    except ZeroDivisionError:
+        groundwater_precision = 0
+    try:
+        groundwater_recall = groundwater_tp / (groundwater_tp + groundwater_fn)
+    except ZeroDivisionError:
+        groundwater_recall = 0
```
stijnvermeeren-swisstopo, line 246: This code is now more or less duplicated three times in this method (coordinates, ground water, ground water depth). Would it be possible to create a new method that computers the metrics for any of these properties (taking the exact property as a parameter) and that can then be called three times?

src/stratigraphy/main.py
```diff
@@ -234,6 +248,17 @@ def start_pipeline(
...
+        # print the metrics
+        logger.info("Performance metrics:")
+        logger.info(metrics)
+        logger.info("Groundwater level metrics:")
+        logger.info("Accuracy:")
+        logger.info(round(metrics["groundwater_accuracy"], 3))
+        logger.info("TP:")
+        logger.info(metrics["groundwater_tp"])
+        logger.info("Coordinate metrics:")
+        logger.info(round(metrics["coordinate_accuracy"], 3))
```
stijnvermeeren-swisstopo, line 260: We need to agree on what we output in the logger and how we format it, because this is starting to become rather messy.

src/stratigraphy/util/draw.py
```diff
@@ -107,6 +137,20 @@ def draw_coordinates(page: fitz.Page, coordinates: Coordinate) -> None:
...
+
+    Args:
+        page (fitz.Page): The page to draw on.
+        groundwater_information (GroundwaterInformation): The groundwater information to draw.
+    """
+    if isinstance(groundwater_information, dict):
+        groundwater_information = GroundwaterInformation(**groundwater_information)
+
+    if (page.number + 1) == groundwater_information.page:  ## page.number is 0-based
+        page.draw_rect(groundwater_information.rect, color=fitz.utils.getColor("pink"))
```
stijnvermeeren-swisstopo, line 151: Will need to be adjusted after https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/73 is merged

src/stratigraphy/groundwater/groundwater_extraction.py
```diff
@@ -0,0 +1,420 @@
...
+
+@dataclass
+class GroundwaterInformation(metaclass=abc.ABCMeta):
+    """Abstract class for Groundwater Information."""
+
+    depth: float  # Depth of the groundwater relative to the surface
+    date: str | None = (
+        None  # Date of the groundwater measurement, if several dates
+        # are present, the date of the document the last measurement is taken
+    )
```
stijnvermeeren-swisstopo, line 27: Why not already parse the extracted string and use a https://docs.python.org/3/library/datetime.html#datetime.date object here?

src/stratigraphy/groundwater/groundwater_extraction.py
```diff
@@ -0,0 +1,420 @@
...
+                continue
+            except ValueError:
+                continue
+        else:
+            return False  # date2 does not match any format
+
+        # Compare the parsed dates
+        return parsed_date1 == parsed_date2
+
+    def is_extracted_information_correct(self, extracted_information: "GroundwaterInformation") -> bool:
```
stijnvermeeren-swisstopo, line 110: Essentially this is an "equals" method for GroundwaterInformation objects, correct?

src/stratigraphy/groundwater/groundwater_extraction.py
```diff
@@ -0,0 +1,420 @@
...
+            return None
+
+        # Remove duplicates
+        matches = list(dict.fromkeys(matches))
+
+        # Sort the matches by their error counts (ascending order)
+        matches.sort(key=lambda x: x[1])
+
+        # Return the top three matches (lines only)
+        return [match[0] for match in matches[:5]]
```
stijnvermeeren-swisstopo, line 208: "top three" and "matches[:5]" don't seem to fit together?
What's the motivation for only considering an limited number of matches anyway?

src/stratigraphy/groundwater/groundwater_extraction.py
```diff
@@ -0,0 +1,420 @@
...
+                    extracted_groundwater_informations.append(extracted_gw_information)
+            except ValueError:
+                logger.warning("Could not extract groundwater information from the lines near the key.")
+
+        return self.select_best_groundwater_information(extracted_groundwater_informations)
+
+    def select_best_groundwater_information(
+        self, extracted_groundwater_informations: list[GroundwaterInformation]
+    ) -> GroundwaterInformation | None:
+        """Selects the best groundwater information from a list of extracted groundwater information.
```
stijnvermeeren-swisstopo, line 269: Why do this, instead of returning all extracted groundwater information?
The Zürich ground truth does contain at most one entry per borehole profile, but that is a limitation of the ground truth, not something that we actually want to imitate when extracting ground water data. Here, we do want to extract ALL available ground water level measurements, if possible.

stijnvermeeren-swisstopo (changes requested):
The computer vision (shape detection / template matching) are not yet part of this implementation, is that correct? Could you summarize (here and maybe also in the Jira ticket) the results of your experiments with those methods, and why you did not include them in this implementation yet?

I would not include the Jupyter notebooks in the repository. If you have anything that you think will be useful regularly, then please integrate it in the main code base. If you think that we should proceed with one of the computer vision approaches, then you could create a new ticket and a corresponding new Git branch for that, and already include the code in that branch.

src/stratigraphy/groundwater/utility.py
dcleres: I like this idea. Thank you! I will implement that!

src/stratigraphy/main.py
dcleres: I would say we can print the performance metrics.

src/stratigraphy/util/draw.py
dcleres: Yes, indeed.

src/stratigraphy/groundwater/groundwater_extraction.py
dcleres: Yes

src/stratigraphy/groundwater/utility.py
stijnvermeeren-swisstopo: Also, why not immediately match all cases at once, like `(\d{1,2}\.\d{1,2}\.\d{2,4})`?

Also, we need to make sure that we don't stop the rexeg in the middle of a number. Because the current implementation only extracts "9.1.20" from "9.1.2018".

src/stratigraphy/groundwater/groundwater_extraction.py
dcleres: Thank you for catching this. I started with the three best matches but realized that the performance was better with five. I think here the logic of limiting the number of matches was to improve the speed (not a big performance boost, but still) as well as to avoid using any match that could reduce the performance of extraction. Therefore, using the five best seemed like a good compromise. As you can see, the list of groundwater keywords is quite long, as there are many different representations in the different documents. This is a bit tricky to handle and the more words we have the higher is the chance for a false positive.

src/stratigraphy/groundwater/groundwater_extraction.py
dcleres: Ah I see. I was trying always to return the latest groundwater measurement that was performed or the most complete (elevation, depth, and date). I observed that by returning the "best" / most complete, the accuracy also improved since, in general, for a false positive keyword match, it would be hard to find elevation/depth/date around the keyword.

dcleres:
One of the challenges about the CV segmentation of the groundwater logo here is also that it is not only about finding the groundwater logo but also, once it is found, mapping it to the proper depth from the other information available in the borehole profile.

src/stratigraphy/benchmark/ground_truth.py
```diff
@@ -29,9 +29,21 @@ def __init__(self, path: Path):
                 if parse_text(layer["material_description"]) != ""
             ]
             metadata = ground_truth_item["metadata"]
+
+            # Add the groundwater to the metadata if it is present
+            if "groundwater" in ground_truth_item:
+                metadata["groundwater"] = ground_truth_item["groundwater"]
```
stijnvermeeren-swisstopo, line 35: I'm not such a fan of including groundwater data with the "metadata". For me, the groundwater data belongs to the main geological content of the borehole profile, not to the metadata. Therefore, I would rather keep them separate in the ground truth as well.

dcleres: I totally agree. I refactored the code to change this in our presentation.

src/stratigraphy/groundwater/groundwater_extraction.py
dcleres: Implemented in the latest version of the code.

src/stratigraphy/groundwater/groundwater_extraction.py
dcleres: I chose to overwrite the __eq__ operator instead to make it clearer.

dcleres:
@stijnvermeeren-swisstopo, I was able to address most of your comments and, as discussed this morning, left the comment where we select the N-best matches open for now. I will make a separate assessment of this later today. I do believe the code is ready for a second review round.

src/stratigraphy/util/draw.py
dcleres: Merged the code.

stijnvermeeren-swisstopo:
@dcleres I implemented extracting more than one ground water measurement per borehole profile, as I think this is quite an essential requirement. It also has a major influence on how we compute the scores. It is now no longer sufficient to just distinguish between "correct" and "incorrect", but we need to count TP, FP and FN, as we already do with the layers. The required a significant amount of changes in the code.

The metrics for groundwater extract have deteriorated a little bit, but I feel like we can improve them again by some relatively simple rules, such as removing duplicates in the extracted measurements, and having some validation rules to get rid of entries that are not plausible. It also seems like the depth and elevation values are sometimes mixed up (e.g. the elevation value is extracted as "depth"). I did not have time to try implementing any of these this yet. Could you have a go at that?

The metrics for coordinate extracted also were affected a little. I'm not 100% sure why. Probably it is related to how we deal with missing values / None values in the computations. It needs more investigation.

Further things included in my commit:
- Groundwater was still mixed up with the metadata in the ground truth classes, this is now kept separate
- Rename most occurrences of "groundwater_information" to simple "groundwater", as the word "information" does not add much clarity and it makes the variable names unnecessarily long.
- Use the same ISO 8601 date format "YYYY-MM-DD" everywhere, instead of using different formats in different places.
- Try to more cleanly separate the actual groundwater data for evaluation (depth, elevation, date) from the location in the PDF (which we only have for the extracted data, but not for the ground truth)

What I would like to do still in this ticket:
- Remove duplicates and add some simple validation rules for the extracted ground water measurements, to reduce the number of false positives.
- Reevaluate the choice of only considering the N-best matches.

What I would like to create follow-ups for:
- Find a better way to associate the "correctness" from the evaluation with the individual extracted values, so that we can show (red/green) the correctness of each extracted groundwater measurement also in the visualisation.
- Have more shared code for the computations of TP, FP, FN, F1 and other metrics for the layers on the one hand and for the groundwater and coordinates data on the other hand.
- Make the naming more consistent. (For example, ground truth JSON uses "date" while the predictions.json uses "measurement_date".)

dcleres:
@stijnvermeeren-swisstopo thank you for looking into the metrics and how to improve them regarding the scenario we are facing. 

In the latest commit, I addressed the following issues: 
- Removed the `_information` extension in all occurences, as indeed the extension does not add a lot of value and just makes it harder to read
- Added the accuracy to the metrics. I think this got lost in the different rebases. 
- A hardcoded threshold for the depth at 200m was added; as discussed, this should only be a temporary fix for the occasional confusion of the extractor between elevation and depth of the groundwater. I do believe that once we manage the merge of the elevation PR (#74), we should be able to build a logic that is able to identify these confusion cases. Otherwise, I was also thinking that we could take advantage of the depth_column information where the depths are also extracted and recorded. If the depth of the groundwater is larger than the maximum depth from the depth_column, there might be a good chance that there is a problem. Adding this threshold decreased the number of False positives 68 to 59.
- I also reevaluate the choice of only considering the N-best matches: There was no effect on the metrics. 

Currently the metrics are: 

```
{
   "F1":0.8790759488149966,
   "precision":0.8845491726684556,
   "recall":0.8736700405245328,
   "depth_interval_accuracy":0.0,
   "de_F1":0.8790759488149966,
   "de_precision":0.8845491726684556,
   "de_recall":0.8736700405245328,
   "de_depth_interval_accuracy":0.0,
   "coordinate_accuracy":0.592814371257485,
   "coordinate_precision":0.9339622641509434,
   "coordinate_recall":0.61875,
   "coordinate_f1":0.7443609022556391,
   "coordinates_tp":99,
   "coordinates_fp":7,
   "coordinates_fn":61,
   "groundwater_accuracy":0.013392857142857142,
   "groundwater_precision":0.03,
   "groundwater_recall":0.023622047244094488,
   "groundwater_f1":0.026431718061674006,
   "groundwater_tp":3,
   "groundwater_fp":97,
   "groundwater_fn":124,
   "groundwater_depth_accuracy":0.22043010752688172,
   "groundwater_depth_precision":0.41,
   "groundwater_depth_recall":0.3228346456692913,
   "groundwater_depth_f1":0.36123348017621143,
   "groundwater_depth_tp":41,
   "groundwater_depth_fp":59,
   "groundwater_depth_fn":86
}
```

stijnvermeeren-swisstopo:
@dcleres 
I removed the n-best cutoff, because indeed, this does not seem to make a difference anymore.
I also improved the deduplication logic. Previously, the same line could still be extracted multiple times, if it was matched multiple times with different values for "fuzzy_counts". This improved the "groundwater depth F1" metric by another 3%.

Even without those changes, my numbers are better than yours ("groundwater depth F1" is now at 47.7%). Are you using the latest ground truth file from https://jira.swisstopo.ch/browse/LGVISIUM-61 ?

I feel like many of the remaining mistakes are actually due to incomplete data in the ground truth. E.g. several groundwater measurements are correctly extracted, but the ground truth only contains one measurement. Maybe we can spend some time on fixing those in the ground truth as well? We can use the existing open ticket https://jira.swisstopo.ch/browse/LGVISIUM-61 for that.

Could you create follow-up tickets for the remaining three open points from my previous comment?

dcleres:
> I removed the n-best cutoff, because indeed, this does not seem to make a difference anymore.

@stijnvermeeren-swisstopo thank you for double-checking this

> Even without those changes, my numbers are better than yours ("groundwater depth F1" is now at 47.7%). Are you using the latest ground truth file from https://jira.swisstopo.ch/browse/LGVISIUM-61 ?

Actually, I was not. I only saw that you made some changes right now in the comments from the issue. Thank you for that! 

> I feel like many of the remaining mistakes are actually due to incomplete data in the ground truth. E.g. several groundwater measurements are correctly extracted, but the ground truth only contains one measurement. Maybe we can spend some time on fixing those in the ground truth as well? We can use the existing open ticket https://jira.swisstopo.ch/browse/LGVISIUM-61 for that.

I feel so too. Yes indeed let's clean the groundtruth asap

> Could you create follow-up tickets for the remaining three open points from my previous comment?

Ofc 

Thank you for the several reviews and I happy we now reached a stage where we can merge the code!

stijnvermeeren-swisstopo (approved):

PR #75, opened 2024-09-02
Close #LGVISIUM-64: initial api version

dcleres:
This is the initial version of the API that Geowerkstatt will call to build the borehole application with additional ML and AI features. 

Currently, the API is able:
- to extract coordinates and text 
- to convert a given PDF image (stored in S3) into a PNG image on the same S3 bucket.


src/stratigraphy/util/extract_text.py
```diff
@@ -44,3 +44,47 @@ def extract_text_lines(page: fitz.Page) -> list[TextLine]:
             current_line_words = []
 
     return lines
+
+
+def extract_text_lines_from_bbox(page: fitz.Page, bbox: fitz.Rect) -> list[TextLine]:
```
stijnvermeeren-swisstopo, line 49: Could the extract_text_lines method call this method (using the full page rect as bbox parameter) to avoid duplicated code?

stijnvermeeren-swisstopo (changes requested):

README.md
```diff
@@ -224,6 +229,81 @@ The project structure and the most important files are as follows:
...
+            "--reload",
+            "--host",
+            "0.0.0.0",
+            "--port",
+            "8002",
+        ],
+        "console": "integratedTerminal",
+        "justMyCode": true
+    }
+    ```
```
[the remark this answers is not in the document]
dcleres: That is a good point.

src/app/api/v1/endpoints/create_pngs.py
```diff
@@ -0,0 +1,74 @@
...
+
+    # Check if the PDF name is valid
+    if not aws_filename.endswith(".pdf"):
+        raise HTTPException(status_code=400, detail="Invalid request. The filename must end with '.pdf'.")
+    if len(aws_filename.split("/")) < 4:
+        error_details = "The filename must be in the format 'pdfs/{dataset_name}/{dataset_type}/{filename}.pdf'."
+        raise HTTPException(
+            status_code=400,
+            detail=error_details,
+        )
```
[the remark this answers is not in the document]
dcleres: We do not have to require this. I do believe that there is a misunderstanding on my side since I tried to replicate the structure we currently have for the data. So, just to be sure, the S3 bucket should have the `/pdf` and `/png' folders, but we do not split the data into train, validation, and test? This is indeed simpler. 

Regarding the prefixes, we can indeed automatically add them.

src/app/api/v1/endpoints/extract_data.py
```diff
@@ -0,0 +1,213 @@
...
+        y0 = extracted_coords.bbox.y0 * png_page_height / pdf_page_height
+        x1 = extracted_coords.bbox.x1 * png_page_width / pdf_page_width
+        y1 = extracted_coords.bbox.y1 * png_page_height / pdf_page_height
+
+        return ExtractCoordinatesResponse(
+            bbox=BoundingBox(x0=x0, y0=y0, x1=x1, y1=y1),
+            coordinates=Coordinates(
+                east=extracted_coords.coordinates.east,
+                north=extracted_coords.coordinates.north,
+                page=extracted_coords.coordinates.page,
```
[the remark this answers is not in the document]
dcleres: Good catch. Indeed, we should only be able to get single pages since Geowerkstatt will work with the PNG images from a single page. Thank you.

src/app/api/v1/endpoints/extract_data.py
```diff
@@ -0,0 +1,213 @@
...
+        pdf_page (fitz.Page): The PDF page.
+        user_defined_bbox (fitz.Rect): The user-defined bounding box.
+
+    Returns:
+        ExtractDataResponse: The extracted elevation.
+    """
+    # Extract the elevation
+    elevation = 444
+    bbox = BoundingBox(x0=20.0, y0=40.0, x1=60.0, y1=80.0)
+    return ExtractElevationResponse(bbox=bbox, elevation=elevation)
```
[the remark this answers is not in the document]
dcleres: Yes, that is just a placeholder. I was waiting for the elevation to be merged. I can now move ahead.

src/app/api/v1/endpoints/extract_data.py
```diff
@@ -0,0 +1,213 @@
...
+
+        return ExtractCoordinatesResponse(
+            bbox=bbox,
+            coordinates=Coordinates(
+                east=extracted_coord.east.coordinate_value,
+                north=extracted_coord.north.coordinate_value,
+                page=extract_data_request.page_number,
+                spacial_reference_system="LV95",
+            ),
+        )
```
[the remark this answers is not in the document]
dcleres: Good catch.

src/app/api/v1/endpoints/extract_data.py
```diff
@@ -0,0 +1,213 @@
...
+    text = ""
+    for text_line in text_lines:
+        text += text_line.text + " "
+
+    bbox = BoundingBox(
+        x0=user_defined_bbox.x0,
+        y0=user_defined_bbox.y0,
+        x1=user_defined_bbox.x1,
+        y1=user_defined_bbox.y1,
+    )
```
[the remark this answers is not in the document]
dcleres: Yes, this is indeed more coherent with the other functions. When writing this code, I thought we would only take the text within the user-specified bounding box. However, after the discussion with Geowerkstatt, I believe that this should be the behavior.

dcleres:
@stijnvermeeren-swisstopo I added most of the functionalities that were missing so far: 
- Correct error messages and codes when things are going wrong
- The calls to extract elevation and numbers are not placeholders anymore
- Implementation of tests with S3 bucket mocking

Please feel free to review the code more in-depth when you have some time.

src/app/api/v1/endpoints/create_pngs.py
```diff
@@ -0,0 +1,65 @@
...
+        raise HTTPException(
+            status_code=400, detail="Invalid request. 'filename' parameter is required and must be a string."
+        )
+
+    # Check if the PDF name is valid
+    if not aws_filename.endswith(".pdf"):
+        raise HTTPException(status_code=400, detail="Invalid request. The filename must end with '.pdf'.")
+
+    # Get the filename from the path
+    filename = aws_filename.split(".")[0]
```
stijnvermeeren-swisstopo, line 32: what if we have filenames that contain a `.` already before the file extension? For examples `file.one.pdf` and `file.two.pdf` ? They would both end up with `filename == 'file'`, no?

src/app/api/v1/endpoints/extract_data.py
```diff
@@ -0,0 +1,290 @@
...
+                target_height=png_page_height,
+                target_width=png_page_width,
+            ),
+            coordinates=Coordinates(
+                east=extracted_coords.coordinates.east,
+                north=extracted_coords.coordinates.north,
+                projection=extracted_coords.coordinates.projection,
+            ),
+        )
+    elif extract_data_request.format == FormatTypes.ELEVATION:
```
stijnvermeeren-swisstopo, line 87: I'm starting to question whether we need "elevation" as a dedicated format in this API at all. Wouldn't "number" do the trick as well as far as the "intelligent annotation" use case is concerned, as long as the user draws that bounding box precisely enough (i.e. so that no other numbers are inside the box)?

src/app/api/v1/endpoints/extract_data.py
```diff
@@ -0,0 +1,290 @@
...
+def extract_number(pdf_page: fitz.Page, user_defined_bbox: fitz.Rect):
+    """Extract numbers from a PNG image.
+
+    Args:
+        extract_data_request (ExtractDataRequest): The request data.
+        pdf_page (fitz.Page): The PDF page.
+        user_defined_bbox (fitz.Rect): The user-defined bounding box.
+
+    Returns:
+        ExtractDataResponse: The extracted numbers.
```
stijnvermeeren-swisstopo, line 253: we extract only one number per request, correct?

src/app/api/v1/endpoints/extract_data.py
```diff
@@ -0,0 +1,290 @@
...
+            number=extracted_number.number,
+        )
+    else:
+        raise ValueError("Invalid format type.")
+
+
+def extract_coordinates(
+    extract_data_request: ExtractDataRequest, pdf_page: fitz.Page, user_defined_bbox: fitz.Rect
+) -> ExtractDataResponse | None:
+    """Extract coordinates from a PNG image.
```
stijnvermeeren-swisstopo, line 142: At this point, we are not really concerned with the PNG image anymore, but we are rather extracting from the PDF page, I would say. So maybe we should change the docstring accordingly, here and for the other methods as well.

src/app/common/aws.py
```diff
@@ -0,0 +1,88 @@
...
+
+def load_data_from_aws(filename: str, format: str) -> bytes:
+    """Load a PNG image from AWS S3.
+
+    Args:
+        filename (str): The filename of the PNG image.
+        format (str): The format of the file.
+
+    Returns:
+        bytes: The loaded PNG image.
```
stijnvermeeren-swisstopo, line 64: Method applies not just to PNG images but to any files in the bucket.

stijnvermeeren-swisstopo (changes requested):

src/app/api/v1/endpoints/create_pngs.py
dcleres: Yes, that is certainly true. Currently, I tried to work with the filenames I observed from our data. However, would you like to proceed from here? I can also just remove the `.pdf` extension from the file, if you think that is a better solution.

dcleres: I am just not quite sure on how many different problematic filename we shall optimise for.

src/app/api/v1/endpoints/extract_data.py
dcleres: I see your point. The extraction of elevation using the elevation option might be more robust which could be a nice value add. However, I do not see any reason why the number interface would do the job (if the bbox is drawn correctly). Let me know your preference. If we consider the "intelligent annotation" use case, I guess it would make more sense to remove it.

src/app/api/v1/endpoints/extract_data.py
dcleres: Good catch

src/app/api/v1/endpoints/extract_data.py
dcleres: I tried to clarify the docstring in general. I think it is easy to get lost between the user-defined bbox in PNG coordinates and the PDF coordinates and what is give to what function.

src/app/api/v1/endpoints/create_pngs.py
stijnvermeeren-swisstopo: I think that just removing the `.pdf` extension (and only allowing files with a `.pdf` extension in the first place) should cover all possible cases.

src/app/common/aws.py
dcleres: Thank you for catching that. I do belive I wrote the dedicated the function for PNGs and PDFs later and did not update the docstring.

src/app/api/v1/endpoints/extract_data.py
stijnvermeeren-swisstopo: Let's remove it for now.

src/app/api/v1/endpoints/create_pngs.py
dcleres: Awesome! I will implement this. We were already checking for the `.pdf` extension.

dcleres:
@stijnvermeeren-swisstopo I addressed all the comments and edits you made and required. If there are no additional comments, I will be happy to proceed with merging the ticket. Thank you very much for the review!

src/app/api/v1/endpoints/create_pngs.py
```diff
@@ -29,7 +29,7 @@ def create_pngs(aws_filename: str = Form(...)):
         raise HTTPException(status_code=400, detail="Invalid request. The filename must end with '.pdf'.")
 
     # Get the filename from the path
-    filename = aws_filename.split(".")[0]
+    filename = aws_filename.replace(".pdf", "")
```
stijnvermeeren-swisstopo, line 32:
```suggestion
    filename = aws_filename[:-4]
```
I would do it like this, since we just checked that the filename ends with ".pdf", we are safe to remove the last four characters. You're code has the disadvantage that we might delete ".pdf" also in the middle of the string, in the (admittedly unlikely) case that we have a filename where ".pdf" occurs not just as the file extension, but also somewhere in the middle of the filename.

stijnvermeeren-swisstopo (changes requested):

src/app/api/v1/endpoints/create_pngs.py
dcleres: How about using a path specific function for this like here: https://stackoverflow.com/a/42288083/9543860

I think it would be "easier" to read and understand that we removed the suffix. What do you think? Otherwise, I am happy to merge the code with your suggestion.

stijnvermeeren-swisstopo: Yes, that's even better.
And then (for consistency) the check above can be written as `aws_filename.suffix == ".pdf"` instead of using the `endswith` function.

stijnvermeeren-swisstopo (approved):

PR #76, opened 2024-09-04
LGVISIUM-66: refactor computation of metrics

stijnvermeeren-swisstopo:
My first attempt at making everything a bit more coherent.

Through the refactoring, it became apparent that metrics are not computed very consistently.

- For layers and depth intervals, we track the macro average of the metrics. I.e. we first compute precision, recall and F1 score for each individual document, and then we take the average of those values.
  - Moreover, the overall precision is computed based on the macro-averaged precision and macro-averages recall. This can be different from the macro-averaged F1 score (i.e. the average of the F1 scores of the individual documents)! I don't think that this is a very usual metric, and we probably should compute a proper macro-averaged F1 score instead.
For metadata and groundwater, we use the micro average, i.e. we count the TP, FP and FN across all documents, and compute the overall F1, recall and precision metrics based on those counts.
- Considering that we are mostly interested in the question "how much work does an expert have to fix any mistaking during quality control", and this effort tends to be larger for documents with a lot of data, I would say that the micro average might be a better metric for us.

However, for now, I've left all the metrics exactly as they were computed previously. The code in the evaluate method in score.py now at least gives a clear overview of how each metric is computed.


src/stratigraphy/benchmark/score.py
```diff
@@ -80,95 +55,38 @@ def get_scores(
...
-        }
+        layer_metrics.metrics[filename] = Metrics(
+            tp=hits, fp=len(file_prediction.layers) - hits, fn=number_of_truth_values[filename] - hits
+        )
+        if depth_interval_occurences > 0:
+            depth_interval_metrics.metrics[filename] = Metrics(
+                tp=depth_interval_hits, fp=depth_interval_occurences - depth_interval_hits, fn=0
+            )
+
+    return layer_metrics, depth_interval_metrics
```
dcleres, line 66: Would it make sense to compute these metrics step by step? In general, I do believe that it is nice when a function does one thing at a time. However, in this case, I assume we would have to iterate through all files twice which does not go hand in hand with performance.

src/stratigraphy/benchmark/score.py
```diff
@@ -80,95 +55,38 @@ def get_scores(
...
     """Evaluate the borehole extraction predictions.
 
     Args:
        predictions (dict): The FilePredictions objects.
        number_of_truth_values (dict): The number of layer ground truth values per file.
 
     Returns:
-        tuple[dict, pd.DataFrame]: A tuple containing the overall metrics as a dictionary and the
-        individual document metrics as a DataFrame.
+        dict[str, DatasetMetrics]: A dictionary that maps a metrics name to the corresponding DatasetMetrics object
```
dcleres, line 79: I would tend to suggest creating a data object for this dict. This makes it clearer in my opinion. In the end, the dict will only contain filenames and their metrics, but by creating an object, we could make some methods to make the iterations through the object easier, for instance.

src/stratigraphy/benchmark/score.py
```diff
@@ -80,95 +55,38 @@ def get_scores(
...
-    )
-    document_level_metrics = pd.merge(
-        document_level_metrics, document_level_metrics_groundwater, on="document_name", how="outer"
-    )
-    document_level_metrics = pd.merge(
-        document_level_metrics, document_level_metrics_groundwater_depth, on="document_name", how="outer"
-    )
-    return metrics, document_level_metrics
+    all_metrics = evaluate_layer_extraction(predictions, number_of_truth_values)
+    all_metrics["coordinates"] = get_metrics(predictions, "metadata_is_correct", "coordinates")
```
dcleres, line 82: Could we refactor this to get all the metrics at once?

src/stratigraphy/benchmark/score.py
```diff
@@ -177,125 +95,17 @@ def get_metrics(predictions: dict[str, FilePredictions], field_key: str, field_n
         field_name (str): The name of the field being evaluated.
 
     Returns:
-        dict: The document level metrics and overall metrics.
+        dict: The requested DatasetMetrics object.
```
dcleres, line 98: Return type should be DatasetMetrics or am I mistaken?

src/stratigraphy/benchmark/score.py
```diff
@@ -379,23 +182,102 @@ def create_predictions_objects(
...
+            "de_precision": metrics["de_layer"].macro_precision(),
+            "de_depth_interval_accuracy": metrics["de_depth_interval"].macro_precision(),
+            "fr_F1": metrics["fr_layer"].pseudo_macro_f1(),
+            "fr_recall": metrics["fr_layer"].macro_recall(),
+            "fr_precision": metrics["fr_layer"].macro_precision(),
+            "fr_depth_interval_accuracy": metrics["fr_depth_interval"].macro_precision(),
+            "coordinate_f1": coordinates_metrics.f1,
+            "coordinate_recall": coordinates_metrics.recall,
+            "coordinate_precision": coordinates_metrics.precision,
+            "groundwater_f1": groundwater_metrics.f1,
```
dcleres, line 242: I introduced the notion of groundwater and groundwater depth. However, I do think that this duplicates a lot of the code, and we are mostly interested in the depth of the groundwater, correct? Therefore, I would suggest to delete the metrics for the entire groundwater object.

dcleres (changes requested):
I think the computations of the metrics is much clearer and I am happy to see that the code became much lighter when it comes to compute the metrics. I added a few minor comments and questions that we can try to address together. 
I have this think a bit about which metrics (micro or macro) are the most relevant for us but I do tend to agree with what you wrote on JIRA at the moment.

src/stratigraphy/benchmark/score.py
stijnvermeeren-swisstopo: I agree, that would be cleaner, and I think the affect on performance will not be noticeable. I'll split this into two methods.

src/stratigraphy/benchmark/score.py
stijnvermeeren-swisstopo: Yes, good idea, I already thought about this as well. I don't see much potential yet to simplify the code much by doing this, but at least the evaluate() method in score.py can become a lot shorter, by moving the computation of the document_level_metrics dataframe and the metrics dictionary into the new class.

src/stratigraphy/benchmark/score.py
stijnvermeeren-swisstopo: I'm not so sure what you mean, could you elaborate how you would refactor this?

src/stratigraphy/benchmark/score.py
stijnvermeeren-swisstopo: yes, fixed

src/stratigraphy/benchmark/score.py
stijnvermeeren-swisstopo: I tend to disagree, I would say that the elevation and especially the measurement date are also relevant. So either we add metrics for those separately, of we compute a metric for the whole groundwater object, as we currently do.

dcleres: I think you are right when it comes to the date. I had only the elevation of the groundwater in mind. Then we can keep things like this.

src/stratigraphy/benchmark/score.py
dcleres: It was rather a small fix. What I meant: we are getting the `all_metrics` object, and then we are computing the metrics function for every single feature. To me, it seems to be nicer with the all_metrics` would directly have the `coordinates`, etc. metrics right at creation.

stijnvermeeren-swisstopo: We agreed not to change this now. We will likely touch this code again during the refactoring ticket that is already in progress. It is certainly true that the current implementation is not very consistent. However, another factor to consider, is that we want to keep the scoring of the different properties somewhat independent (e.g. allow scoring coordinates without necessarily also scoring everything else).

src/stratigraphy/benchmark/metrics.py
```diff
@@ -0,0 +1,155 @@
...
+
+    @property
+    def precision(self) -> float:
+        """Calculate the precision."""
+        if self.tp + self.fp > 0:
+            return self.tp / (self.tp + self.fp)
+        else:
+            return 0
+
+    @property
```
dcleres, line 33: IMO this could be a one lineer

`return self.tp / (self.tp + self.fn) if (self.tp + self.fn) > 0 else 0`

src/stratigraphy/benchmark/metrics.py
```diff
@@ -0,0 +1,155 @@
...
+    def recall(self) -> float:
+        """Calculate the recall."""
+        if self.tp + self.fn > 0:
+            return self.tp / (self.tp + self.fn)
+        else:
+            return 0
+
+    @property
```
dcleres, line 41: Could also be a one liner

src/stratigraphy/benchmark/metrics.py
```diff
@@ -0,0 +1,155 @@
...
+    def macro_precision(self) -> float:
+        """Compute the macro precision score."""
+        if self.metrics:
+            return sum([metric.precision for metric in self.metrics.values()]) / len(self.metrics)
+        else:
+            return 0
+
+    def macro_recall(self) -> float:
+        """Compute the macro recall score."""
+        if self.metrics:
```
dcleres, line 80: Could also be a one liner

src/stratigraphy/benchmark/metrics.py
```diff
@@ -0,0 +1,155 @@
...
+    def macro_f1(self) -> float:
+        """Compute the macro F1 score."""
+        if self.metrics:
+            return sum([metric.f1 for metric in self.metrics.values()]) / len(self.metrics)
+        else:
+            return 0
+
+    def macro_precision(self) -> float:
+        """Compute the macro precision score."""
+        if self.metrics:
```
dcleres, line 73: Could also be a one liner

src/stratigraphy/benchmark/metrics.py
```diff
@@ -0,0 +1,155 @@
...
+        """Can be used to compute micro averages."""
+        return Metrics(
+            tp=sum(metric.tp for metric in self.metrics.values()),
+            fp=sum(metric.fp for metric in self.metrics.values()),
+            fn=sum(metric.fn for metric in self.metrics.values()),
+        )
+
+    def macro_f1(self) -> float:
+        """Compute the macro F1 score."""
+        if self.metrics:
```
dcleres, line 66: Could also be a one liner

src/stratigraphy/benchmark/metrics.py
```diff
@@ -0,0 +1,155 @@
...
+            return 2 * self.precision * self.recall / (self.precision + self.recall)
+        else:
+            return 0
+
+
+class DatasetMetrics:
+    """Keeps track of a particular metrics for all documents in a dataset."""
+
+    def __init__(self):
+        self.metrics: dict[str, Metrics] = {}
```
dcleres, line 54: Would it make sense for the metric to carry the filename as an attribute? This would make it possible here only to use a file of metrics that already know their filename and avoid creating the dict just to keep track of the filenames. I also had that issue in the refactoring of the metadata metrics and was also not sure about the best solution.

dcleres (commented):
There are a few minor comments that would be nice to address. Thank you very much for incorporating my previous comments.

src/stratigraphy/benchmark/metrics.py
stijnvermeeren-swisstopo: I think I'd like to keep the filename separate from the Metrics class for two reasons:
- Currently the Metrics class is also used to compute the metrics for the entire dataset (`DatasetMetrics.overall_metrics()`); in this case there is no specific filename that can be associated with the Metrics object
- In many places where a Metrics object is instantiated, we currently don't have access to the filename. I don't really want to pass additional parameters around, just for this.

That does not mean that we cannot improve the coupling between metrics and filename, but I think it should probably happen at a different level in the code. Maybe we can have a look at it in the context of your refactoring PR?

src/stratigraphy/benchmark/metrics.py
stijnvermeeren-swisstopo: That's true, but in my humble opinion, this would actually make the code less easily readable/understandable, but that's of course just a subjective opinion.
If you don't feel strongly about this, I would keep the current code style.

src/stratigraphy/benchmark/metrics.py
dcleres: Yes, I tend to share your opinion. Then all good for me!

dcleres:
@stijnvermeeren-swisstopo while running the code in the `main` and the `LGVISIUM-66` branches, I saw that the results on mlflow are the same but the fields like `groundwater_depth_fn` do not exist anymore. I am assuming that this change was made on purpose?

stijnvermeeren-swisstopo:
@dcleres On purpose, as discussed on Teams ;)

image

dcleres (approved):

PR #80, opened 2024-09-13
Close #LGVISIUM-73: Create a metadata object and look into the file organisation

dcleres:
I tried to move the different files from the `utils` directory into meaningful new directories that are grouped by the extracted borehole feature. Furthermore, I refactored the `BoreholeMetadata` class and created a way to evaluate the metadata on its own. There will be a follow-up ticket for the layers and other parts of the pipeline. The goal I was pursuing was to aim for no dictionaries that are filled on the fly but to create data classes that have attributes for the different features. 

I also tried to conceptually differentiate between classes that hold data and classes that compute and evaluate the extracted information.


dcleres:
Screenshot 2024-09-16 at 14 43 34

Overall, the idea is that the refactored architecture follows the following pattern. The `FilePrediction` object we currently have as a large `dict` should become an object that hosts sub-objects such as the `BoreholeMetadata` object. This `BoreholeMetadata` can then be evaluated individually using a `BoreholeMetadataEvaluator`. 

Grouping the different objects into classes make it easier to keep track of the fields in the object as they need to be declarer in advance and not on of the fly as we currently do in the `dict` objects.

dcleres:
Screenshot 2024-09-16 at 16 01 20

Need to investigate why the metrics are different. Currently, checked the formula but did not find anything obvious.

stijnvermeeren-swisstopo (changes requested):
I'm not sure about the use case of the "metadata pipeline" (executing only metadata extraction) yet. Seems to be a significant amount of code to maintain, for something that we will very rarely use, unless it's significantly faster (but currently it does not seem to be?).

Apart from that, the code structure seems to become significantly cleaner; I like it.

What would be the plan for further development in this direction? Apply the same structure for groundwater and for layers as well?

src/stratigraphy/evaluation/evaluation_dataclasses.py
```diff
@@ -0,0 +1,138 @@
...
+
+import abc
+from dataclasses import dataclass
+
+import pandas as pd
+
+
+@dataclass
+class Metrics(metaclass=abc.ABCMeta):
+    """Metrics for metadata."""
```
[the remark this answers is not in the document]
dcleres: That was a left over from a legacy implementation.

src/stratigraphy/evaluation/evaluation_dataclasses.py
```diff
@@ -0,0 +1,138 @@
...
+        Args:
+            metric_list (list): The list of metrics.
+
+        Returns:
+            Metrics: Combined metrics.
+        """
+        tp = sum([metric.tp for metric in metric_list])
+        fp = sum([metric.fp for metric in metric_list])
+        fn = sum([metric.fn for metric in metric_list])
+        return Metrics(tp=tp, fp=fp, fn=fn)
```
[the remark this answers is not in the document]
dcleres: I kept the implementation with the static method as this seems to be more flexible than having to build a DatasetMetrics object every time. Currently, only the `groundwater_metrics` are also using this functionality. I would suppose that during the refactoring of this class, we will find a better solution. Does that work for you @stijnvermeeren-swisstopo?

src/stratigraphy/metadata/elevation_extraction.py
```diff
@@ -197,4 +193,8 @@ def extract_elevation(self) -> ElevationInformation | None:
         for page in self.doc:
             page_number = page.number + 1  # page.number is 0-based
 
-            return self.extract_elevation_from_bbox(page, page_number)
+            # Extract the elevation information from the page
+            # Make the search area a bit larger than the page itself - This is needed for "267123120-bp.pdf" where the
+            # elevation is located outside the page boundaries apparently.
+            search_area = page.rect * 1.25
+            return self.extract_elevation_from_bbox(page, page_number, search_area)
```
stijnvermeeren-swisstopo, line 200: This should not be necessary. The bbox is an optional param in this method, it should correctly take the whole page when you omit this param (None value).

src/stratigraphy/evaluation/evaluation_dataclasses.py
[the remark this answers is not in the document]
stijnvermeeren-swisstopo: It's better now, using this method everywhere, and having removed the duplicated method in the DatasetMetrics class.

I still don't really like the fact that we now have averaging methods in different places, this one here, and others (e.g. macro_f1) in the DatasetMetrics class. However, I'm happy to look into that at a later point in time. It's not a great situation anyway that we need so many different averaging methods... For now, maybe just add a "TODO" comment about this?

I would also rename this method to something like `micro_average`, as this defines more clearly what this method does and how it differs from the other methods in DatasetMetrics.

stijnvermeeren-swisstopo (changes requested):

stijnvermeeren-swisstopo (changes requested):
Elevation and coordinates are always drawn with a red line in the visualisations, even when the values are correct.

src/stratigraphy/benchmark/score.py
```diff
@@ -203,34 +218,64 @@ def create_predictions_objects(
...
+    metadata_metrics_list = evaluate_metadata_extraction(metadata_per_file, ground_truth_path)
+    metadata_metrics = metadata_metrics_list.get_cumulated_metrics()
+    document_level_metrics = metadata_metrics_list.get_document_level_metrics()
+
+    document_level_metrics.to_csv(
+        temp_directory / "document_level_metadata_metrics.csv"
+    )  # mlflow.log_artifact expects a file
+
+    # print the metrics
+    logger.info("Metadata Performance metrics:\n\n")
```
stijnvermeeren-swisstopo, line 240: why the double newline?

src/stratigraphy/benchmark/score.py
```diff
@@ -203,34 +218,64 @@ def create_predictions_objects(
...
+            metrics.document_level_metrics_df().to_csv(
+                temp_directory / "document_level_metrics.csv", index_label="document_name"
+            )  # mlflow.log_artifact expects a file
+            metrics_dict = metrics.metrics_dict()
 
-        if mlflow_tracking:
-            import mlflow
+            # Format the metrics dictionary to limit to three digits
+            formatted_metrics = {k: f"{v:.3f}" for k, v in metrics_dict.items()}
+            logger.info("Performance metrics:\n\n %s", formatted_metrics)
```
stijnvermeeren-swisstopo, line 271: Why the double newline? I thought that log statements should generally be on one line, as many monitoring tools rely on that?

src/stratigraphy/evaluation/evaluation_dataclasses.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+@dataclass
+class OverallBoreholeMetadataMetrics(metaclass=abc.ABCMeta):
+    """Metrics for borehole metadata."""
+
+    borehole_metadata_metrics: list[BoreholeMetadataMetrics] = None
+
+    def __init__(self):
+        """Initializes the FileBoreholeMetadataMetrics object."""
```
stijnvermeeren-swisstopo, line 117:
```suggestion
        """Initializes the OverallBoreholeMetadataMetrics object."""
```

src/stratigraphy/evaluation/evaluation_dataclasses.py
```diff
@@ -0,0 +1,138 @@
...
+                "coordinate": [self.coordinates_metrics.f1],
+            }
+        )
+
+
+@dataclass
+class OverallBoreholeMetadataMetrics(metaclass=abc.ABCMeta):
+    """Metrics for borehole metadata."""
+
+    borehole_metadata_metrics: list[BoreholeMetadataMetrics] = None
```
stijnvermeeren-swisstopo, line 114:
```suggestion
    borehole_metadata_metrics: list[FileBoreholeMetadataMetrics] = None
```

src/stratigraphy/evaluation/evaluation_dataclasses.py
```diff
@@ -0,0 +1,138 @@
...
+            elevation_metrics=elevation_metrics, coordinates_metrics=coordinates_metrics
+        ).to_json()
+
+    def get_document_level_metrics(self) -> pd.DataFrame:
+        """Get the document level metrics."""
+        # Collect all the data frames in a list
+        frames = [metadata.get_document_level_metrics() for metadata in self.borehole_metadata_metrics]
+
+        # Concatenate them once at the end
+        return pd.concat(frames, ignore_index=True)
```
stijnvermeeren-swisstopo, line 138: This does not return the input documents in the original order. Where does the order get lost?

Maybe the implementation approach from `DatasetMetricsCatalog.document_level_metrics_df` is more robust?

stijnvermeeren-swisstopo (changes requested):

src/stratigraphy/evaluation/evaluation_dataclasses.py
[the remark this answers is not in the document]
dcleres: Added TODOs in both place in the code and renamed the function to micro_average.

src/stratigraphy/main.py
```diff
@@ -127,122 +187,123 @@ def start_pipeline(
...
+                                    page, geometric_lines, scale_factor=line_detection_params["pdf_scale_factor"]
+                                )
+                                mlflow.log_image(img, f"pages/{filename}_page_{page.number + 1}_lines.png")
+
+                if part == "all":
+                    predictions[filename]["layers"] = layer_predictions_list
+                    predictions[filename]["depths_materials_column_pairs"] = depths_materials_column_pairs_list
+                    predictions[filename]["page_dimensions"] = metadata.page_dimensions
+
+                    assert len(metadata.page_dimensions) == doc.page_count, "Page count mismatch."
```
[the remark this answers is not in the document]
dcleres: I moved this into the BoreholeMetadata class as this seems the best place to place it.

src/stratigraphy/main.py
```diff
@@ -127,122 +187,123 @@ def start_pipeline(
...
+
+    logger.info("Metadata written to %s", metadata_path)
+    with open(metadata_path, "w", encoding="utf8") as file:
+        json.dump(metadata_per_file.to_json(), file, ensure_ascii=False)
+
+    if part == "all":
```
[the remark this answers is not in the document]
dcleres: Yes, I think we should be able to solve this with the refactoring of the layers objects

src/stratigraphy/benchmark/score.py
dcleres: I thought it was easier to read in the terminal. However, if you do not share this opinion, I am happy to roll back.

src/stratigraphy/benchmark/score.py
dcleres: Thank you for the explanation, I rolled back on this.

src/stratigraphy/benchmark/score.py
dcleres: Based on this I rolled back: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/80#discussion_r1764541075

src/stratigraphy/evaluation/evaluation_dataclasses.py
dcleres: I cannot reproduce the issues you are mentioning. 

I tried something along these lines: 

```python

        document_level_metrics = pd.DataFrame(columns=["document_name", "elevation", "coordinate"])
        for metadata in self.borehole_metadata_metrics:
            document_level_metrics = document_level_metrics.merge(metadata.get_document_level_metrics(), how="outer")

```

But this actually sorted the filenames, and the order was lost. Is there a reason why the other is important here in your opinion?

dcleres: I initially had this implementation 

```
document_level_metrics = pd.DataFrame(columns=["document_name", "elevation", "coordinate"])
for metadata in self.borehole_metadata_metrics:
    document_level_metrics = pd.concat([document_level_metrics, metadata.get_document_level_metrics()])
```

but this raises a deprecation warning.

dcleres: I see. I already suggested a new fix. Maybe you can have a look at it. There is even an assertion that makes sure the files are in order.

stijnvermeeren-swisstopo: You assertion ensures that the order is unchanged (i.e. execution order), but this is not necessarily alphabetical, at least on my system it isn't.

stijnvermeeren-swisstopo (approved):

PR #81, opened 2024-09-16
Close #LGVISIUM-70: Combine groundwater information with elevation to determine groundwater depth

dcleres:
Combined the information from the elevation with the groundwater information. This makes it possible to infer the elevation or depth from a groundwater measurement based on the elevation information from the borehole profile. The groundwater metrics benefited quite a bit from these changes, and with this feature, we can fill in missing depth or elevation information which were required. 

During the work on this issue, I also saw that some of the information from the Groundtruth seems to be wrong. 

There are still a few false positives due to the fact that the depth of the groundwater sometimes seems to be determined against the elevation of the drilling and sometimes against the elevation of the borehole. It is not exactly clear to me what the difference is and how we can mitigate this.

### DIFF
--- a/config/matching_params.yml
+++ b/config/matching_params.yml
@@ -166,7 +166,6 @@ elevation_keys:
   - Höhe
   - Kote
   - OK Terrain
-  - Tiefe ab OKT
   - OKT
   - Ansatzhöhe
   - Terrainkote
--- a/config/matching_params.yml
+++ b/config/matching_params.yml
@@ -135,6 +135,8 @@ groundwater_keys:
   - Wsp.
   - Wsp
   - GW-Spiegel
+  - Grundwasser
+  - Spiegel
 
   # French
   - nappe phréatique
--- a/config/matching_params.yml
+++ b/config/matching_params.yml
@@ -134,6 +134,7 @@ groundwater_keys:
   - GrW Sp
   - Wsp.
   - Wsp
+  - GW-Spiegel
 
   # French
   - nappe phréatique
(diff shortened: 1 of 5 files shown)


src/stratigraphy/groundwater/groundwater_extraction.py
```diff
@@ -267,7 +269,7 @@ def get_groundwater_info_from_lines(self, lines: list[TextLine], page: int) -> G
         else:
             raise ValueError("Could not extract all required information from the lines provided.")
 
-    def extract_groundwater(self) -> list[GroundwaterInformationOnPage]:
+    def extract_groundwater(self, elevation: ElevationInformation | None) -> list[GroundwaterInformationOnPage]:
```
stijnvermeeren-swisstopo, line 272: Maybe naming this `terrain_elevation` or similar would be better, to distinguish it from the groundwater elevation?

config/matching_params.yml
```diff
@@ -163,6 +166,7 @@ elevation_keys:
   - Höhe
   - Kote
   - OK Terrain
+  - Tiefe ab OKT
```
stijnvermeeren-swisstopo, line 169: Depth is something different from elevation, this does not seem to belong here. For which input does this bring an improvement?

stijnvermeeren-swisstopo (changes requested):

src/stratigraphy/groundwater/groundwater_extraction.py
dcleres: That sounds like a good idea!

config/matching_params.yml
dcleres: I agree. This used to make a difference at some point. However, now I checked again and I cannot see any effect anymore. Thank you for making me look into it a second time after I added all the changes.

stijnvermeeren-swisstopo (approved):

PR #82, opened 2024-09-16
minor refactoring / fixes

stijnvermeeren-swisstopo:
No changes to the behaviour of the pipeline


dcleres (approved):
I checked everything with mlflow as well. Performance is preserved, and I do not have a major issue with the way you implemented the improvement. Thank you!

PR #83, opened 2024-09-20
Close LGVISIUM-63: Extraction of the groundwater logo using computer vision

dcleres:
Addition of template matching to the code. This should make it possible to find groundwater information without the respective keywords. 

In this PR, I also tried to remove the keywords that were not used and to remove them if needed. Furthermore, I added a list of FP keywords. This list contains keywords that were leading to false positives in the detections.


stijnvermeeren-swisstopo (changes requested):
I would not enable this functionality by default at the moment, as it is just too slow and does not lead to an improvement that is significant enough to justify this slow-down. So maybe we should add a config parameter that can be used to decide whether to apply template matching or not. Then we can keep the code for now, and create a follow-up ticket to look into ways to optimize it.

One potential idea would be to look into using the OpenCV2 implementation of template matching instead of the scikit-learn one. Some people seem to claim that the former is more performant (e.g. https://www.reddit.com/r/opencv/comments/g8kdcs/question_the_speed_of_matchtemplate/).

src/stratigraphy/groundwater/groundwater_extraction.py
```diff
@@ -137,9 +141,11 @@ class GroundwaterLevelExtractor(DataExtractor):
     feature_name = "groundwater"
 
     # look for elevation values to the left, right and/or immediately below the key
-    search_left_factor: float = 2
+    search_left_factor: float = 3  # NOTE: check files 267125334-bp.pdf, 267125338-bp.pdf, and 267125339-bp.pdf if this
+    # value is too high, as it might lead to false positives
```
stijnvermeeren-swisstopo, line 145: Yes, this, in combination with the new `search_above_factor`, indeed seems to lead to too many false positives. But maybe the ongoing work in https://jira.swisstopo.ch/browse/LGVISIUM-77 will already make this more robust again?

src/stratigraphy/groundwater/groundwater_extraction.py
```diff
@@ -287,10 +476,18 @@ def extract_groundwater(self, terrain_elevation: Elevation | None) -> list[Groun
...
+            found_groundwater = self.get_groundwater_near_key(lines, page_number)
+            if not found_groundwater:
+                logger.info("No groundwater found near the key on page %s.", page_number)
+                found_groundwater, confidence_list = self.get_groundwater_from_illustration(
+                    lines, page_number, terrain_elevation
+                )
+                logger.info("Confidence list: %s", confidence_list)
+                print("Confidence list: %s", confidence_list)
+                logger.info("Found groundwater from illustration on page %s: %s", page_number, found_groundwater)
+                print("Found groundwater from illustration on page %s: %s", page_number, found_groundwater)
```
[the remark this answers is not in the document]
dcleres: Indeed. I adapted this.

src/stratigraphy/groundwater/utility.py
```diff
@@ -45,12 +45,15 @@ def extract_depth(text: str, max_depth: int) -> float | None:
...
-                break
+            try:
+                depth = float(depth_match.group(1).replace(",", "."))
+                if depth > max_depth:
+                    # If the extracted depth is greater than the max depth, set it to None and continue searching.
+                    depth = None
+                else:
+                    break
+            except ValueError:
+                continue
```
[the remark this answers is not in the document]
dcleres: The code was essentially testing whether the year in tha date has two digits or four digits and then converted it into a date based format. We would run into the scenario of raising an exeception if the date has two digits to indicate the year instead of four. Since this implementation was not clear I changed it to something that is easier to read.

src/stratigraphy/groundwater/groundwater_extraction.py
dcleres: The issue I was facing with the 267125334-bp.pdf, 267125338-bp.pdf, and 267125339-bp.pdf bore profiles was that False Positives were generated if the left search factor was too large. In these profiles, the algorithm would find the depth column and extract data. 

I think the best option performance-wise would be to use the default values from the `main` branch.

src/stratigraphy/groundwater/utility.py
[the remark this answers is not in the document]
dcleres: @stijnvermeeren-swisstopo and @dcleres looked into it together, and we decided to roll back to the current version in the `main` branch and create a follow-up ticket (LGVISIUM-100) to address this issue.

dcleres:
I addressed the comments you raised @stijnvermeeren-swisstopo . Thank you very much for the review. I added the possibility of running the template matching on demand by editing the environment variable `IS_SEARCHING_GROUNDWATER_ILLUSTRATION`. 

Even when not running the template matching, the metrics were improved: 

Screenshot 2024-10-14 at 14 28 38

main branch was run: `vaunted-mink-342`

pyproject.toml
```diff
@@ -35,6 +35,7 @@ dependencies = [
     "opencv-python-headless",
     "quads>=1.1.0",
     "numpy<2",
+    "scikit-image==0.24.0"
```
stijnvermeeren-swisstopo, line 38: move to optional dependencies + add corresponding documentation to README

src/stratigraphy/data_extractor/data_extractor.py
```diff
@@ -63,6 +66,11 @@ def __init__(self, document: fitz.Document):
 
         self.doc = document
         self.feature_keys = read_params("matching_params.yml")[f"{self.feature_name}_keys"]
+        self.feature_fp_keys = (
+            read_params("matching_params.yml")[f"{self.feature_name}_fp_keys"]
+            if read_params("matching_params.yml")[f"{self.feature_name}_fp_keys"]
+            else []
+        )
```
stijnvermeeren-swisstopo, line 73: is this the same as `read_params("matching_params.yml")[f"{self.feature_name}_fp_keys"] or []` ?

src/stratigraphy/data_extractor/data_extractor.py
```diff
@@ -105,7 +113,15 @@ def find_feature_key(self, lines: list[TextLine], allowed_error_rate: float = 0.
...
                 if match:
-                    matches.add(line)
+                    # Make sure the key is not in the false positive list
+                    is_fp_key = False
+                    for fp_key in self.feature_fp_keys:
+                        if fp_key in line.text:
+                            is_fp_key = True
+                            break
+
+                    if not is_fp_key:
```
stijnvermeeren-swisstopo, line 123: Is this the same as 
```
if all(not fp_key in line.text for fp_key in self.feature_fp_keys):
```

src/stratigraphy/groundwater/groundwater_extraction.py
```diff
@@ -2,12 +2,16 @@
...
+import math
+import os
 from dataclasses import dataclass
 from datetime import date as dt
 from datetime import datetime
+from pathlib import Path
 
 import fitz
 import numpy as np
+import skimage as ski
```
stijnvermeeren-swisstopo, line 14: Move all logic related to template matching (including this import) to a separate Python file, and only import this file when template matching is enabled.

src/stratigraphy/groundwater/groundwater_extraction.py
```diff
@@ -219,6 +232,12 @@ def get_groundwater_info_from_lines(self, lines: list[TextLine], page: int) -> G
                     if extracted_date_str:
                         text = text.replace(extracted_date_str, "").strip()
                         date = extracted_date
+                        matched_lines_rect.append(line.rect)
+                else:
+                    # in case several dates are present, we skip the other dates
```
stijnvermeeren-swisstopo, line 237: Clearer description:
"If a second date is present in the lines around the groundwater key, then we skip this line, instead of potentially falsely extracting a depth value from the date."

src/stratigraphy/groundwater/groundwater_extraction.py
```diff
@@ -269,6 +288,184 @@ def get_groundwater_info_from_lines(self, lines: list[TextLine], page: int) -> G
...
+                confidence = np.max(result)  # TODO - use confidence to filter out bad matches
+                if confidence < TEMPLATE_MATCH_THRESHOLD:
+                    # skip this template if the confidence is too low to avoid false positives
+                    continue
+                top_left = (ij[1], ij[0])
+                illustration_rect = fitz.Rect(
+                    top_left[0], top_left[1], top_left[0] + template.shape[1], top_left[1] + template.shape[0]
+                )
+
+                # remove the matched area from the result to avoid finding the same area again
```
stijnvermeeren-swisstopo, line 348:
```suggestion
                # remove the matched area from the template matching result to avoid finding the same area again for the same template
```

src/stratigraphy/groundwater/groundwater_extraction.py
```diff
@@ -285,12 +482,24 @@ def extract_groundwater(self, terrain_elevation: Elevation | None) -> list[Groun
...
-
-            found_groundwater = (
-                self.get_groundwater_near_key(lines, page_number)
-                # or XXXX # Add other techniques here
-            )
+            page_number = page.number + 1  # NOTE: page.number is 0-based
+
+            found_groundwater = self.get_groundwater_near_key(lines, page_number)
+            if not found_groundwater:
+                logger.info("No groundwater found near the key on page %s.", page_number)
```
stijnvermeeren-swisstopo, line 489: Not necessary to log this.

stijnvermeeren-swisstopo (changes requested):

src/stratigraphy/groundwater/groundwater_extraction.py
dcleres: Removed the logging about this

src/stratigraphy/groundwater/groundwater_extraction.py
```diff
@@ -285,12 +482,24 @@ def extract_groundwater(self, terrain_elevation: Elevation | None) -> list[Groun
...
+
+                ### Extract groundwater from illustration
+                if self.is_searching_groundwater_illustration:
+                    found_groundwater, confidence_list = self.get_groundwater_from_illustration(
+                        lines, page_number, terrain_elevation
+                    )
+                    if found_groundwater:
+                        logger.info("Confidence list: %s", confidence_list)
```
dcleres, line 497: Is this helpful logging according to you, @stijnvermeeren-swisstopo? In a previous iteration, we removed the logging in case no groundwater was found. We can also remove the logging when groundwater is found, as I mostly use it for debugging purposes.

pyproject.toml
dcleres: Done.

src/stratigraphy/data_extractor/data_extractor.py
dcleres: I used this experession as it seemed a bit easier to read for me: 

`if not any(fp_key in line.text for fp_key in self.feature_fp_keys):`

src/stratigraphy/data_extractor/data_extractor.py
dcleres: Yes, and shorter! Thanks!

src/stratigraphy/groundwater/groundwater_extraction.py
dcleres: Thank you!

src/stratigraphy/groundwater/groundwater_extraction.py
dcleres: Done.

dcleres:
@stijnvermeeren-swisstopo I do believe I implemented all the changes we discussed today. Main change, the template matching is now in an independent and separate file.

pyproject.toml
```diff
@@ -54,6 +54,9 @@ visualize = [
 devtools = [
     "tqdm"
 ]
+groundwater_illustration_matching = [
+    "scikit-image==0.24.0"
+]
 
 all = ["swissgeol-boreholes-dataextraction[test, lint, experiment-tracking, visualize, devtools]"]
```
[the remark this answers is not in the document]
dcleres: Oh thanks! I was not aware of that!

dcleres:
@stijnvermeeren-swisstopo thank you for your review. I add the `groundwater_illustration_matching` to the toml installation script.

pyproject.toml
```diff
@@ -54,8 +54,11 @@ visualize = [
 devtools = [
     "tqdm"
 ]
+groundwater_illustration_matching = [
```
stijnvermeeren-swisstopo, line 57: It still doesn't quite work for me to do install scikit-image with `pip install -e '.[all]'` .
Probably need to use hyphens here instead of underscores (i.e. `groundwater-illustration-matching`).
I think there's some normalisation happening in the background that interferes with underscores: see e.g. https://peps.python.org/pep-0685/
We had the same issue already once before with Renato: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/54#discussion_r1625849873

stijnvermeeren-swisstopo (changes requested):

pyproject.toml
dcleres: Ah thank you for letting me know!

stijnvermeeren-swisstopo (approved):
LGTM!

PR #84, opened 2024-09-24
Close #LGVISIUM-82: Refactor the FilePrediction object and its evaluation

dcleres:
I refactored the FilePrediction object to remove dictionary-based data structures and replaced them with specialized data class objects. I also created a new object to contain the depth material column pairs. This should make the content of the different objects more transparent for users.

### DIFF
--- a/src/stratigraphy/benchmark/score.py
+++ b/src/stratigraphy/benchmark/score.py
@@ -10,7 +10,7 @@ from stratigraphy import DATAPATH
 from stratigraphy.annotations.draw import draw_predictions
 from stratigraphy.benchmark.ground_truth import GroundTruth
 from stratigraphy.benchmark.metrics import DatasetMetrics, DatasetMetricsCatalog, Metrics
-from stratigraphy.util.predictions import FilePredictions, OverallFilePredictions
+from stratigraphy.util.predictions import OverallFilePredictions
 from stratigraphy.util.util import parse_text
 
 load_dotenv()
@@ -100,7 +100,7 @@ def evaluate_borehole_extraction(
     return all_metrics
 
 
-def get_metrics(predictions: dict[str, FilePredictions], field_key: str, field_name: str) -> DatasetMetrics:
+def get_metrics(predictions: OverallFilePredictions, field_key: str, field_name: str) -> DatasetMetrics:
     """Get the metrics for a specific field in the predictions.
 
     Args:
@@ -113,8 +113,8 @@ def get_metrics(predictions: dict[str, FilePredictions], field_key: str, field_n
     """
     dataset_metrics = DatasetMetrics()
 
-    for file_name, file_prediction in predictions.items():
-        dataset_metrics.metrics[file_name] = getattr(file_prediction, field_key)[field_name]
+    for file_prediction in predictions.file_predictions_list:
+        dataset_metrics.metrics[file_prediction.file_name] = getattr(file_prediction, field_key)[field_name]
 
     return dataset_metrics
 
@@ -150,7 +150,8 @@ def evaluate_layer_extraction(
 
     for language, language_predictions in predictions_by_language.items():
         language_number_of_truth_values = {
-            file_name: number_of_truth_values[file_name] for file_name in language_predictions.file_predictions_list
+            prediction.file_name: number_of_truth_values[prediction.file_name]
+            for prediction in language_predictions.file_predictions_list
         }
         all_metrics.metrics[f"{language}_layer"] = get_layer_metrics(
             language_predictions, language_number_of_truth_values
--- a/src/stratigraphy/benchmark/score.py
+++ b/src/stratigraphy/benchmark/score.py
@@ -10,10 +10,7 @@ from stratigraphy import DATAPATH
 from stratigraphy.annotations.draw import draw_predictions
 from stratigraphy.benchmark.ground_truth import GroundTruth
 from stratigraphy.benchmark.metrics import DatasetMetrics, DatasetMetricsCatalog, Metrics
-from stratigraphy.evaluation.evaluation_dataclasses import OverallBoreholeMetadataMetrics
-from stratigraphy.evaluation.metadata_evaluator import MetadataEvaluator
-from stratigraphy.metadata.metadata import BoreholeMetadataList
-from stratigraphy.util.predictions import FilePredictions
+from stratigraphy.util.predictions import FilePredictions, OverallFilePredictions
 from stratigraphy.util.util import parse_text
 
 load_dotenv()
@@ -23,7 +20,7 @@ logging.basicConfig(format="%(asctime)s %(levelname)-8s %(message)s", level=logg
 logger = logging.getLogger(__name__)
 
 
-def get_layer_metrics(predictions: dict, number_of_truth_values: dict) -> DatasetMetrics:
+def get_layer_metrics(predictions: OverallFilePredictions, number_of_truth_values: dict) -> DatasetMetrics:
     """Calculate F1, precision and recall for the layer predictions.
 
     Calculate F1, precision and recall for the individual documents as well as overall.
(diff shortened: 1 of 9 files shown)


dcleres:
I think that currently, we have two ways of keeping track of information regarding a document. 

```python
self.metrics: dict[str, Metrics] = {}
```

and the second one is to build a class that has the filename as atrribute. 

For me, both are fine. I introduced the second one, but if you, @stijnvermeeren-swisstopo, prefer the first version, I am happy to keep it this way. I do believe we need to be consistent with it.

stijnvermeeren-swisstopo (changes requested):
I like it, makes the code significantly cleaner and easier to read and understand. I just have a few minor comments / ideas.

src/stratigraphy/benchmark/score.py
```diff
@@ -198,27 +193,21 @@ def create_predictions_objects(
...
-        metadata = metadata_per_file.get_metadata(file_name)
-        if not metadata:
-            raise ValueError(f"Metadata for file {file_name} not found.")
+    for file_predictions in predictions.file_predictions_list:
+        # prediction_object = FilePredictions.create_from_json(file_predictions, file_predictions.file_name)
 
-        prediction_object = FilePredictions.create_from_json(file_predictions, metadata, file_name)
-
-        predictions_objects[file_name] = prediction_object
+        # predictions_objects[file_name] = prediction_object
```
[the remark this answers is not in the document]
dcleres: Thank you. I commented it out to test whether the new code works but did not remove the code in case I had to roll back. However, since this version seems functional the commented code is not needed anymore.

src/stratigraphy/util/predictions.py
```diff
@@ -276,7 +202,72 @@ def _find_matching_layer(
...
+        return {
+            self.file_name: {
+                "metadata": self.metadata.to_json(),
+                "layers": [layer.to_json() for layer in self.layers],
+                "depths_materials_column_pairs": [
+                    depths_materials_columns_pairs.to_json()
+                    for depths_materials_columns_pairs in self.depths_materials_columns_pairs
+                ],
+                "page_dimensions": self.metadata.page_dimensions,
+                # TODO: This should be removed. As already in metadata.
```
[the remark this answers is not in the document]
dcleres: This applies to the page_dimensions. Do you have a preference where to place this information?

src/stratigraphy/main.py
```diff
@@ -280,25 +266,42 @@ def start_pipeline(
...
+                        )
+                    )
+                else:
+                    predictions.add_file_predictions(
+                        FilePredictions(
+                            file_name=filename,
+                            metadata=metadata,
+                            groundwater_entries=None,
+                            layers=None,
+                            depths_materials_columns_pairs=None,
```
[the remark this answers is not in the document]
dcleres: Thank you, I adjusted this.

src/stratigraphy/benchmark/score.py
```diff
@@ -198,27 +193,21 @@ def create_predictions_objects(
...
+                number_of_truth_values[file_predictions.file_name] = len(ground_truth_for_file["layers"])
 
-    return predictions_objects, number_of_truth_values
+    return predictions, number_of_truth_values
 
 
 def evaluate(
-    predictions,
-    metadata_per_file: BoreholeMetadataList,
+    predictions: OverallFilePredictions,
```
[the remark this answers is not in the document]
dcleres: I will address that as well

dcleres:
@stijnvermeeren-swisstopo I tried to work on the comments you made on the PR. I also made it possible to rerun the scoring script. To prevent it from being broken in a future PR, I added it to the tests as well. Looking forward what you think from the update.

src/stratigraphy/util/predictions.py
```diff
@@ -266,8 +260,40 @@ def evaluate_metadata_extraction(self, ground_truth_path: Path) -> OverallBoreho
...
+
+        Args:
+            prediction_from_file (dict): A dictionary representing the predictions.
+
+        Returns:
+            OverallFilePredictions: The object.
+        """
+        overall_file_predictions = OverallFilePredictions()
+        for file_name, file_data in prediction_from_file.items():
+            metadata = BoreholeMetadata.from_json(file_data.get(file_name)["metadata"], file_name)
```
stijnvermeeren-swisstopo, line 281: I would expect that you can just write `file_data[""metadata]` here, but we seem to have some "double nesting" of the filename in predictions.json. Do you know where this comes from?

image

src/stratigraphy/depths_materials_column_pairs/depths_materials_column_pairs.py
```diff
@@ -41,3 +41,19 @@ def to_json(self) -> dict:
...
+    def from_json(cls, json_depths_materials_column_pairs: dict) -> "DepthsMaterialsColumnPairs":
+        """Converts a dictionary to an object.
+
+        Args:
+            json_depths_materials_column_pairs (dict): A dictionary representing the depths materials column pairs.
+
+        Returns:
+            DepthsMaterialsColumnPairs: The depths materials column pairs object.
+        """
+        depth_column = DepthColumn.from_json(json_depths_materials_column_pairs["depth_column"])
```
stijnvermeeren-swisstopo, line 55: How does it know which implementation of the abstract `from_json` method to call here, the one from `BoundaryDepthColumn` or the one from `LayerDepthColumn`?

src/stratigraphy/depthcolumn/depthcolumn.py
```diff
@@ -53,14 +53,29 @@ def noise_count(self, all_words: list[TextWord]) -> int:
...
 
+    @abc.abstractmethod
     def to_json(self):
-        rect = self.rect()
-        return {
-            "rect": [rect.x0, rect.y0, rect.x1, rect.y1],
-            "entries": [entry.to_json() for entry in self.entries],
-        }
+        """Converts the object to a dictionary."""
+        pass
```
stijnvermeeren-swisstopo, line 72: I don't see why you had to move the implementation from the parent class to the child classes, the implementation still look identical to me, or am I missing something?

src/stratigraphy/benchmark/score.py
```diff
@@ -179,10 +186,9 @@ def create_predictions_objects(
     Args:
         predictions (dict): The predictions from the predictions.json file.
```
stijnvermeeren-swisstopo, line 187: wrong type

stijnvermeeren-swisstopo (changes requested):

stijnvermeeren-swisstopo:
Thanks for the work and for fixing the score.py script! Just a few smaller comments remaining :)

src/stratigraphy/groundwater/groundwater_extraction.py
```diff
@@ -110,6 +110,27 @@ def to_json(self) -> dict:
...
+                depth=json_groundwater_information_on_page["depth"],
+                date=json_groundwater_information_on_page["date"],
+                elevation=json_groundwater_information_on_page["elevation"],
+            ),
+            page=json_groundwater_information_on_page["page"],
+            rect=fitz.Rect(json_groundwater_information_on_page["rect"]),
+        )
+
     @staticmethod
     def from_json_values(date: str | None, depth: float | None, elevation: float | None, page: int, rect: list[float]):
```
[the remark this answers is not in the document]
dcleres: good catch, I checked and no it is indeed not used anymore same for the other method `from_json`

src/stratigraphy/util/predictions.py
dcleres: I do believe that this section of the code was not used in the end. However, you are right the `to_json` method of the `OverallFilePredictions` object was generating double-nested json files. I addressed this and also wrote some tests to ensure that the output format does not change unintentionally over time.

src/stratigraphy/depthcolumn/depthcolumn.py
dcleres: I moved it to the child classes as the `self.entries` attribute is defined in the child classes and does not exist in the parent class. We can leave it there if you want or maybe move the `entries` attribute to the parent class.

dcleres:
@stijnvermeeren-swisstopo I think I managed to address all of your comments. Some questions are still open, and I need your input on these (open conversations in the PR). Thank you very much for the review!

src/stratigraphy/depthcolumn/depthcolumn.py
stijnvermeeren-swisstopo: I see the point. Probably, the cleanest solution would be to define a common parent class X for LayerDepthColumnEntry and BoundaryDepthColumnEntry, and then define `entries: list[X]` in the DepthColumn parent class. However, since we don't have this yet, I would say that this might be a bit out-of-scope for this PR, and I'm happy to keep your implementation.

stijnvermeeren-swisstopo (approved):
I made one small change to the structure of the FilePrediction.to_json() return value.

LGTM now!

src/stratigraphy/depthcolumn/depthcolumn.py
dcleres: Awesome! Happy I managed to convince you! :-)

PR #85, opened 2024-09-24
Close #LGVISIUM-85: Not found features do now return a 404

dcleres:
Error+Detail instead of Success+Detail for extract_data if no data could be read out

### DIFF
--- a/src/app/api/v1/endpoints/extract_data.py
+++ b/src/app/api/v1/endpoints/extract_data.py
@@ -70,7 +70,7 @@ def extract_data(extract_data_request: ExtractDataRequest) -> ExtractDataRespons
         )
 
         if not extracted_coords:
-            return HTTPException(status_code=404, detail="Coordinates not found.")
+            raise HTTPException(status_code=404, detail="Coordinates not found.")
 
         # Convert the bounding box to PNG coordinates and return the response
         return ExtractCoordinatesResponse(
@@ -196,7 +196,7 @@ def extract_text(pdf_page: fitz.Page, user_defined_bbox: fitz.Rect) -> ExtractDa
         bbox = BoundingBox.load_from_fitz_rect(text_based_bbox)
         return ExtractTextResponse(bbox=bbox, text=text)
     else:
-        return HTTPException(status_code=404, detail="Text not found.")
+        raise HTTPException(status_code=404, detail="Text not found.")
 
 
 def extract_number(pdf_page: fitz.Page, user_defined_bbox: fitz.Rect) -> ExtractNumberResponse | HTTPException:
@@ -226,7 +226,7 @@ def extract_number(pdf_page: fitz.Page, user_defined_bbox: fitz.Rect) -> Extract
             )
             return ExtractNumberResponse(bbox=bbox, number=number[0])
 
-    return HTTPException(status_code=404, detail="Number not found.")
+    raise HTTPException(status_code=404, detail="Number not found.")
 
 
 def extract_number_from_text(text: str) -> list[float]:
--- a/src/app/api/v1/endpoints/extract_data.py
+++ b/src/app/api/v1/endpoints/extract_data.py
@@ -14,8 +14,8 @@ from app.common.schemas import (
     ExtractNumberResponse,
     ExtractTextResponse,
     FormatTypes,
-    NotFoundResponse,
 )
+from fastapi import HTTPException
 from stratigraphy.metadata.coordinate_extraction import CoordinateExtractor, LV03Coordinate, LV95Coordinate
 from stratigraphy.text.extract_text import extract_text_lines_from_bbox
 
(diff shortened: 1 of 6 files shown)


stijnvermeeren-swisstopo (changes requested):
How do we get these error responses to be properly documented in the Swagger / OpenAPI docs? At the moment, I only see documentation for the success response and for a 422 validation error (which I'm not sure we even use?).

image

src/app/api/v1/endpoints/extract_data.py
```diff
@@ -189,11 +192,14 @@ def extract_text(pdf_page: fitz.Page, user_defined_bbox: fitz.Rect) -> ExtractDa
         text += text_line.text + " "
         text_based_bbox = text_based_bbox + text_line.rect
 
-    bbox = BoundingBox.load_from_fitz_rect(text_based_bbox)
-    return ExtractTextResponse(bbox=bbox, text=text)
+    if text:
+        bbox = BoundingBox.load_from_fitz_rect(text_based_bbox)
+        return ExtractTextResponse(bbox=bbox, text=text)
+    else:
+        return HTTPException(status_code=404, detail="Text not found.")
```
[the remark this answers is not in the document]
dcleres: Thank you for catching this. I was also not super happy about the way the exceptions were handled when I worked on it. I will try again today to clean this up and see if, with fresh eyes, I am able to find a nicer solution.

dcleres:
@stijnvermeeren-swisstopo Thank you very much for the good code review. FastAPI is creating the 422 by default. I had to disable it manually by changing the OpenAPI Settings. Now the error code should only be the one that we can actually trigger. 

I also cleaned up the code regarding how we raise the different validation errors.

src/app/main.py
```diff
@@ -22,6 +23,33 @@
 app = FastAPI(root_path=f"/{root_path}")
 
 
+def custom_openapi():
+    """Custom function to modify the OpenAPI schema and remove 422 errors."""
```
stijnvermeeren-swisstopo, line 27: Please add a link to https://github.com/fastapi/fastapi/discussions/6695, which is where this implementation seems to come from, and is also where additional context can be found (as without additional context, this code looks really hacky and suspicious).

src/app/api/v1/endpoints/extract_data.py
```diff
@@ -168,7 +171,7 @@ def create_response(coord, srs):
     return None
 
 
-def extract_text(pdf_page: fitz.Page, user_defined_bbox: fitz.Rect) -> ExtractDataResponse:
+def extract_text(pdf_page: fitz.Page, user_defined_bbox: fitz.Rect) -> ExtractDataResponse | HTTPException:
```
stijnvermeeren-swisstopo, line 174: Can be reverted to just `ExtractDataResponse` as return type? Also, can't we be more specific and put `ExtractTextResponse` as return type?

src/app/api/v1/endpoints/extract_data.py
```diff
@@ -106,6 +106,9 @@ def extract_data(extract_data_request: ExtractDataRequest) -> ExtractDataRespons
     elif extract_data_request.format == FormatTypes.NUMBER:
         extracted_number = extract_number(pdf_page, user_defined_bbox.to_fitz_rect())
 
+        if isinstance(extracted_number, HTTPException):
+            return extracted_number
```
stijnvermeeren-swisstopo, line 110: No longer needed?

src/app/api/v1/endpoints/extract_data.py
```diff
@@ -93,6 +90,9 @@ def extract_data(extract_data_request: ExtractDataRequest) -> ExtractDataRespons
     elif extract_data_request.format == FormatTypes.TEXT:
         extracted_text: ExtractTextResponse = extract_text(pdf_page, user_defined_bbox.to_fitz_rect())
 
+        if isinstance(extracted_text, HTTPException):
+            return extracted_text
```
stijnvermeeren-swisstopo, line 94: No longer needed?

src/app/api/v1/endpoints/extract_data.py
```diff
@@ -189,11 +192,14 @@ def extract_text(pdf_page: fitz.Page, user_defined_bbox: fitz.Rect) -> ExtractDa
...
-    return ExtractTextResponse(bbox=bbox, text=text)
+    if text:
+        bbox = BoundingBox.load_from_fitz_rect(text_based_bbox)
+        return ExtractTextResponse(bbox=bbox, text=text)
+    else:
+        raise HTTPException(status_code=404, detail="Text not found.")
 
 
-def extract_number(pdf_page: fitz.Page, user_defined_bbox: fitz.Rect) -> ExtractNumberResponse:
+def extract_number(pdf_page: fitz.Page, user_defined_bbox: fitz.Rect) -> ExtractNumberResponse | HTTPException:
```
stijnvermeeren-swisstopo, line 202: Can be reverted to just `ExtractNumberResponse` as return type?

stijnvermeeren-swisstopo (changes requested):
Nice, this looks much cleaner already. Only the removal of the 422 response from the docs looks quite hacky, but as this is the recommended solution on the Github discussion, I'm happy to keep it, but I would just add a link to that discussion. Apart from that, here are just a few small bits that can probably be cleanup up before we merge.

src/app/main.py
dcleres: Yes, exactly. This is where I found the fix. I will add it to the code.

src/app/api/v1/endpoints/extract_data.py
dcleres: Indeed. Good catch!

src/app/api/v1/endpoints/extract_data.py
dcleres: Exactly!

stijnvermeeren-swisstopo (approved):

PR #86, opened 2024-09-25
Close #LGVISIUM-75: Made the extraction of information robust to page rotations

dcleres:
The current version of the API was not able to extract information from bounding boxes when the PyMuPDF library was recognizing a page rotation. This is addressed in this PR and tested with tests.


stijnvermeeren-swisstopo (approved):
The page rotation seems to be handled correctly now, thanks :)

While testing, I notices that the bbox was computed incorrectly when there are multiple lines of text, and also, that the text itself always contains a redundant whitespace at the end. I've added fixed for this to the code. Sorry for temporarily breaking the unit tests :).

I also noticed that the bbox from the response is not always contained in the bbox from the request. Sometimes, it is a little larger, as in the screenshot below (`x1` value). Do you understand why this can happen? (In general, I don't think this is a bad thing, but I just don't fully really understand the current behavior.)

image

dcleres:
@stijnvermeeren-swisstopo Thank you for the review and the improvements you made to the code. Very useful! 

Regarding your question, I thought that I was observing something similar to yours, but I assumed that my bounding box was not well defined and, therefore, had more text than expected. 

This is the line that contains all the "magic": 

`for x0, y0, x1, y1, word, block_no, line_no, _word_no in page.get_text("words", clip=bbox)` -> in `src/stratigraphy/text/extract_text.py`

The documentation is not entirely transparent on the behavior of the function: https://pymupdf.readthedocs.io/en/latest/annot.html#Annot.get_text 

In this discussion, they discuss whether single characters are included in the extraction (https://github.com/pymupdf/PyMuPDF/discussions/3067). 

Regarding the size of the bbox that is returned. I am returning the bbox of what was extracted. Therefore it is normal that the size of the returned bbox is different to the one provided by the user. 

I hope this answers your question. I could not find much more information about this function from the documentation and the GitHub issues. There is also a Discord channel, but I have not looked into it yet.

stijnvermeeren-swisstopo:
@dcleres Thanks for looking into this. I think that we need to look into this topic in more detail. There already was a discussion on Teams about the clipping behavior as well, which has not been fully addressed in the implementation yet. I created a new ticket for this: https://jira.swisstopo.ch/browse/LGVISIUM-89

PR #87, opened 2024-09-30
Close #LGVISIUM-80: Refactor the groundwater object and its evaluation

dcleres:
I restructured the code to use the structure from the groundwater ticket better. 

The idea is that each feature is represented consistently at different levels (page, document, dataset) for all items. 

By doing so, we have: 

Class `Feature`: Contains the individual information about the feature (e.g., Groundwater) 
Class `FeatureOnPage`: contains the information on a feature, together with the number of the page and the location on that page where the feature was extracted from
Class `FeatureInDocument`: contains a list of FeatureOnPage instances, for features that might be found multiple times in a single document
Class `OverallFeature`: contains a list of `FeatureInDocument` and is able to compute metrics at the dataset level. 

Currently, this is done for Groundwater, but it needs to be extended to the other data classes. 

The FilePredictions objects are also directly building from Objects and not dictionary representations of objects.


src/stratigraphy/benchmark/metrics.py
```diff
@@ -53,23 +53,62 @@ def to_dataframe(self, name: str, fn: Callable[[Metrics], float]) -> pd.DataFram
...
+
+    def set_layer_metrics(self, metrics: OverallMetrics):
+        """Set the layer metrics."""
+        self.layer_metrics = metrics
+
+    def set_depth_interval_metrics(self, metrics: OverallMetrics):
+        """Set the depth interval metrics."""
+        self.depth_interval_metrics = metrics
+
+    def set_de_layer_metrics(self, metrics: OverallMetrics):
```
dcleres, line 76: Having one setter per language might be a bit too much. We could think of passing the language as a parameter. This would save us the declaration of four functions.

src/stratigraphy/metadata/metadata.py
```diff
@@ -107,6 +107,14 @@ def __init__(self):
         """Initializes the StratigraphyMetadata object."""
         self.metadata_per_file = []
 
+    def add_metadata(self, metadata: BoreholeMetadata):
```
dcleres, line 110: The question for the metadata to me is whether we would like to copy the behavior we now have for the layers and groundwatr with having:

Elevation
ElevationOnPage
ElevationInDocument
OverallElevation

Or if we keep everything in the Metadata Object.

src/stratigraphy/benchmark/metrics.py
stijnvermeeren-swisstopo: Is is even necessary to define these setter methods at all? It's not like the class' instance variables are private.

src/stratigraphy/depthcolumn/depthcolumn.py
```diff
@@ -130,10 +174,21 @@ def is_valid(self) -> bool:
     def identify_groups(
         self,
         description_lines: list[TextLine],
-        geometric_lines: list[Line],
-        material_description_rect: fitz.Rect,
+        geometric_lines: list[Line],  # TODO: Parameter not used. Shall we keep?
+        material_description_rect: fitz.Rect,  # TODO: Parameter not used. Shall we keep?
```
stijnvermeeren-swisstopo, line 178: These are used in the other implementation of this method, the one from BoundaryDepthColumn, not LayerDepthColumn. And since they both need to conform to the signature of the same abstract method from the parent class DepthColumn, these parameter are also required here. Of do you know of a better way to deal with this situation?

src/stratigraphy/depthcolumn/depthcolumn.py
```diff
@@ -148,8 +203,11 @@ def identify_groups(
 
             matched_blocks = interval.matching_blocks(description_lines, line_index, next_interval)
             line_index += sum([len(block.lines) for block in matched_blocks])
-            groups.append({"depth_intervals": [interval], "blocks": matched_blocks})
-
+            groups.append(
+                # TODO: This seems to be the only case where a list is passed and most of the time it is a list of one
+                # element. Seem to need the function: transform_groups().
```
stijnvermeeren-swisstopo, line 208: In the identify_groups implementation of the BoundaryDepthColumn subclass, we definitely have some cases were we pass a list of more than one depth interval.

However, I agree with you that we need to clean up the IntervalBlockGroup class, as it is very confusing to allow both a list as well as a single element for the parameters. Either we only allow a list, or we create separate classes for when we need a list and when we need a single element. I'm not sure yet, what would be the best option.

src/stratigraphy/extract.py
```diff
@@ -138,24 +142,41 @@ def process_page(
...
+    )
+    layer_predictions.remove_empty_predictions()
+    return layer_predictions, json_filtered_pairs
+
+
+def convert_to_boundary_interval(layer_depth_column: LayerDepthColumnEntry) -> BoundaryInterval:
+    """Convert a dictionary to a BoundaryInterval object.
+
+    Args:
+        layer_depth_column (LayerDepthColumnEntry): The dictionary to convert.
```
stijnvermeeren-swisstopo, line 171: this is not a dictionary?

src/stratigraphy/extract.py
```diff
@@ -33,9 +34,13 @@
 
 def process_page(
     lines: list[TextLine], geometric_lines, language: str, page_number: int, **params: dict
-) -> list[dict]:
+) -> tuple[LayersOnPage, list[DepthsMaterialsColumnPairs]]:
     """Process a single page of a pdf.
 
+    # TODO: Ideally, one function does one thing. This function does a lot of things. It should be split into
+    # smaller functions.
+    # TODO: Return objects instead of dictionaries. This will make the code more readable and easier to understand.
```
stijnvermeeren-swisstopo, line 42: agreed

stijnvermeeren-swisstopo (changes requested):
Not a full review yet, but a few comments to start with.

src/stratigraphy/benchmark/metrics.py
dcleres: Yes, that is actually true. I will remove and replace this.

dcleres:
@stijnvermeeren-swisstopo I was able to refactor the code to cope with the changes made in the previous refactoring issue. I think the code is ready to see a second pair of eyes, and I look forward to your feedback. If we merge this code, we will remove almost all dictionary objects from the code and replace them with data objects. 

I still identified a few TODOs in the code, and I flagged them to make sure that we look into them in the future. Most of them seemed a bit out of scope for me for this PR. However, if you see some things that need urgent improvement, I can address these as well.

src/stratigraphy/util/predictions.py
```diff
@@ -265,31 +162,232 @@ def from_json(cls, prediction_from_file: dict) -> "OverallFilePredictions":
...
                     file_name=file_name,
                     metadata=metadata,
                     depths_materials_columns_pairs=depths_materials_columns_pairs,
-                    groundwater_entries=groundwater_entries,
+                    groundwater=groundwater_in_document,
                 )
             )
         return overall_file_predictions
 
+    def get_groundwater_entries(self) -> list[GroundwaterInDocument]:
```
stijnvermeeren-swisstopo, line 190: is there any value in defining a dedicated method for this one-liner?

src/stratigraphy/util/predictions.py
```diff
@@ -265,31 +162,232 @@ def from_json(cls, prediction_from_file: dict) -> "OverallFilePredictions":
...
+        """
+        overall_metrics = OverallMetrics()
+
+        for file_prediction in self.file_predictions_list:
+            overall_metrics.metrics[file_prediction.file_name] = getattr(file_prediction, field_key)[field_name]
+
+        return overall_metrics
+
+
+def get_layer_metrics(predictions: OverallFilePredictions, number_of_truth_values: dict) -> OverallMetrics:
```
stijnvermeeren-swisstopo, line 329: A lot of duplicated methods between this file and `score.py`. Did you mean to remove the methods in score.py, or what is the point of copying this methods into this file?

stijnvermeeren-swisstopo (changes requested):

stijnvermeeren-swisstopo:
If I understand correctly, the PR description is not entirely correct. Better would be:
- Class `FeatureOnPage`: contains the information on a feature, together with the number of the page and the location on that page where the feature was extracted from
- Class `FeatureInDocument`: contains a list of FeatureOnPage instances, for features that might be found multiple times in a single document

stijnvermeeren-swisstopo:
I made a small review proposal here: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/101

src/stratigraphy/benchmark/metrics.py
```diff
@@ -58,23 +58,30 @@ def get_metrics_list(self) -> list[Metrics]:
...
     def __init__(self):
-        self.metrics: dict[str, DatasetMetrics] = {}
+        self.layer_metrics: OverallMetrics = None
+        self.depth_interval_metrics: OverallMetrics = None
+        self.de_layer_metrics: OverallMetrics = None
+        self.de_depth_interval_metrics: OverallMetrics = None
+        self.fr_layer_metrics: OverallMetrics = None
+        self.fr_depth_interval_metrics: OverallMetrics = None
+        self.groundwater_metrics: OverallMetrics = None
+        self.groundwater_depth_metrics: OverallMetrics = None
```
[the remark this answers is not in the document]
dcleres: empty OverallMetrics() instance sounds like a good way to go. Thank you.

src/stratigraphy/benchmark/metrics.py
```diff
@@ -87,20 +94,18 @@ def metrics_dict(self) -> dict[str, float]:
...
             {
-                "F1": self.metrics.get("layer", DatasetMetrics()).pseudo_macro_f1(),
-                "recall": self.metrics.get("layer", DatasetMetrics()).macro_recall(),
-                "precision": self.metrics.get("layer", DatasetMetrics()).macro_precision(),
-                "depth_interval_accuracy": self.metrics.get("depth_interval", DatasetMetrics()).macro_precision(),
+                "F1": self.layer_metrics.pseudo_macro_f1() if self.layer_metrics else None,
+                "recall": self.layer_metrics.macro_recall() if self.layer_metrics else None,
+                "precision": self.layer_metrics.macro_precision() if self.layer_metrics else None,
+                "depth_interval_accuracy": self.depth_interval_metrics.macro_precision()
+                if self.layer_metrics
```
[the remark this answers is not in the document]
dcleres: Indeed, thank you!

src/stratigraphy/benchmark/metrics.py
```diff
@@ -58,23 +58,30 @@ def get_metrics_list(self) -> list[Metrics]:
...
     """Keeps track of all different relevant metrics that are computed for a dataset."""
 
     def __init__(self):
-        self.metrics: dict[str, DatasetMetrics] = {}
+        self.layer_metrics: OverallMetrics = None
+        self.depth_interval_metrics: OverallMetrics = None
+        self.de_layer_metrics: OverallMetrics = None
+        self.de_depth_interval_metrics: OverallMetrics = None
+        self.fr_layer_metrics: OverallMetrics = None
+        self.fr_depth_interval_metrics: OverallMetrics = None
```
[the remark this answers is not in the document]
dcleres: Yes, I also had problems with that. I implemented an alternative now where the Metric Catalog is initialized with languages that are previously automatically detected. Let me know your thoughts about it.

src/stratigraphy/depthcolumn/depthcolumn.py
```diff
@@ -83,11 +98,21 @@ class DepthColumnFactory:
...
+        Returns:
+            DepthColumn: The depth column object.
+        """
         column_type = data.get("type")
         if column_type == "BoundaryDepthColumn":
             return BoundaryDepthColumn.from_json(data)
         elif column_type == "LayerDepthColumn":
             return LayerDepthColumn.from_json(data)
+        elif column_type == "LayerIdentifierColumn":
+            return LayerIdentifierColumn.from_json(data)
```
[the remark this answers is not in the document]
dcleres: I think it would be better. However, I thought the refactoring might be out of scope here, and I did not know whether it would make sense. If i does make sense, I can refactor that class with you, or we can open a follow-up ticket that aims to clean these classes and their inheritance patterns.

src/stratigraphy/extract.py
```diff
@@ -147,22 +150,41 @@ def process_page(
...
+    Args:
+        layer_depth_column (LayerDepthColumnEntry): The layer depth column entry.
+
+    Returns:
+        BoundaryInterval: The converted boundary interval.
+    """
+    start = layer_depth_column.start if layer_depth_column.start is not None else None
+    end = layer_depth_column.end if layer_depth_column.end is not None else None
+
+    return BoundaryInterval(start=start, end=end)
```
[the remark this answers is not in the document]
dcleres: Indeed. The attributes are never none. Good catch.

src/stratigraphy/util/predictions.py
dcleres: I initially thought that it would make the code a bit clearer. However, since we are not doing any processing, I am happy to remove that function. I do believe your code in branch `LGVISIUM-80/review-stijn` already fixes this.

src/stratigraphy/util/predictions.py
dcleres: Yes, indeed I was moving code from `score.py` to the prediction file. I missed removing the unused code.

tests/test_predictions.py
```diff
@@ -85,19 +88,6 @@ def test_overall_file_predictions():
     assert result == {"test_file": {"some_data": "test"}}
 
 
-def test_evaluate_groundwater(sample_file_prediction):
```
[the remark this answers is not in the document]
dcleres: This was indeed missing in the tests. I build a seperate testing file for the groundwater.

tests/test_predictions.py
```diff
@@ -62,17 +76,6 @@ def test_to_json(sample_file_prediction):
     assert result["metadata"]["coordinates"]["E"] == 2789456
 
 
-def test_count_against_ground_truth():
```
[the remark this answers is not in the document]
dcleres: I put it back. Thank you.

tests/test_predictions.py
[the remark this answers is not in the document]
dcleres: This was moved to a more complete series of tests that can be found here: tests/test_groundwater.py

dcleres:
@stijnvermeeren-swisstopo I edited the description of the PR and addressed your comment. I am sorry I forced-push when I wanted to bring my code from the fork to the main repo. I did only find a better way to do it later. I hope you will still be able to see the changes I did.

dcleres:
Just as FYI, the issue we discussed with `.groundwater.groundwater`, which can be confusing, was not addressed yet. I will handle that case in the morning.

src/stratigraphy/main.py
```diff
@@ -225,12 +225,18 @@ def start_pipeline(
...
+                # Initialize common variables
+                groundwater_entries = None
+                layers = None
+                depths_materials_columns_pairs = None
+
                 if part == "all":
                     # Extract the groundwater levels
-                    groundwater_extractor = GroundwaterLevelExtractor(document=doc)
-                    groundwater = groundwater_extractor.extract_groundwater(terrain_elevation=metadata.elevation)
+                    groundwater_in_document = GroundwaterInDocument.from_document(doc, metadata.elevation)
```
stijnvermeeren-swisstopo, line 236: Directly assign the result to `groundwater_entries`, instead of using this redundant intermediate variable `groundwater_in_document`, and copying the value later on.

Same for `layers` and `depths_materials_columns_pairs`.

stijnvermeeren-swisstopo (changes requested):

src/stratigraphy/evaluation/groundwater_evaluator.py
```diff
@@ -0,0 +1,128 @@
...
+                [entry.format_date() for entry in gt_groundwater],
+            )
+
+            file_groundwater_metrics = GroundwaterMetrics(
+                groundwater_metrics=groundwater_metrics,
+                groundwater_depth_metrics=groundwater_depth_metrics,
+                groundwater_elevation_metrics=groundwater_elevation_metrics,
+                groundwater_date_metrics=groundwater_date_metrics,
+                filename=filename,
+            )  # TODO: This clashes with the OverallMetrics object
```
stijnvermeeren-swisstopo, line 124: how do you mean this exactly, what is the "clash" that you are seeing?

src/stratigraphy/benchmark/metrics.py
```diff
@@ -58,23 +58,32 @@ def get_metrics_list(self) -> list[Metrics]:
         return list(self.metrics.values())
 
 
-class DatasetMetricsCatalog:
+class OverallMetricsCatalog:
     """Keeps track of all different relevant metrics that are computed for a dataset."""
 
-    def __init__(self):
-        self.metrics: dict[str, DatasetMetrics] = {}
+    def __init__(self, languages: list[str]):
```
stijnvermeeren-swisstopo, line 64: `languages`  has type `set[str]` where we call this constructor in predictions.py.

src/stratigraphy/benchmark/score.py
```diff
@@ -21,162 +21,6 @@
...
-        all_metrics.metrics["layer"].macro_f1() * 100,
-        all_metrics.metrics["layer"].macro_precision() * 100,
-        all_metrics.metrics["layer"].macro_recall() * 100,
-        all_metrics.metrics["depth_interval"].macro_precision() * 100,
-    )
-
-    return all_metrics
-
-
 def create_predictions_objects(
```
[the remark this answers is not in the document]
dcleres: Yes, indeed, we can remove this. Thank you for identifying this.

src/stratigraphy/main.py
dcleres: Makes sense!

src/stratigraphy/benchmark/metrics.py
dcleres: I just checked as well, and you are correct. Thank you!

src/stratigraphy/evaluation/groundwater_evaluator.py
dcleres: I was not super happy with the fact that I could not directly re-use the `OverallMetrics` class and had to create this new class as we also wanted to unify the evaluation across the different objects. However, I think it could be worse. If you wish, we can open a ticket to look into this. I do believe this ticket is a bit out of scope.

dcleres:
Thank you for your feedback @stijnvermeeren-swisstopo I created the follow-up ticket for the`LayerIdentifierColumn` class ([LGVISIUM-102](https://jira.swisstopo.ch/browse/LGVISIUM-102))

I also addressed and clarified based on all your comments.

stijnvermeeren-swisstopo (approved):
LGTM!

PR #88, opened 2024-10-01
Close #LGVISIUM-83: Extract coordinates with non integer values

dcleres:
Made it possible to extract digits from the coordinates by adapting the regex code. Adapted the tests to this new case and uploaded a new test file to test these cases.


src/stratigraphy/annotations/draw.py
```diff
@@ -138,10 +138,10 @@ def draw_metadata(
     """
     # TODO associate correctness with the extracted coordinates in a better way
     coordinate_color = "green" if is_coordinate_correct else "red"
-    coordinate_rect = fitz.Rect([5, 5, 200, 25])
+    coordinate_rect = fitz.Rect([5, 5, 250, 30])
```
dcleres, line 141: I needed to increase the size of the boxes; otherwise, the coordinates would not be printed because the box was too small.

src/stratigraphy/metadata/coordinate_extraction.py
```diff
@@ -247,6 +263,18 @@ def _match_text_with_rect(
         lines_with_position = []
         for line in lines:
             preprocessed_text = preprocess(line.text)
+
+            # Special case for coordinates in the file: 269126143-bp.pdf. The matched string is 269578211260032 because
+            # the OCR has made a mistake between '11' and '/1'.
+            # Regex pattern to match exactly 7 digits and no '/'
+            pattern_ = r"^(?!.*\/)(?=.*\d{7})[A-Za-z0-9]*\d{7}[A-Za-z0-9]*$"
```
dcleres, line 270: I'm not happy about this change, but there was no single regex command I could find that would preserve the performance. I'm happy to spend more time trying, but I want to show you this version first, @stijnvermeeren-swisstopo.

stijnvermeeren-swisstopo (changes requested):
Having looked at it a little more, I think that I do have a solution that keeps the performance status-quo without requiring the additional preprocessing hack. I would use the following regex:
```
r"(?:([12])[\.\s'‘’]{0,2})?(\d{3})[\.\s'‘’]{0,2}(\d{3})(?:\.(\d{1,}))?"
```
The difference is that IF we extract decimal digits, THEN the decimal point is required. The current regex has the decimal point always as optional, regardless of whether we have some decimal digits or not.

src/stratigraphy/metadata/coordinate_extraction.py
```diff
@@ -14,14 +14,17 @@
 
 logger = logging.getLogger(__name__)
 
-COORDINATE_ENTRY_REGEX = r"(?:([12])[\.\s'‘’]{0,2})?(\d{3})[\.\s'‘’]{0,2}(\d{3})\.?\d?"
+# COORDINATE_ENTRY_REGEX = r"(?:([12])[\.\s'‘’]{0,2})?(\d{3})[\.\s'‘’]{0,2}(\d{3})\.?\d?"
+# COORDINATE_ENTRY_REGEX = r"(?:([12])[\.\s'‘’]{0,2})?(\d{3})[\.\s'‘’]{0,2}(\d{1,3}))?"
+COORDINATE_ENTRY_REGEX = r"(?:([12])[\.\s'‘’]{0,2})?(\d{3})[\.\s'‘’]{0,2}(\d{3})\.?(\d{1,})?"
+# COORDINATE_ENTRY_REGEX = r"(?:([12])[\.\s'‘’]{0,2})?(\d{3})[\.\s'‘’]{0,2}(\d{3})(?:\.(\d{1,}))?"
```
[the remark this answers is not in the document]
dcleres: No, these are indeed not needed.

src/stratigraphy/metadata/coordinate_extraction.py
```diff
@@ -104,7 +118,9 @@ class LV95Coordinate(Coordinate):
 
     def is_valid(self):
         """Reference: https://de.wikipedia.org/wiki/Schweizer_Landeskoordinaten#Beispielkoordinaten."""
-        return 2324800 < self.east.coordinate_value < 2847500 and 1074000 < self.north.coordinate_value < 1302000
+        return (
+            2324800.0 < self.east.coordinate_value < 2847500.0 and 1074000.0 < self.north.coordinate_value < 1302000.0
```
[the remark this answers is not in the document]
dcleres: I wanted to add the digit to make sure that we are comparing floats with floats and to make it clear to the user. If that is okay, I'd like to keep it.

tests/test_coordinate_extraction.py
```diff
@@ -211,3 +222,35 @@ def test_CoordinateExtractor_get_coordinates_from_lines_rect():  # noqa: D103
...
     assert coordinates[0].page == 1
+
+    # Example from 269126143-bp.pdf (a slash in the middle of the coordinates as misread by OCR as the digit 1)
+    lines = _create_simple_lines(["269578211260032"])
+    coordinates = extractor.get_coordinates_from_lines(lines, page=1)
+    assert coordinates[0].east.coordinate_value == 2695782
+    assert coordinates[0].north.coordinate_value == 1260032
+
+
+def test_get_signle_decimal_coordinates():
```
dcleres, line 233: @stijnvermeeren-swisstopo, if you prefer, we could also extend the `test_CoordinateExtractor_get_coordinates_from_lines_rect` test.

tests/test_coordinate_extraction.py
```diff
@@ -228,3 +228,29 @@ def test_CoordinateExtractor_get_coordinates_from_lines_rect():  # noqa: D103
     coordinates = extractor.get_coordinates_from_lines(lines, page=1)
     assert coordinates[0].east.coordinate_value == 2695782
     assert coordinates[0].north.coordinate_value == 1260032
+
+
+def test_get_signle_decimal_coordinates():
+    """Test the extraction of decimal coordinates from a list of text lines."""
+    lines = _create_simple_lines(["start", "615.790.6 / 157.500.5", "end"])
```
stijnvermeeren-swisstopo, line 235: Maybe the "start" and "end" don't need to be included in every test? It's good to test that the coordinates can be found within other text lines, but I would only test this once (in `test_CoordinateExtractor_get_coordinates_from_lines_rect`). Testing that the extraction of decimal values works, should ideally test just that, and don't mix in other redundant context such as these "start" and "end" lines.

tests/test_coordinate_extraction.py
stijnvermeeren-swisstopo: I have no clear preference, like this is fine with me as well.

stijnvermeeren-swisstopo (changes requested):

tests/test_coordinate_extraction.py
dcleres: Understood! Thank you!

stijnvermeeren-swisstopo (approved):
LGTM!

PR #91, opened 2024-10-03
Close #LGVISIUM-77: Wrong elevation value selected within bounding box

dcleres:
Started to sort vertically the different text lines matched around the matched keyword. By sorting, it was possible to test which was the closest to the keyword first and thus reduce the number of false positives or errors in elevation and groundwater detection. 

@stijnvermeeren-swisstopo In today's stand-up, we discussed the possibility of vertically and horizontally sorting the text. Currently, all the issues I found are solved with this code in combination with the improvements you made in the OCR processing. However, if you wish, I can also try this technique. 

Screenshot 2024-10-03 at 09 30 24

The metrics from the `main` branch are in silent-quail-641, and the performance of this branch is `ambitious-mule-939`

### DIFF
--- a/src/stratigraphy/data_extractor/data_extractor.py
+++ b/src/stratigraphy/data_extractor/data_extractor.py
@@ -10,7 +10,7 @@ from dataclasses import dataclass
 import fitz
 import regex
 from stratigraphy.lines.line import TextLine
-from stratigraphy.util.util import calculate_distance, read_params
+from stratigraphy.util.util import read_params
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


src/stratigraphy/data_extractor/data_extractor.py
```diff
@@ -130,6 +130,11 @@ def get_lines_near_key(self, lines, key_line: TextLine) -> list[TextLine]:
         )
         feature_lines = [line for line in lines if line.rect.intersects(elevation_search_rect)]
 
-        # makes sure the line with the key is included first in the extracted information and the duplicate removed
+        # Insert key_line first and remove duplicates
         feature_lines.insert(0, key_line)
-        return list(dict.fromkeys(feature_lines))
+        feature_lines = list(dict.fromkeys(feature_lines))
+
+        # Sort by distance from key_line
```
stijnvermeeren-swisstopo, line 137:
```suggestion
        # Sort by vertical distance between the top of the feature line and the top of key_line
```

stijnvermeeren-swisstopo (changes requested):
Could you add a comment to https://jira.swisstopo.ch/browse/LGVISIUM-77 to describe the decision to only sort by vertical distance, as this already improves the results sufficiently, and that we therefore did not yet test the idea of looking horizontally and vertically (but not diagonally) from the key line.

dcleres:
@stijnvermeeren-swisstopo I added a description on the rational of only sorting vertically to Jira and how this could be improved.

stijnvermeeren-swisstopo (approved):
Thanks for adding the comment on the Jira ticket. LGTM!

PR #93, opened 2024-10-08
Close #LGVISIUM-92: Add GitHub actions and push Docker images to GHCR

dcleres:
This PR makes it possible to generate the Docker Container of the Borehole data extraction API in a similar way to https://github.com/swisstopo/swissgeol-boreholes-suite. 

Upon merging to the main, a GitHub Actions creates automatically creates a new container with the latest version of the code and an `edge` tag. The container is then stored in the GitHub Container Registry.


.github/workflows/release.yml
```diff
@@ -0,0 +1,96 @@
...
+            -f body="$(./get-changelog.sh)"
+
+      - name: Checkout new branch and patch changelog
+        run: |
+          git checkout -b $GIT_BRANCH_NAME
+          sed -i "/^\#\# \[Unreleased\]$/a \\\n\#\# $TAG_NAME - $(date '+%Y-%m-%d')" CHANGELOG.md
+
+      - name: Commit, push and create pull request
+        run: |
+          git config --global user.email "office@geowerkstatt.ch"
```
dcleres, line 92: @stijnvermeeren-swisstopo is there a similar address at SwissTopo that I could use?

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: Collect Docker image metadata (api)
+        id: meta-api
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.BASE_IMAGE_NAME }}-api
+          labels: |
+            org.opencontainers.image.created=${{ env.COMMITED_AT }}
+            org.opencontainers.image.version=v${{ env.VERSION }}
+            org.opencontainers.image.maintainer=Stijn Vermeeren <stijn.vermeeren@swisstopo.ch>
```
dcleres, line 36: Is it okay to provide your address here @stijnvermeeren-swisstopo  ? I could also use a generic contact address at SwissTopo

dcleres:
@stijnvermeeren-swisstopo I do believe that we made all the necessary initial checks with Oliver. There are just two open questions for you regarding how to reach a maintainer in case of problems.

.github/workflows/pre-release.yml
stijnvermeeren-swisstopo: Yes, this is fine.
However, I noticed that https://github.com/opencontainers/image-spec/blob/main/annotations.md lists a keyorg.opencontainers.image.authors but no key org.opencontainers.image.maintainer. Should we adjust this accordingly?

.github/workflows/release.yml
stijnvermeeren-swisstopo: No, I don't think so. Feel free to use my own swisstopo email address.

.github/workflows/release.yml
```diff
@@ -0,0 +1,96 @@
...
+          git config --global user.name "SwissTopo-Build"
```
stijnvermeeren-swisstopo, line 93: Please always write swisstopo in lowercase. Alternatively, it would also be ok to just name it "Github build action" or similar.

.github/workflows/release.yml
```diff
@@ -0,0 +1,96 @@
...
+          echo GIT_COMMIT_MESSAGE=Mark version ${TAG_NAME#v} as released >> $GITHUB_ENV
+          echo RELEASE_ID=$(gh api -H "Accept: application/vnd.github+json" /repos/${GITHUB_REPOSITORY}/releases/tags/${TAG_NAME} | jq '.id') >> $GITHUB_ENV
+
+      - name: Get changelog for this specific release and update release notes
+        run: |
+          gh api \
+            --method PATCH \
+            --header "Accept: application/vnd.github+json" \
+            /repos/${GITHUB_REPOSITORY}/releases/${RELEASE_ID} \
+            -f body="$(./get-changelog.sh)"
```
stijnvermeeren-swisstopo, line 83: where is this get-changelog.sh script coming from?

stijnvermeeren-swisstopo (changes requested):

.github/workflows/pre-release.yml
dcleres: I took it from the other GitHub project. However, you are right, I cannot find it in the documentation as well.

.github/workflows/release.yml
dcleres: It was missing from the code in the fork. I added it now. Thank you for the catch.

.github/workflows/release.yml
```diff
@@ -0,0 +1,96 @@
...
```
dcleres, line 83: I will discuss with Oliver in detail how this script works. Even without the script, the pipeline was running which is surprizing

.github/workflows/release.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           labels: |
             org.opencontainers.image.created=${{ env.COMMITED_AT }}
             org.opencontainers.image.version=v${{ env.VERSION }}
-            org.opencontainers.image.maintainer=Stijn Vermeeren <stijn.vermeeren@swisstopo.ch>
+            org.opencontainers.image.author=Stijn Vermeeren <stijn.vermeeren@swisstopo.ch>
```
stijnvermeeren-swisstopo, line 40: Should be `authors` in plural.

stijnvermeeren-swisstopo (changes requested):

.github/workflows/release.yml
stijnvermeeren-swisstopo: I guess that without the script, it just skips the automatic updating of the changelog?

dcleres: Yes, I would have expected the pipeline to crash, but the behavior you described is the one observed.

stijnvermeeren-swisstopo (approved):

PR #96, opened 2024-10-16
Close #LGVISIUM-94: Renaming of the output from create png api path

dcleres:
As the API is returning the key within the S3 bucket where the created PNG file is stored, it felt necessary to adapt the naming in the API response as well to avoid any type of confusion regarding what is returned.

### DIFF
--- a/src/app/common/schemas.py
+++ b/src/app/common/schemas.py
@@ -42,7 +42,7 @@ class PNGRequest(BaseModel):
 class PNGResponse(BaseModel):
     """Response schema for the create_pngs endpoint."""
 
-    key: list[str]  # key in the S3 bucket
+    keys: list[str]  # keys in the S3 bucket
 
 
 ########################################################################################################################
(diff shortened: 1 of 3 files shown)


dcleres:
[@flenny ](https://github.com/flenny) do you think you can adapt the borehole application to cope with this future change?

src/app/common/schemas.py
```diff
@@ -42,7 +42,7 @@ class Config:
 class PNGResponse(BaseModel):
     """Response schema for the create_pngs endpoint."""
 
-    png_urls: list[str]
+    key: list[str]  # key in the S3 bucket
```
stijnvermeeren-swisstopo, line 45: Maybe `keys` (plural) would be better than `key` , since we have a list of values, not just a single value?

stijnvermeeren-swisstopo (changes requested):

src/app/common/schemas.py
dcleres: I think this is a good idea!

dcleres:
@stijnvermeeren-swisstopo, thank you for your feedback. I made the required changes and re-ran the tests to see if everything was alright.

stijnvermeeren-swisstopo (approved):
LGTM!

PR #99, opened 2024-10-21
Close #LGVISIUM-101: Removed deprecated AWS Lambda script and removed the AWS Region from the environement variables

dcleres:
After working together with Oliver from Geowerkstatt, we were able to find a setup that makes it possible to integrate the data extraction API into the borehole application. This mainly involved figuring out which environment variables AWS needed to connect to the right S3 bucket in the correct region. 

As it is also very unlikely that we will deploy the API using AWS Lambda in the near future I removed the code to avoid any sort of confusion.


.env.template
```diff
@@ -4,3 +4,4 @@ MLFLOW_TRACKING_URI="http://127.0.0.1:5000"
 AWS_ACCESS_KEY_ID=your_access_key_id
 AWS_SECRET_ACCESS_KEY=your_secret_access_key
 AWS_ENDPOINT=your_endpoint_url
+AWS_S3_BUCKET=your_bucket_name
```
stijnvermeeren-swisstopo, line 7: Could you clarify why we need both AWS_ENDPOINT and AWS_S3_BUCKET? Is there any use case where these might refer to different buckets? Or if they always refer to the same bucket, then why can't we always connect using the endpoint URL (which contains the bucket name, if I recall correctly)?

README.md
```diff
@@ -425,11 +416,12 @@ Another option is to store the credentials in a .env file and load them into you
 AWS_ACCESS_KEY_ID=your_access_key_id
 AWS_SECRET_ACCESS_KEY=your_secret_access_key
 AWS_ENDPOINT=your_endpoint_url
+AWS_S3_BUCKET=your_bucket_name
 ```
 
 You can find an example for such a `.env` file in `.env.template`. If you rename this file to `.env` and add your AWS credentials you should be good to go. 
 
-5. **Access the API**
+1. **Access the API**
```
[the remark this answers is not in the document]
dcleres: I think VSCode is trying to be smart.

.env.template
dcleres: Geowerkstatt is using this structure for its environment variables, so I need to comply with it. When we developed the code, we passed all the information into the `AWS_ENDPOINT` variable, but Oliver told me that they prefer to do it this way. I think that this indeed leaves you the flexibility to open several S3 buckets from the same AWS account by adapting the bucket name. 

If you have a strong opinion about having everything in the `AWS_ENDPOINT` then I can ask Oliver whether it is possible for them to adapt their code.

dcleres:
@stijnvermeeren-swisstopo I added the information provided by Oliver from Geowerkstatt to the README.

README.md
```diff
@@ -352,61 +353,74 @@ This command will start the container and map port 8000 of the container to port
...
+- In some environments, the `AWS_ENDPOINT` may look like this:
+    
+    a. **Kubernetes cluster environment**
+
+    ```bash
+    AWS_ENDPOINT=https://s3.<region>.amazonaws.com
+    AWS_S3_BUCKET=your-[dev|prod]-bucket
+    ```
+
+    b. **In local boreholes development**, a local S3-compatible service like [MinIO](https://min.io/) is used, configured as:
```
dcleres, line 394: @stijnvermeeren-swisstopo I added an indent here because it was a bit clearer. I remember that you spotted some issues with thepast indents in the . Could you check if it work on your end? Otherwise I will remove them.

stijnvermeeren-swisstopo (approved):

README.md
```diff
@@ -267,16 +267,18 @@ Activate your virtual environment. On Unix systems, this can be done with the fo
 source env/bin/activate
 ```
 
+<a name="env"></a>
```
dcleres, line 270: I like the idea of adding an anchor here. Shall we do this for other parts of the README in the future as well?

dcleres (commented):
Thank you for adapting the README. Overall, I think it makes sense to remove some information regarding the environment variables as this was taking quite some space in the README. The rationale behind having all this information was that even people who are not super experienced with the concept of environment variables can have a quick start. With some of the info removed, I think it is still feasible but requires a bit of research. That is fine for me. I will keep that in mind for future README edits

README.md
stijnvermeeren-swisstopo: I believe that the only information that was really removed, is the Windows-specific commands for setting environment variables. I don't think that the goal of our README is to teach people how to use different operating systems. At most, I would add a external link to some additional documentation for people who need it.

And yes, I think that there can be other good use cases for internal links in the README was well, feel free to add more.

dcleres: Okay understood!

PR #103, opened 2024-10-31
LGVISIUM-89: Improved clipping behavior documentation

dcleres:
I clarified the clipping behavior via tests and documented them in the ticket JIRA. I also clarified how I generated the test documents. The OpenAPI endpoints were also clarified so that users are aware of the data extraction behavior when using the bounding boxes to extract coordinates, number or text.


src/app/common/schemas.py
```diff
@@ -59,19 +86,52 @@ class FormatTypes(str, Enum):
...
-    """Bounding box schema."""
+    """Bounding box schema for defining a rectangular area within an image.
 
-    x0: float = Field(..., example=0.0)
-    y0: float = Field(..., example=0.0)
-    x1: float = Field(..., example=100.0)
-    y1: float = Field(..., example=100.0)
+    This schema represents the coordinates of the box’s corners, which can be used
+    to specify an area of interest in image processing tasks. Coordinates are
+    defined with the origin at the top-left of the image.
```
stijnvermeeren-swisstopo, line 93: maybe add that the coordinates are defined in pixels?

src/app/common/schemas.py
```diff
@@ -31,18 +41,35 @@ class PNGRequest(BaseModel):
...
+        ...,
+        description="""
+            List of unique identifiers (keys) for the generated PNG files stored in the S3 bucket. Each key allows 
+            access to a specific file within the bucket.
+        """,
+        example=[
+            "dataextraction/file1-1.png",
+            "dataextraction/file1-2.png",
+            "dataextraction/file2-1.png",
+            "dataextraction/file3-1.png",
```
stijnvermeeren-swisstopo, line 70: do we ever have more than one file in a single request/response?

src/app/common/schemas.py
```diff
@@ -125,28 +185,112 @@ def load_from_fitz_rect(rect: fitz.Rect) -> "BoundingBox":
...
 
 class Coordinates(BaseModel):
-    """Coordinates schema."""
+    """Coordinates schema for representing geographical data points.
 
-    east: float = Field(..., example=1.0)
-    north: float = Field(..., example=2.0)
-    projection: str = Field(..., example="LV95")
+    This schema defines the format for specifying location data using east/north coordinates
+    along with the projection system used for geo-referencing.
```
stijnvermeeren-swisstopo, line 191:
```suggestion
    along with the projection system used.
```
Let's remove the "for geo-referencing" as this is something slightly different to my understanding (e.g. aligning a satellite image with a fixed coordinate system, whereas here we are just reading some specific coordinates, not aligning different coordinate systems).

src/app/common/schemas.py
```diff
@@ -125,28 +185,112 @@ def load_from_fitz_rect(rect: fitz.Rect) -> "BoundingBox":
...
 
 class ExtractDataRequest(ABC, BaseModel):
-    """Request schema for the extract_data endpoint.
+    """Request schema for the `extract_data` endpoint.
+
+    **Coordinate Systems:**
+    - **PNG coordinates:** Pixels are measured from the top-left corner (0, 0), where x increases rightward
+    and y downward.
+    - **PDF coordinates:** Also measured from the top-left corner (0, 0), though any transformations between
+    PDF and PNG coordinates are managed internally by the `BoundingBox.rescale` method.
```
stijnvermeeren-swisstopo, line 227: Is it necessary to explain PDF coordinates here? I don't immediately see where the API user needs to know about PDF coordinates, as they are not used in the API anywhere.

src/app/common/schemas.py
```diff
@@ -125,28 +185,112 @@ def load_from_fitz_rect(rect: fitz.Rect) -> "BoundingBox":
...
+
+    ### Fields
+    Each field below includes inline examples to aid users in creating requests. See `json_schema_extra`
+    for a complete example.
+
+    **Attributes:**
+    - **filename** (`Path`): Path to the file (PNG or PDF). _Example_: `"document.png"`
```
stijnvermeeren-swisstopo, line 234: We only support PDF here, don't we?

src/app/common/schemas.py
```diff
@@ -125,28 +185,112 @@ def load_from_fitz_rect(rect: fitz.Rect) -> "BoundingBox":
...
-    page_number: int = Field(..., example=1)  # 1-based index
-    bbox: BoundingBox = Field(..., example={"x0": 0.0, "y0": 0.0, "x1": 100.0, "y1": 100.0})
-    format: FormatTypes = Field(..., example=FormatTypes.COORDINATES.value)
+    filename: Path = Field(
+        ...,
+        description="""
+            Path to the input document file (PNG or PDF) that contains the data to be extracted. This should be
+            a valid file path, and the file should be accessible to the API.
+        """,
+        example=Path("document.png"),
```
stijnvermeeren-swisstopo, line 261: Again, only PDF supported, as far as I understand the code?

Additionally, the PDF file must have been processed by the create_pngs endpoint first, right? We should mention that in the docs as well.

src/app/common/schemas.py
```diff
@@ -183,40 +327,81 @@ class Config:
...
+    This schema includes a `coordinates` field with east/north values and projection information.
+    """
+
+    coordinates: Coordinates = Field(
+        ...,
+        description="""
+            Geographical coordinates extracted from the document, including east, north values, page number,
+            and projection type.
+        """,
+        example={"east": 1.0, "north": 2.0, "page": 1, "projection": "LV95"},
```
stijnvermeeren-swisstopo, line 363: no page here?

src/app/common/schemas.py
```diff
@@ -59,19 +86,52 @@ class FormatTypes(str, Enum):
...
+        ...,
+        description="""
+            The x-coordinate of the top-left corner of the bounding box. This value marks the horizontal starting 
+            point of the box.
+        """,
+        example=0.0,
+    )
+    y0: float = Field(
+        ...,
+        description=""""
```
stijnvermeeren-swisstopo, line 106: 4 instead of 3 double quotes.

stijnvermeeren-swisstopo (changes requested):
image

Can we get Swagger UI to display the all of attribute documentation in regular font (as is the case for y0 thanks to the redundant quotes) instead of this purple monospace font?

src/app/common/schemas.py
```diff
@@ -125,28 +185,112 @@ def load_from_fitz_rect(rect: fitz.Rect) -> "BoundingBox":
...
-    north: float = Field(..., example=2.0)
-    projection: str = Field(..., example="LV95")
+    This schema defines the format for specifying location data using east/north coordinates
+    along with the projection system used for geo-referencing.
+    """
+
+    east: float = Field(
+        ...,
+        description="""
+            Easting coordinate, representing the horizontal position of the point. The value should be in the
```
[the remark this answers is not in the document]
dcleres: Yes, I do agree with you now that I am re-reading what I wrote. It seemed clearer to me when I wrote it.

src/app/common/schemas.py
dcleres: I like that idea.

src/app/common/schemas.py
dcleres: I got carried ways here indeed. Very good comment. Thank you.

src/app/common/schemas.py
dcleres: I see. Sorry, I was not aware of this difference! Good lesson learned for today!

src/app/common/schemas.py
dcleres: I agree with you. That will only confuse people. I also removed the PDF reference a bit later in the text.

src/app/common/schemas.py
dcleres: Fully agree, this was not clear enough. I made a suggestion in the next commit. I added it to the schema here as well as in the `router.py` in the description of the `post_extract_data` endpoint.

stijnvermeeren-swisstopo (approved):
I added three minor fixes. Now everything looks good to me!

PR #104, opened 2024-11-05
LGVISIUM-79: refactor layers object

stijnvermeeren-swisstopo:


src/stratigraphy/layer/layer.py
```diff
@@ -41,118 +39,69 @@ def to_json(self) -> dict:
...
-            all_layers.extend(layers_on_page.layers_on_page)
-        return all_layers
+    depth_intervals: list[Interval]
+    blocks: list[TextBlock]
 
 
 @dataclass
-class IntervalBlockGroup:
-    """A class to represent a group of depth interval blocks."""
+class IntervalBlockPair:
```
lhaibach, line 103: why do we make a distinction between IntervalBlockGroup and InveralBlockPair?

stijnvermeeren-swisstopo: I added a commit with better docstrings for these classes.

I also moved IntervalBlockGroup to the depthcolumn.py file, since it is only used in that file as a helper class.

src/stratigraphy/annotations/draw.py
```diff
@@ -50,7 +48,7 @@ def draw_predictions(
         predictions (dict): Content of the predictions.json file.
         directory (Path): Path to the directory containing the pdf files.
         out_directory (Path): Path to the output directory where the images are saved.
-        document_level_metadata_metrics (pd.DataFrame): Document level metadata metrics.
+        document_level_metadata_metrics (None | pd.DataFrame): Document level metadata metrics.
```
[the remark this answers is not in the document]
stijnvermeeren-swisstopo: Because the method evaluate() in score.py, where the value for `document_level_metadata_metrics` comes from, might return `None` in case there is no ground truth available.

We could consider changing that into returning an empty dataframe instead of None. However, I think that such changes are a bit out of scope for this PR.

PR #105, opened 2024-11-11
LGVISIUM-102: common parent class "Sidebar" for LayerIdentifierColumn and DepthColumn

stijnvermeeren-swisstopo:
A lot of lines changes, but a lot of it is simple renaming, moving code to different files and documentation.

Main renamed classes:
- `BoundaryDepthColumn` -> `AAboveBSidebar`
- `LayerDepthColumn` -> `AToBSidebar`
- `LayerIdentifierColumn` -> `LayerIdentifierSidebar`

These three classes now extend from a common parent class (`Sidebar`).
Logic for finding these different "sidebars" in borehole profiles is contained in respective SidebarExtractor classes.
Code for all of these is now more cleanly contained in a single `sidebar` package.

The bounding boxes for depth columns and material descriptions, that are part of the predictions.json output, are now more elegantly represented by a new class BoundaryBoxes, removing the need to implement JSON serialization logic from the actual Sidebar (formerly DepthColumn) classes.

There is a very small / almost negligible effect on the benchmarking scores, caused by the fact that we no longer automatically reject material descriptions to which we cannot associate any depth interval.

This PR also updates the documentation for the predictions.json format, moving it to a separate markdown file, and removing a few redundant entries from the JSON output


lhaibach (approved):
I like the structure and naming -- it's much more intuitive for me now to understand which class does what.
Question regarding parent classes: `AAboveBSidebar` and `AToBSidebar`  and `LayerIdentifierSidebar` are child classes of Sidebar, which makes perfect sense to me. For the corresponding extractors `AAboveBSidebarExtractor` and `AtoBSidebarExtractor` and `LayerIdentifierSidebarExtractor` this not the case. How do you decide when to use parent classes for certain classes?
Regarding `the #todo: automatic AAboveBInterval`: will it be possible somehow to determine the interval type based on the sidebar type?

stijnvermeeren-swisstopo:
@lhaibach 
Re `SidebarExtractor`: potentially it would make sense to have these extend from a common parent class. This might also be what the comment `TODO: Similar to AToBSidebarExtractor.find_in_words(). Refactoring may be desired.` in `layer_identifier_sidebar_extractor.py` is hinting at. However, I don't immediately see a significant amount of shared logic that we could move into a shared parent class. But feel free to give it a try, if you think there is potential to simplify the code like that.

Generally speaking, main reasons for using a parent class are:
- Shared code between different classes
- Allowing different classes to be passed as a value for the same parameter (with transparent type annotations)

Re the TODO in `extract.py`: I don't think the problem is determining the interval type (at this point in the code, we already have an interval instance with a specific type). I think the solution should be to allow any interval type (not just AAboveBInterval) in the Layer class, and to deal with any implications that this might have (in particular, concerning JSON serialization and deserialization).

stijnvermeeren-swisstopo:
I have now created a ticket for that last TODO: https://jira.swisstopo.ch/browse/LGVISIUM-104

lhaibach (approved):

lhaibach:
@stijnvermeeren-swisstopo thanks for explaining :) !

PR #107, opened 2024-12-04
Lgvisium 100/extract depth value error handling

lhaibach:
Changes done:

- In the extract_depth function, the ValueError handling has been incorporated, and the matched pattern are now more specific to groundwater depth. This change reduced the number of ValueErrors encountered during float conversion, which previously caused the process to exit the loop prematurely. However, this refinement has led to an increase in false positives.

- To reduce these false positives, the logic for finding groundwater information has been changed. Groundwater search is now directly tied to the material description bounding box, meaning groundwater data is only search for in areas associated with material descriptions. This adjustment reduces unnecessary searches on pages without material descriptions and ensures groundwater extraction is more contextually related to material description areas. Since groundwater details can sometimes appear at the end of a page or farther from the material description, the search around the bounding box is not overly strict.

- To further reduce false postitives the logic for finding lines near keys was changed (impacting coordinate, elevation, and depth extraction). Instead of searching in a rectangle ( including entries diagonal to the key), the updated code now searches along the x and y axes. This adjustment improves the F1 score for groundwater extraction while slightly worsening performance for elevation extraction (data/geoquat/validation/ A11486.pdf).

More in Ticket LGVISIUM-100

### DIFF
--- a/src/stratigraphy/data_extractor/data_extractor.py
+++ b/src/stratigraphy/data_extractor/data_extractor.py
@@ -10,7 +10,6 @@ from typing import Generic, Self, TypeVar
 
 import fitz
 import regex
-from stratigraphy.data_extractor.utility import get_lines_near_rect
 from stratigraphy.lines.line import TextLine
 from stratigraphy.util.util import read_params
 
@@ -193,7 +192,7 @@ class DataExtractor(ABC):
             list[TextLine]: The lines close to the key.
         """
         key_rect = key_line.rect
-        feature_lines = self.get_lines_near_rect(lines, key_rect)
+        feature_lines = self.get_axis_aligned_lines(lines, key_rect)
 
         # Insert key_line first and remove duplicates
         feature_lines.insert(0, key_line)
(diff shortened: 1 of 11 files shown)


src/stratigraphy/data_extractor/data_extractor.py
```diff
@@ -204,21 +203,38 @@ def get_lines_near_key(self, lines, key_line: TextLine) -> list[TextLine]:
 
         return feature_lines_sorted
 
-    def get_lines_near_rect(self, lines, rect: fitz.Rect) -> list[TextLine]:
-        """Find the lines of the text that are close to a given rectangle.
+    def get_axis_aligned_lines(self, lines: list[TextLine], rect: fitz.Rect) -> list[TextLine]:
```
stijnvermeeren-swisstopo, line 206: I would suggest writing a few unit tests for this method. As this method is part of the abstract class DataExtractor, we will have to pick a particular implementation for the unit tests. For coordinate extraction, we already have a few similar methods that are tested in `tests/test_coordinate_extraction.py`, so you can look there for inspiration. The actual tests for `get_axis_aligned_lines` could be added there for `CoordinateExtractor`, but you could also write them for `ElevationExtractor` or `GroundwaterExtractor` (one is enough for now, I think).

stijnvermeeren-swisstopo (changes requested):

stijnvermeeren-swisstopo (approved):
Looks good to me

PR #109, opened 2025-01-20
LGVISIUM-96: more robust depth values

stijnvermeeren-swisstopo:
- More robust detection / exclusion of arithmetic progressions (or very similar sequences) as depth values.
- When computing the score of a "(material description rect + sidebar) match", use the noise count relative to the total number of depth entries, because longer sidebars are likely to have more noise anyway.

### DIFF
--- a/src/stratigraphy/sidebar/a_above_b_sidebar.py
+++ b/src/stratigraphy/sidebar/a_above_b_sidebar.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-import math
+import statistics
 from dataclasses import dataclass
 
 import fitz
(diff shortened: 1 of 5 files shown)


src/stratigraphy/sidebar/a_above_b_sidebar_validator.py
```diff
@@ -59,6 +59,8 @@ def is_valid(self, sidebar: AAboveBSidebar, corr_coef_threshold: float = 0.99) -
         # Check if the entries are strictly increasing.
         if not sidebar.is_strictly_increasing():
             return False
+        if sidebar.close_to_arithmetic_progression():
```
lhaibach, line 62: dont we check for close arithmetic twice for the same sidebar ( once in filtered_columns and once in the validator?

src/stratigraphy/sidebar/a_above_b_sidebar.py
```diff
@@ -64,32 +65,29 @@ def depth_intervals(self) -> list[AAboveBInterval]:
...
-        progression = np.array(range(len(self.entries)))
-        entries = np.array([entry.value for entry in self.entries])
-
-        # Avoid warnings in the np.corrcoef call, as the correlation coef is undefined if the standard deviation is 0.
-        if np.std(entries) == 0:
+        values = [entry.value for entry in self.entries]
+
+        first = values[0]
+        last = values[-1]
+        if last > first:
```
lhaibach, line 77:
```suggestion
        differences = [values[i + 1] - values[i] for i in range(len(values) - 1)]
        step = round(statistics.median(differences), 2)
        if step <= 0:
            return False

        arithmethic_progression = {
            # ensure we have nicely rounded numbers, without inaccuracies from floating point arithmetic
            round(value * step, 2)
            for value in range(int(first / step), int(last / step) + 1)
        }
```
improves F1 of Geoquat to 0.59796 and Zurich to 0.8911

lhaibach (changes requested):
Adjusting the close_to_arithmetic_progression improves F1 for both Zurich and Geoquat. The fact that we check if a sidebar is close to a arithmetic progression, I would check out after merging the branch into LGD-564? As I have no clear idea whether it is really necessary to check twice for the same sidebar if it is a progression and if not where its better to check for ( in the validator or in extractor).

src/stratigraphy/sidebar/a_above_b_sidebar.py
stijnvermeeren-swisstopo: Geoquat indeed improved significantly. This is due to the fact that we now also consider arithmetic progressions with a step that is not a power of 10 (0.01, 0.1, 1, 10, etc.). Indeed, there are at least 4 profiles in the Geoquat validation dataset with fixed depth values at an interval of 1.5m: A1287.pdf, A1289.pdf, A1290.pdf and B627.pdf.

The Zurich F1 score was already at 89.11%, so there is no change there.

Still, the improvement for Geoquat is quite significant, thanks a lot for spotting this! I have included your proposed change in the PR now.

src/stratigraphy/sidebar/a_above_b_sidebar_validator.py
stijnvermeeren-swisstopo: I believe that removing the call from the extractor would likely give the same accuracy, but the extraction might become slower. By already checking for close_to_arithmetic_progression() when obtaining the filtered_columns, we can already exclude certain potential sidebars at this point, so that we have a smaller number of sidebars to pass to the (computationally expensive) reduce_until_valid method. So I would probably leave both calls for the time being.

stijnvermeeren-swisstopo:
@lhaibach Thanks for the review and for finding the improvement for the Geoquat dataset by using the median step! I've included this change in the PR now.

What do you mean exactly by "I would check out after merging the branch into LGD-564?" ?

lhaibach:
@stijnvermeeren-swisstopo Great :) Regarding "I would check out after merging the branch into LGD-564?" , I referred to the double checking for arithmetic progressions in the extractor and the validator, but as you said, keeping both seems reasonable.

lhaibach (approved):

PR #110, opened 2025-01-22
Tag git branch and last commit in MLFlow

stijnvermeeren-swisstopo:
This change allows us to track the current git branch as well as the last commit (message + id) in MLFlow.

This could be helpful in tracking which experiment run corresponds to which state of the code, reducing the need to recall this from memory or to write manually a description for each run.

image

Also sets the default experiment name to "Boreholes data extraction" (more accurate than the previous "Boreholes Stratigraphy")

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,8 @@ lint = [
     "pre-commit==3.6.2",
 ]
 experiment-tracking = [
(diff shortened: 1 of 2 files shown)


lhaibach (approved):
this is indeed pretty helpful 👍

PR #111, opened 2025-01-24
Lgd 564/optimize a above b sidebar extractor

lhaibach:
Subsequent to [#108 ](https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/108), which optimized the logic to find AAboveBSidebars using clusters, the changes in this branch serve to reduce runtime in the AAboveBSidebar extractor.
Main changes
- noise count was previously calculated multiple times for the same sidebar and then again for the material description rect + sidebar pair. Thus a new class SidebarNoise (wrapping the sidebar and its noise count) is used and MaterialDescRectWithSidebar has noise_count of its sidebar as a property.
- the noise_count is calculated for the words intersecting with the sidebar.rect and not all words from the whole page are used.
- all words on a page are indexed by their position on a page for faster lookup using an rtree.
- reduce runtime of smaller functions ( is_strictly _increasing() + pearson_correlation_coef())

We might want to overthink the logic in reduce_until_valid() form the AAboveBSidebarValidator used in the Sidebarextraction, as we potentially call some functions for long sidebars very often if they are not valid. Another idea would be to return after a certain number of iterations of reducing the sidebar None. However, this I would look at in another ticket. 

Most of the runtime of this code is now due to other parts in the code, further described in [LDG-564](https://jira.swisstopo.ch/browse/LGD-564). F1 for zurich: from 0.8911 to  0.8891 and for geoquat: 0.5979 to 0.5972


src/stratigraphy/sidebar/a_above_b_sidebar.py
```diff
@@ -95,10 +99,13 @@ def pearson_correlation_coef(self) -> float:
         entries = np.array([entry.value for entry in self.entries])
 
         # Avoid warnings in the np.corrcoef call, as the correlation coef is undefined if the standard deviation is 0.
```
stijnvermeeren-swisstopo, line 101: Comment is outdated as there is no longer a np.corrcoef call

src/stratigraphy/sidebar/a_above_b_sidebar.py
```diff
@@ -95,10 +99,13 @@ def pearson_correlation_coef(self) -> float:
...
-        if np.std(entries) == 0 or np.std(positions) == 0:
+        std_positions = np.std(positions)
+        std_entries = np.std(entries)
+        if std_positions == 0 or std_entries == 0:
             return 0
 
-        return np.corrcoef(positions, entries)[0, 1].item()
+        covariance = np.mean((positions - np.mean(positions)) * (entries - np.mean(entries)))
+        return covariance / (std_positions * std_entries)
```
stijnvermeeren-swisstopo, line 108: Add a comment here, explaining that we are computing the correlation coefficient here, but that we avoid calling np.corrcoef to avoid computing the standard deviations twice.

src/stratigraphy/sidebar/a_above_b_sidebar.py
```diff
@@ -45,7 +45,11 @@ def strictly_contains(self, other: AAboveBSidebar) -> bool:
         )
 
     def is_strictly_increasing(self) -> bool:
-        return all(i.value < j.value for i, j in zip(self.entries, self.entries[1:], strict=False))
+        length = len(self.entries)
+        for i in range(length - 1):
+            if self.entries[i].value >= self.entries[i + 1].value:
+                return False
+            return True
```
stijnvermeeren-swisstopo, line 52: The return statement is on the wrong indentation level. Like this, after checking the first two values, we immediately return, and we never check any later values.

Actually, we can still use all(), even when we want to avoid the zip call:
```suggestion
        return all(
          self.entries[i].value < self.entries[i + 1].value
          for i in range(len(self.entries) - 1)
        )
```
I think that's more readable and less prone to mistakes.

src/stratigraphy/extract.py
```diff
@@ -75,37 +77,48 @@ def process_page(
...
 
     # If there is a layer identifier sidebar, then we use this directly.
     # Else, we search for sidebars with depths.
     # We could also think of some scoring mechanism to decide which one to use.
     if not material_descriptions_sidebar_pairs:
         words = sorted([word for line in lines for word in line.words], key=lambda word: word.rect.y0)
-        a_to_b_sidebars = AToBSidebarExtractor.find_in_words(words)
+        word_rtree = rtree.index.Index()
+        for i, word in enumerate(words):
+            word_rtree.insert(i, (word.rect.x0, word.rect.y0, word.rect.x1, word.rect.y1))
```
stijnvermeeren-swisstopo, line 89: According to the documentation (https://rtree.readthedocs.io/en/latest/class.html), you can use the word itself as the third parameter of the insert method. Then the TextWord object is stored in the rtree, not just the id. You can then retrieve it directly in the intersection method like this:
```
def _get_intersecting_words(word_rtree: rtree.index.Index, rect: fitz.Rect) -> list[TextWord]:
    """Retrieve all words from the page intersecting with Sidebar bounding box."""
    intersecting_words: list[TextWord] = list(
        word_rtree.intersection((rect.x0, rect.y0, rect.x1, rect.y1), objects="raw")
    )
    return [word for word in intersecting_words if any(char.isalnum() for char in word.text)]
```
Then this method (and by extension the noise_count method) does not need to `all_words` parameter anymore, so this would simplify our method signatures quite a bit.

stijnvermeeren-swisstopo (changes requested):

PR #113, opened 2025-01-29
Lgd 564/improve ocr mistake handling

lhaibach:
Follow-up to [Pull#111](https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/111):
   - Moved OCR corrections to the make_ascending() function, eliminating the need for OCR corrections in reduce_until_valid().
   - Added correction for entries by a factor of 10 where possible within make_ascending().
   - Optimized is_strictly_increasing() and pearson_correlation_coef() to improve computation time.

These changes reduced processing time for Bearsol.pdf from 12 minutes to 5.4 minutes.

The F1 score for Zurich is now 0.8902 (previously 0.8911), with the slight drop due to a single bore profile. In one entry, the digit 1 was misrecognized as 4, and the decimal point was not detected by the OCR. The current correction method only adjusts for either the decimal point or the misrecognized digit.

The F1 score for Geoquat remains at 0.5979.

### DIFF
--- a/src/stratigraphy/sidebar/a_above_b_sidebar.py
+++ b/src/stratigraphy/sidebar/a_above_b_sidebar.py
@@ -45,7 +45,7 @@ class AAboveBSidebar(Sidebar[DepthColumnEntry]):
         )
 
     def is_strictly_increasing(self) -> bool:
-        return all(self.entries[i].value < self.entries[i + 1].value for i in range(len(self.entries) - 1))
+        return all(i.value < j.value for i, j in zip(self.entries, self.entries[1:], strict=False))
 
     def depth_intervals(self) -> list[AAboveBInterval]:
         """Creates a list of depth intervals from the depth column entries.
@@ -94,15 +94,11 @@ class AAboveBSidebar(Sidebar[DepthColumnEntry]):
         positions = np.array([entry.rect.y1 for entry in self.entries])
         entries = np.array([entry.value for entry in self.entries])
 
-        std_positions = np.std(positions)
-        std_entries = np.std(entries)
-        if std_positions == 0 or std_entries == 0:
+        # Avoid warnings in the np.corrcoef call, as the correlation coef is undefined if the standard deviation is 0.
+        if np.std(entries) == 0 or np.std(positions) == 0:
             return 0
 
-        # We calculate the Pearson correlation coefficient manually
-        # to avoid redundant standard deviation calculations that would occur with np.corrcoef.
-        covariance = np.mean((positions - np.mean(positions)) * (entries - np.mean(entries)))
-        return covariance / (std_positions * std_entries)
+        return np.corrcoef(positions, entries)[0, 1].item()
 
     def remove_entry_by_correlation_gradient(self) -> AAboveBSidebar | None:
         if len(self.entries) < 3:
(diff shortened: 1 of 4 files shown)


src/stratigraphy/sidebar/a_above_b_sidebar.py
```diff
@@ -117,18 +114,35 @@ def remove_entry_by_correlation_gradient(self) -> AAboveBSidebar | None:
...
+                        self.entries[i] = DepthColumnEntry(rect=entry.rect, value=alternative_value)
+                        break
 
         return self
 
+    def _valid_value(self, index: int, new_value: float) -> bool:
+        """Check if new value at given index is maintaining ascending order."""
+        previous_ok = index == 0 or all(
+            other_entry.value < new_value for other_entry in self.entries[:index]
+        )  ## too strict?
```
stijnvermeeren-swisstopo, line 142: Why would it be too strict? Please give more details, or remove the comment.

src/stratigraphy/sidebar/a_above_b_sidebar.py
```diff
@@ -94,15 +95,11 @@ def pearson_correlation_coef(self) -> float:
...
-        if std_positions == 0 or std_entries == 0:
+        # Avoid warnings in the np.corrcoef call, as the correlation coef is undefined if the standard deviation is 0.
+        if np.std(entries) == 0 or np.std(positions) == 0:
             return 0
 
-        # We calculate the Pearson correlation coefficient manually
-        # to avoid redundant standard deviation calculations that would occur with np.corrcoef.
-        covariance = np.mean((positions - np.mean(positions)) * (entries - np.mean(entries)))
-        return covariance / (std_positions * std_entries)
+        return np.corrcoef(positions, entries)[0, 1].item()
```
stijnvermeeren-swisstopo, line 102: Is it intentional that this was reverted to the original implementation? Did the change not make a measurable difference?

src/stratigraphy/sidebar/a_above_b_sidebar.py
```diff
@@ -117,18 +114,35 @@ def remove_entry_by_correlation_gradient(self) -> AAboveBSidebar | None:
...
-                if previous_ok and next_ok:
-                    # Create a new entry instead of modifying the value of the current one, as this entry might be
-                    # used in different sidebars as well.
-                    self.entries[i] = DepthColumnEntry(rect=entry.rect, value=factor100_value)
+                new_values.extend([entry.value / 100, entry.value / 10])
+                for new_value in new_values:
+                    if self._valid_value(i, new_value):
+                        # Create a new entry instead of modifying the value of the current one, as this entry might be
+                        # used in different sidebars as well.
+                        self.entries[i] = DepthColumnEntry(rect=entry.rect, value=new_value)
```
stijnvermeeren-swisstopo, line 124: We could set `entry`  to the new value as well, then we have the possibility to apply a factor 10/100 correction as well as an OCR correction to the same entry.

It does not make a difference for 691251009-bp.pdf though, as the conditions for making the necessary corrections are not given here anyway.

stijnvermeeren-swisstopo (changes requested):
I mean to do "request changes", not yet "approve".

src/stratigraphy/sidebar/a_above_b_sidebar.py
```diff
@@ -45,7 +46,7 @@ def strictly_contains(self, other: AAboveBSidebar) -> bool:
         )
 
     def is_strictly_increasing(self) -> bool:
-        return all(self.entries[i].value < self.entries[i + 1].value for i in range(len(self.entries) - 1))
+        return all(i.value < j.value for i, j in zip(self.entries, self.entries[1:], strict=False))
```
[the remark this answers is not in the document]
lhaibach: "You're absolutely right! I accidentally kept the changes from main when merging it into this branch. The issue is now corrected, thanks for catching that.

src/stratigraphy/sidebar/a_above_b_sidebar.py
lhaibach: same as above

src/stratigraphy/sidebar/a_above_b_sidebar.py
lhaibach: I updated make_ascending, but I’m not entirely sure if it aligns with your suggestion. When I included entry.value in new_values (new_values = [entry.value]), F1 dropped to 0.8897 for Zurich and 0.5874 for Geoquat. Because of this, I left new_values empty in my latest commit. However, including it would significantly reduce runtime (e.g., half the time for Bearsol). Let me know what you think about the current version :)

src/stratigraphy/sidebar/a_above_b_sidebar.py
lhaibach: If an error isn't corrected (detected) in an earlier stage, any corrections made later won’t pass _valid_value because previous_ok will be False. That’s why I thought the condition might be too strict or worth looking into.
f.e if the very first entry (100 instead of 1.00) the following entry 50 in [2,3,4,50,6,...]  wont be corrected.

stijnvermeeren-swisstopo (approved):
In make_ascending, the code now explicitly only applies one change (either factor 10/100 or OCR mistake correction), never both for a single entry.

PR #114, opened 2025-02-04
Lgvisium 106/reduce a above b sidebar deepwells

lhaibach:
**Problem**
As described in [LGVISIUM-106](https://jira.swisstopo.ch/browse/LGVISIUM-106) the correct depth descriptions for deep wells often overlap with depth scales. The goal of this pull request was to filter out these arithmetic progressions while retaining only the correct depth descriptions. Typically, correct depth entries are reported as floats, whereas arithmetic progressions are often integers.

**Changes**
- Filtering arithmetic progressions:
    - Arithmetic progressions of integers in AAboveBSidebar entries are now removed in find_in_words.
    - The noise_count for the sidebar remains unchanged.
    - is_close_to_arithmetic_progression now requires more than two entries to be considered an arithmetic sequence (previously >= 2).
 - Data type adjustments:
    - DepthColumnEntry can now be either int or float (previously only float).
    - If we would want them to be of strictly one type, I would need to adjust my current approach.
 
**Outcome**
- The runtime does not increase.
- Deep Wells
   - For Baersol and Jungfraukeil, the correct sidebar is selected, and integer depth scales are successfully removed.
   - Open issue: In profiles spanning multiple pages, some material descriptions are not recognized due to sidebar entries not covering the full page. I would address this in the ticket [LGVISIUM-44](https://jira.swisstopo.ch/browse/LGVISIUM-44)
- Zurich & GeoQuat
    - Zurich: F1 score slightly decreases from 0.8903 to 0.8886 due to changes in 268124125-bp.pdf and 678249001-bp.pdf (the former has arithmetic progression as depth entries including hand written depth entries as floats, in the latter the OCR did not detect the numbers after the decimal leading to arithmetically progressing integers.
    - GeoQuat: F1 score remains unchanged at 0.5979.


stijnvermeeren-swisstopo (approved):
LGTM!

Issue #120, opened 2025-02-18
Improve import statements and repository structure


stijnvermeeren-swisstopo:
[Migrated from LGVISIUM-108.]

## 1. Import Statement Cleanup

We recently encountered and fixed a circular import issue caused by interdependencies between depth and sidebar modules. While the issue is resolved, the current import structure should be reviewed and potential inconsistencies remove as they could lead to future problems.

## 2. Repository Restructuring

All scripts are currently under stratigraphy/, but the project scope has expanded to extracting other information as well (e.g metadata extraction, f.e coordinate). Moving non-stratigraphic components to a more appropriate module would lead to a more accurate representation of the functionality.

Other issues / things that are counterintuitive:
- Directory `lines` includes code for both geometric lines as well as text lines 
- Classes related to `depth` are spread across different directories
- ...

Plan:
- During the sprint, make a precise plan of which files/directories we want to rename or move
- In between sprints, when all other PRs are closed, execute the refactoring (avoiding conflicts)

## 3. Change all `fitz` imports to `pymupdf`

Possible since https://github.com/pymupdf/PyMuPDF/discussions/3458.

PR #128, opened 2025-02-18
implement LayerDepths class for better typed depth interval handling

### LINKED ISSUE #122: Layer class should allow any interval type
[Migrated from LGVISIUM-104.]

Currently, the Layer class only allows an AAboveBInterval for the depth_interval field. This is not correct, as a layer may have any interval type.

The current situation forces other intervals to be coerced into AAboveBInterval in extract.py, where the code already contains a TODO _"don't automatically convert any interval to an AAboveBInterval"_.

We should allow any interval type (not just AAboveBInterval) in the Layer class, and deal with any implications that this might have (in particular, concerning JSON serialization and deserialization).

stijnvermeeren-swisstopo:
closes #122 

There is a marginal drop in F1 scores, but this is remediated with the follow-up PR https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/129

Also improve the visualisation of depth values, in particular where the depths are extraction from the material description.
For example GeoQuat 10781.pdf...
Before:
image

After:
image


src/stratigraphy/depth/a_to_b_interval_extractor.py
```diff
@@ -59,30 +57,39 @@ def from_material_description_lines(cls, lines: list[TextLine]) -> AToBInterval
...
+        for index, word in enumerate(text_line.words):
+            char_index_to_word_index.extend([index] * (len(word.text) + 1))
 
         query = r"-?([0-9]+(\.[0-9]+)?)[müMN\]*[\s-]+([0-9]+(\.[0-9]+)?)[müMN\\.]*"
         if not require_start_of_string:
             query = r".*?" + query
         regex = re.compile(query)
         match = regex.match(input_string)
+
+        def rect_from_group_index(index):
```
lhaibach, line 82: maybe include some documentation as what exactly is done and it would probably be helpful to have a unit test for this function. Also we might want to change fitz to pymupdf as in the rest of the repository:)

stijnvermeeren-swisstopo: I've added comments and unit tests.

Regarding the import statements for `fitz`/`pymupdf`: this should ideally be done for all files in the project at once, and I think that https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/120 is a better place to do that, so I added it as a requirement to that issue.

tests/test_interval.py
```diff
@@ -1,47 +1,47 @@
...
+    ]
+    interval = AToBIntervalExtractor.from_text(TextLine(words), require_start_of_string=True)
+    assert interval is None, "With require_start_of_string=True, matches that are not at the start are not allowed."
+
+    words = [
+        TextWord(pymupdf.Rect(0, 0, 4, 1), "1,70-5,80m:", page=1),
+        TextWord(pymupdf.Rect(4, 0, 6, 1), "COLLUVIONS", page=1),
+    ]
+    interval = AToBIntervalExtractor.from_text(TextLine(words), require_start_of_string=False)
+    assert interval is not None, "With require_start_of_string=True, matches that are at the start are allowed."
```
lhaibach, line 47:
```suggestion
    assert interval is not None, "With require_start_of_string=False, matches that are at the start are allowed."
```

Issue #119, opened 2025-02-18
Improve entry extraction for LayerIdentifierSidebar


stijnvermeeren-swisstopo:
[Migrated from LGVISIUM-109.]

The class method find_layer_identifier_sidebar_entries from LayerIdentifierSidebarExtractor identifies `m)` as a correct entry in lines like:
```brecciösen Lagen (348.3 -348.9, 249.2 - 350.0m)```
in the deepwell Arsch.

Because a LayeridentifierSidebar is found (with two entries, both `m)`), the search for the correct AaboveBSidebar is not triggered.

![Image](https://github.com/user-attachments/assets/3a8fa691-2658-4e03-b541-5bda84ab569a)

Suggestions:
- make the matching for LayerIdentifierSiebars more robust
- dont allow the same value for entries in one LayerIdentifierSidebar
- maybe allow the search for AAboveB and AToBsidebars even if a LayerIdentifiersidebar was found and find logic to decide which to choose?

PR #129, opened 2025-02-19
Improve logic for scoring and selecting sidebar-material-description pairs

stijnvermeeren-swisstopo:
Compared to the baseline before https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/128...
- F1 score for Zurich improves from 89.0% to 89.9%
- F1 score for GeoQuat improves from 59.8% to 60.4%

There are a few individual inputs that are negatively affected, mostly where the sidebar and the material descriptions are quite far away from each other / have a significantly different height (A11120.pdf, A11294.pdf). I think we can accept these as edge cases for now.

Main changes:
- Instead of sequentially searching for 1) indicator sidebars, 2) a-to-b and a-above-b sidebars, and 3) material description rects without sidebar, and stopping as soon as we have found something, we now search for everything before selecting the best one(s)
- Some optimization of the sidebar-material-description scoring
- Sidebar-material-description pairs with a score that is too low are rejected
- Some refactoring, putting some previously stand-alone methods in `extract.py` into a class `MaterialDescriptionRectWithSidebarExtractor`.
- Using the existing clustering logic also in the AToBSidebarExtractor
- More robust noise count computation:
  - based on text lines instead of words
  - lines that intersect any entries of the sidebar are ignored in the computation


lhaibach (approved):
The F1 score improvements for both Zurich and GeoQuat look solid! I also ran the code for Baersol, Jungfraukeil, and Arsch. There's an improvement in the sidebar selection for Arsch compared to the previous implementation [Issue 119](https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/119)

However, we still allow overlapping of multiple MaterialDescriptionRectWithSidebar instances [Issue 130](https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/130). This is something that should be addressed (in another ticket) to ensure no unintended overlaps between material descriptions and sidebars.

I really like the new approach of first retrieving all sidebar types and deciding which to use based on the scoring! I think that this makes the selection process much more systematic and less arbitrary :)

PR #143, opened 2025-03-03
Improve setup documentation

TicaGit:
Description:

While setting up the project, I ran into a few small issues that might affect other as well. I’ve updated the README.md with additional setup instructions. Additionally, I made minor modifications to config.py because it wasn't correctly sourcing the environment variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,31 @@ source env/bin/activate
 
 2. **Download the borehole profiles, optional**
 
-Use `boreholes-download-profiles` to download the files to be processed from an AWS S3 storage. In order to do so, you need to authenticate with aws first. We recommend to use the aws CLI for that purpose. This step is optional, you can continue with step 3 on your own set of borehole profiles.
+Use `boreholes-download-profiles` to download the files to be processed from an AWS S3 storage. In order to do so, you need to authenticate with aws first. We recommend using the aws CLI for that purpose, or storing your credentials in the ~/.aws configuration files This step is optional, you can continue with step 3 on your own set of borehole profiles.
+
+Alternativelly, you can download the data directly using the AWS CLI:
+```bash 
+brew install awscli
+aws s3 sync s3://stijnvermeeren-boreholes-data ./data
+```
+
+If you choose to use the ~/.aws files, here how they should look like.
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -167,6 +169,35 @@ The project structure and the most important files are as follows:
 
 - `main.py` : This is the main script of the project. It runs the data extraction pipeline, which analyzes the PDF files in the `data/Benchmark` directory and saves the results in the `predictions.json` file.
 
+## Data
+
+To locally download the data from the AWS S3 bucket to the `./data` directory, run the following commands:
```
stijnvermeeren-swisstopo, line 174: Isn't this already covered by the `boreholes-download-profiles` script that is documented higher up in this README?

stijnvermeeren-swisstopo (changes requested):

README.md
TicaGit: Yes, it missed it :) I moved my explanations on the right section. I can just remove them if you prefer, but it might be usefull to still have the method with the cli that I used.

README.md
```diff
@@ -119,7 +121,31 @@ source env/bin/activate
...
-Use `boreholes-download-profiles` to download the files to be processed from an AWS S3 storage. In order to do so, you need to authenticate with aws first. We recommend to use the aws CLI for that purpose. This step is optional, you can continue with step 3 on your own set of borehole profiles.
+Use `boreholes-download-profiles` to download the files to be processed from an AWS S3 storage. In order to do so, you need to authenticate with aws first. We recommend using the aws CLI for that purpose, or storing your credentials in the ~/.aws configuration files This step is optional, you can continue with step 3 on your own set of borehole profiles.
+
+Alternativelly, you can download the data directly using the AWS CLI:
+```bash 
+brew install awscli
+aws s3 sync s3://stijnvermeeren-boreholes-data ./data
+```
+
+If you choose to use the ~/.aws files, here how they should look like.
```
stijnvermeeren-swisstopo, line 132:
```suggestion
If you choose to use the ~/.aws files, then they should look like this:
```

stijnvermeeren-swisstopo (approved):

Issue #147, opened 2025-03-06
Split all the keywords in the params file by language


letao:
Keywords for material description in 'config/matching_params.yml' are already split by language, wheres keywords for coordinates, groundwater and elevation are not. We should split the keywords by language for all properties (Groundwater and Coordinates + Elevation). 

Currently these descriptions are fetched as part of the '/src/stratigraphy/data_extractor/data_extractor.py' script. One option would be to amend the current approach to allow the usage of extracted language data (e.g. from metadata) to filter for the necessary descriptions.

https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/331275c633397c835a28a0d97fadaef5cee0e616/src/stratigraphy/data_extractor/data_extractor.py#L122

PR #149, opened 2025-03-12
Added english language expressions to material desciptions in the mat…

letao:
Added English language support to material descriptions in the matching_params.yml. 

Created a list of including and excluding expressions for the English language in material description sections. Current List is a composed of translated German and French expressions combined with expressions found specifically in Nagra dataset. 

For additional language support (Coordinates, Elevation & Groundwater) a new issue has been created: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/147

### DIFF
--- a/config/matching_params.yml
+++ b/config/matching_params.yml
@@ -105,6 +105,50 @@ material_description:
       - profondeur
       - désignation
 
+  en:
+    including_expressions:
+      - sand
+      - sandy
+      - sandstone
+      - silt
+      - silty
+      - siltstone
+      - clay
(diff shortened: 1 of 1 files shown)


stijnvermeeren-swisstopo (approved):
LGTM.
I tested it on the Nagra files (where all material descriptions are recognized now) as well as on Zurich and Geoquat validation dataset. In the Geoquat dataset, there are actually two file that are now correctly recognized as English (A1025.pdf and A1026.pdf) and for which the accuracy is slightly improved. Great!

Issue #134, opened 2025-02-20
Restructure ground truth and predictions to allow for multiple boreholes per input PDF


stijnvermeeren-swisstopo:
[Migrated from LGVSIUM-98.]

How many input PDFs do we have where this makes a difference at all (i.e. more than one borehole in a single PDF)?

Known examples in Zurich dataset: 267125358-bp.pdf, 268124232-bp.pdf, 268125295-bp.pdf

How about metadata? Each borehole in a single PDF might have their own metadata. So ideally, we need to allow for this in the output format as well. But we also need to consider the case where the PDF might list certain metadata only once, and this value applies to all boreholes in the PDF (e.g. same date and/or coordinates for all boreholes).

Current behaviour:

- Different stratigraphies (depth column + material descripions) are correctly recognized, but their data is simple concatenated into a single list of layers when generating the output.
- Groundwater data and metadata is directly associated with the file in the output, not with a particular stratigraphy.

stijnvermeeren-swisstopo:
A full list of relevant PDFs in the Zurich dataset according to @TicaGit :
- 695265009-bp.pdf
- 267125358-bp.pdf
- 269126143-bp.pdf
- 675245009-bp.pdf
- 681249142-bp.pdf
- 268124232-bp.pdf
- 268123018-bp.pdf
- 268124112-bp.pdf
- 268125295-bp.pdf

PR #150, opened 2025-03-12
Fix issue #134: Support multiple boreholes per PDF

TicaGit:
Title: Fix issue #134: Support multiple boreholes per PDF

Description:
This PR adds support for extracting multiple boreholes from a single PDF document. Previously, the system treated all identified elements as part of a single borehole, leading to incorrect data extraction.

Changes:
- Many files have been modified to handle lists of boreholes per document.
- Some new classes were added, and others were renamed to reflect these structural changes.
- All features can now be detected multiple times in a single document, including:
   - Multiple elevations
   - Multiple sets of coordinates
   - Multiple borehole layers
   - Multiple bounding boxes
   - Multiple groundwater extraction
- Metadata restructuring:
   - Introduced file_metadata: Stores document-level information (e.g., language).
   - metadata now contains borehole-specific data (e.g., elevation).
- Updated ground truth format: new structure: { "name.pdf": [list of boreholes] }

Notes:
The ground truth JSON files must be downloaded from the S3 bucket, as some manual modifications were made to correctly separate documents containing multiple borehole profiles.


src/stratigraphy/evaluation/groundwater_evaluator.py
```diff
@@ -70,55 +80,79 @@ def evaluate(self) -> OverallGroundwaterMetrics:
...
+            groundwater_elevation_metrics_list = []
+            groundwater_date_metrics_list = []
+
+            # iterate on all the borehole detected in the document
+            for pred_index, groundwaters_in_borehole in enumerate(groundwater_in_doc):
+                # from the matching previously done on the layer description, extract the coresponding gt borehole
+                ground_truth_index = pred_to_gt_matching.get(pred_index)
+                if ground_truth_index is None:
+                    # when the extraction detects more borehole than there actually is in the ground truth, the worst
+                    # predictions have no match and must be skipped for the evaluation
```
stijnvermeeren-swisstopo, line 99: Wouldn't it make morse sense to count such predictions as false positives? Otherwise we are not penalized at all for generating too much output.

stijnvermeeren-swisstopo (changes requested):
Looks pretty good already! I initially did not ecpect that this issue would require so many code changes, thanks for all the hard work!

There are some smaller changes to the accuracy metrics for Zürich and Geoquat, even for some documents that only have one borehole profile. I don't fully understand why yet, I will investigate a little.

I will also look into a possible alternative approach for the prediction-ground-truth associations.

In the meantime, here are some other comments that you could look into.

src/stratigraphy/metadata/elevation_extraction.py
```diff
@@ -129,30 +108,38 @@ def get_elevation_near_key(self, lines: list[TextLine], page: int) -> Elevation
...
         Args:
-            extracted_elevation_list (list[Elevation]): A list of extracted elevation information.
+            extracted_elevation_list (list[FeatureOnPage[Elevation]]): A list of extracted elevation information.
 
         Returns:
-            Elevation | None: The best extracted elevation information.
+            list[FeatureOnPage[Elevation]]: The best extracted elevation information.
         """
+        # first remoove the dupplicate. Even if 2 boreholes have the same elevation, it will be handled later. Keeping
+        # duplicates here could only harm us.
```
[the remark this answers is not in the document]
TicaGit: Because it could happen that the elevation is written twice on the document. So the pipeline would interpret this as if there was 2 distinct borehole profile.

src/stratigraphy/util/predictions.py
```diff
@@ -41,19 +39,164 @@ def to_json(self) -> dict:
...
+                zip(
+                    self._layers_in_document.boreholes_layers,
+                    self._groundwater_in_doc.borehole_groundwaters,
+                    self._bounding_boxes,
+                    self._elevations_list,
+                    self._coordinates_list,
+                    strict=True,
+                )
+            )
+        ]
```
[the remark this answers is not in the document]
TicaGit: Yes I agree, this feels risky. I'll create a ticket.

src/stratigraphy/util/predictions.py
```diff
@@ -41,19 +39,164 @@ def to_json(self) -> dict:
...
+        return lst
+
 
 class OverallFilePredictions:
     """A class to represent predictions for all files."""
 
     def __init__(self) -> None:
         """Initializes the OverallFilePredictions object."""
         self.file_predictions_list: list[FilePredictions] = []
+        self.matching_pred_to_gt_boreholes = dict()  # set when evaluating layers
```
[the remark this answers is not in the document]
TicaGit: Okey, thanks, let me know!

src/stratigraphy/evaluation/groundwater_evaluator.py
TicaGit: We were discussing with Davis about a new metric that would penalise not having the same number of borehole as the ground truth. But this felt as another ticket so I did not think of it further.

src/stratigraphy/metadata/elevation_extraction.py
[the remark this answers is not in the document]
stijnvermeeren-swisstopo: Maybe better would be to decide the number of extracted boreholes based on the number of stratigraphies (depths & layers), and to drop any additional values for elevation, coordinates, etc.

src/stratigraphy/extract.py
```diff
@@ -33,7 +33,7 @@
 class ProcessPageResult:
     """The result of processing a single page of a pdf."""
 
-    predictions: list[Layer]
+    predictions: list[list[Layer]]
     bounding_boxes: list[BoundingBoxes]
```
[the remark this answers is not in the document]
TicaGit: Yes, I can refactor the class.

src/stratigraphy/metadata/metadata.py
```diff
@@ -126,81 +140,51 @@ def from_document(cls, document: fitz.Document) -> "BoreholeMetadata":
...
-            elevation=elevation,
-            coordinates=coordinates,
             language=language,
             page_dimensions=page_dimensions,
             filename=Path(filename),
         )
 
 
 @dataclass
-class OverallBoreholeMetadata:
```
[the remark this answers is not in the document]
TicaGit: No indeed, I'll delete it.

tests/test_groundwater.py
```diff
@@ -24,6 +25,84 @@ def groundtruth():
...
+                    FeatureOnPage.from_json(
+                        {
+                            "depth": 2.22,
+                            "date": "2016-04-18",
+                            "elevation": 448.07,
+                            "page": 1,
+                            "rect": [0, 0, 100, 100],
+                        },
+                        Groundwater,
+                    ),
```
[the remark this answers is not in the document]
TicaGit: Sure, let me clarify that :)

src/stratigraphy/extract.py
```diff
@@ -135,7 +125,7 @@ def process_page(self) -> ProcessPageResult:
             [layer for layer in borehole_layers if layer.description_nonempty()]
             for borehole_layers in layer_predictions
         ]
-        return ProcessPageResult(layer_predictions, bounding_boxes)
+        return [ExtractedBorehole(lays, bbox) for lays, bbox in zip(layer_predictions, bounding_boxes, strict=True)]
```
stijnvermeeren-swisstopo, line 128: It would be cleaner to create each `ExtractedBorehole` instance in a single for loop over `filtered_pairs`, instead of creating the layers and bounding boxes separately and the afterwards zipping them together.

I would also avoid shortening variable names too much. The loss in clarity when writing "lays" instead of "layers" is more important to me than saving a few characters in the source code.

src/stratigraphy/layer/duplicate_detection.py
```diff
@@ -111,7 +114,12 @@ def remove_duplicate_layers(
                 count_consecutive_non_duplicate_layers += 1
         non_duplicated_layers.append(sorted_layers[first_non_duplicated_layer_index:])
```
stijnvermeeren-swisstopo, line 115: Wouldn't it be easier to directly construct the list of `ExtractedBorehole` instances that we want to return, instead of first creating this list of "non_duplicated_layers" and then only "retroactively" adding the bounding boxes to them again?

src/stratigraphy/main.py
```diff
@@ -259,58 +258,53 @@ def start_pipeline(
...
-                    )
-                    # avoid duplicate entries
-                    seen_entry = [
-                        seen for borehole_gw in aggregated_groundwater_entries.values() for seen in borehole_gw
-                    ]
-                    for groundwater_entry in groundwater_entries_near_bbox:
-                        if groundwater_entry not in seen_entry:
-                            aggregated_groundwater_entries[borehole_index].append(groundwater_entry)
+                for borehole_index, extracted_borehole in enumerate(processed_page_results.boreholes_layers_with_bb):
+                    for page_bounding_box in extracted_borehole.bounding_boxes:
```
stijnvermeeren-swisstopo, line 268: at this point, we will only every have one entry in this list of bounding boxes, is that correct?

tests/test_groundwater.py
```diff
@@ -175,13 +129,22 @@ def test_evaluate_with_ground_truth(groundtruth, example_groundwater_1: dict):
...
+def test_evaluate_multiple_documents(groundtruth, groundwater_at_2m22, groundwater_at_3m22):
+    """Test the evaluate method with multiple groundwater entries on two documents.
 
-    Also, the second document has multiple borehole, and the matching is 0->1 and 1->0
+    On the first document, there is one borehole with two groundwater measurement (like in the previous test).
+    On the second document, there is two boreholes, each with one groundwater measurement. For those boreholes, the
+    matching with the ground truth is 0->1 and 1->0. Meaning that the first borehole of the prediction matches the
+    second entry in the groudtruth file.
     """
     # Sample groundwater entries
```
stijnvermeeren-swisstopo, line 140: It's a bit more verbose, but I would find it easier to understand what the test is about, if the evaluator data is constructed more directly/explicitly, e.g. like this:
```
    evaluator = GroundwaterEvaluator(
        groundwater_list=[
            FileGroundwaterWithGroundTruth(
                filename="example_borehole_profile.pdf",
                boreholes=[
                    BoreholeGroundwaterWithGroundTruth(
                        groundwater=GroundwatersInBorehole([groundwater_at_2m22, groundwater_at_3m22]),
                        ground_truth=groundtruth.for_file("example_borehole_profile.pdf").get(0)["groundwater"],
                    )
                ],
            ),
            FileGroundwaterWithGroundTruth(
                filename="example_borehole_profile_2.pdf",
                boreholes=[
                    BoreholeGroundwaterWithGroundTruth(
                        groundwater=GroundwatersInBorehole([groundwater_at_2m22]),
                        ground_truth=groundtruth.for_file("example_borehole_profile_2.pdf").get(1)["groundwater"],
                    ),
                    BoreholeGroundwaterWithGroundTruth(
                        groundwater=GroundwatersInBorehole([groundwater_at_3m22]),
                        ground_truth=groundtruth.for_file("example_borehole_profile_2.pdf").get(0)["groundwater"],
                    )
                ],
            ),
        ]
    )
```
What do you think?

stijnvermeeren-swisstopo (changes requested):

src/stratigraphy/extract.py
TicaGit: Ok, so you mean a classical for-loop, like:
```
for pair in filtered_pairs:
    ...
   layers = ...
   bbox = ...
   all_extracted_boreholes.append(ExtractedBorehole(layers, bbox))
```

Noted! I'll prioritize clarity over shorter names.

src/stratigraphy/layer/duplicate_detection.py
TicaGit: Yes totally, thanks for the suggestion.

src/stratigraphy/main.py
TicaGit: Yes, I believe it's correct. I could change the code to avoid the for-loop. Something like `page_bounding_box = extracted_borehole.bounding_boxes[0]`

stijnvermeeren-swisstopo (approved):
LGTM!

PR #152, opened 2025-03-12
refactor predictions - ground truth associations

stijnvermeeren-swisstopo:
@TicaGit something like this is what I had in mind when reviewing the other PR.

It still needs some cleaning up, and I'm not entirely happy with the amount of "wrapper dataclasses" that I had to define for this, but overall the logic seems much cleaner / easier to follow like this.

Unit tests don't work yet. The main problem there is that we have unit tests that seem to test two things at once: the matching of extracted data with ground truth data, as well as the evaluation of the individual attributes. We should probably split that into two.

In the evaluation, extracted data without a corresponding ground truth now always counts as false positives. This decreases the scores a little bit, also for Geoquat, as there are a few examples where one borehole in the ground truth is extracted in two parts by the pipeline. I think that in this case, it is correct that we score this lower, than when we would extract the same layers as a single borehole.

### DIFF
--- a/src/stratigraphy/util/predictions.py
+++ b/src/stratigraphy/util/predictions.py
@@ -134,7 +134,7 @@ class AllBoreholePredictionsWithGroundTruth:
                 file.filename,
                 [
                     BoreholeMetadataWithGroundTruth(
-                        predictions.predictions.metadata,
+                        predictions.predictions.metadata if predictions.predictions else None,
                         predictions.ground_truth.get("metadata", {}),
                     )
                     for predictions in file.boreholes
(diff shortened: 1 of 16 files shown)


src/stratigraphy/util/predictions.py
```diff
@@ -159,157 +99,87 @@ def build(self) -> list[BoreholePredictions]:
...
         while len(lst) < target_length:
-            lst.append(deepcopy(base_elem))  # Append copies to match the required length
-
-        return lst
-
+            lst.append(base_elem())  # Append copies to match the required length
 
-class OverallFilePredictions:
-    """A class to represent predictions for all files."""
+        return lst[:target_length]
```
TicaGit, line 117: Say we detect only one elevation where there is 2 boreholes on one document. The idea of using deepcopy is that both borehole would have a elevation information, which could be correct (maybe they really have the same elevation and the information was meant to refer to both of the boreholes). In my opinion it is better to have some information, even if it is likelly to be incorrect, rather than no information at all for one of the borehole. What do you think?

TicaGit (approved):

PR #156, opened 2025-03-18
Workaround for issue with cache contamination in pymupdf

stijnvermeeren-swisstopo:
Ensure the cache is clear before creating a visualization of some PDF. Otherwise, it can occur that certain cached objects from one PDF will break the visualization of another PDF. (See also https://github.com/swisstopo/swissgeol-boreholes-suite/issues/1935.)

E.g. this error occurs with Nagra PDFs when first processing `Dossier III_Appendix A_BOZ1-1_Lithostratigraphy_1_2500_0000-1037.39.pdf` and then `Dossier III_Appendix A_BUL1-1_Lithostratigraphy_1_2500_0000-1370.pdf`. The representation of the first (quaternary) layer will then look blue instead of black-on-white.

I will report this as an issue to the PyMuPDF library. -> https://github.com/pymupdf/PyMuPDF/issues/4388

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dependencies = [
     "langdetect>=1.0.9",
     "regex",
     "backoff",
-    "PyMuPDF>=1.23.26",
+    "PyMuPDF>=1.25.4",
     "opencv-python-headless",
     "quads>=1.1.0",
     "numpy<2",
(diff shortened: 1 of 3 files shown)


letao (approved):
Looks good to me. Did not see any issues here.

Issue #145, opened 2025-03-04
Option to output stratigraphy per borehole to CSV files


stijnvermeeren-swisstopo:
Add a command-line option (maybe something like `--csv`) to the `boreholes-extract-all` script to trigger an additional CSV output format (in addition to the `predictions.json` file). If enabled, then...
- Create a CSV file for every boreholes profile (currently one per PDF file, after #134 potentially multiple profiles per file)
- Suggested naming for an input file `example.pdf`: `example.csv`
  - in the case of several profiles per file: `example_1.csv`, `example_2.csv`, etc.
- Log these files as artifacts in MLFlow in a directory `csv/` (similar to how we log the PNG visualizations in a directory `pages/`)
- Include a header `from_depth,to_depth,material_description` as the first line of the CSV file
- Depth values can be left empty, if no value is availble

Example of how the first lines of the CSV should look for the example input file A42.pdf (from the Geoquat dataset):
```
from_depth,to_depth,material_description
,0.25,"graugrüner Sand,stark kiesig,Kies eckig"
0.25,2.00,"brauner nach unten zu dunkelbrauner, siltiger Sand mit Kies,Kieskomponenter vorwiegend fein,weich-bis mittelplas- tisch."
```

![Image](https://github.com/user-attachments/assets/34b503de-132a-4748-acab-a0a37c6d7779)

PR #157, opened 2025-03-21
Implementation of issue 145: option to output stratigraphy to csv files

letao:
Implementation of: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/145 

Added the option to output a csv file for each identified borehole in each file. By default this option is disabled and can be enabled using the '-c' flag during the pipeline execution. The proposed solution adds a to_csv() function to the FilePredictions class that can be applied to borehole_predictions_list and converts layers to csv file. 

For one file that host multiple borehole profiles multiple csv files are created, one for each identified borehole. CVS File names are created from the input file name and in cases where multiple boreholes are identified the names are parametrized with a suffix created from the borehole_index.  

Csv files are logged in MLFlow  as artifacts under the csv/ directory.

Note: Currently the implementation works as part of the main for loop and runs for each file as part of the extraction.

### DIFF
--- a/src/stratigraphy/main.py
+++ b/src/stratigraphy/main.py
@@ -28,7 +28,7 @@ from stratigraphy.metadata.metadata import FileMetadata, MetadataInDocument
 from stratigraphy.text.extract_text import extract_text_lines
 from stratigraphy.util.file_predictions import FilePredictions
 from stratigraphy.util.overall_file_predictions import OverallFilePredictions
-from stratigraphy.util.predictions import BoreholeListBuilder, BoreholePredictions, CsvPredictions
+from stratigraphy.util.predictions import BoreholeListBuilder, BoreholePredictions
 from stratigraphy.util.util import flatten, read_params
 
 load_dotenv()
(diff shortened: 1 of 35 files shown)


src/stratigraphy/util/file_predictions.py
```diff
@@ -24,3 +26,40 @@ def to_json(self) -> dict:
             **self.file_metadata.to_json(),
             "boreholes": [borehole.to_json() for borehole in self.borehole_predictions_list],
         }
+
+    def to_csv(self) -> dict:
```
stijnvermeeren-swisstopo, line 30: Return type is a list, not a dict.

Also, wouldn't it be cleaner to implement the to_csv method on the `BoreholePredictions` class? Then the method could just return the contents of a single csv file, instead of a list. And the usage in `main.py` would also become simpler (no need to instantiate `FilePredictions(borehole_predictions_list, file_metadata, filename)` at all.

src/stratigraphy/main.py
```diff
@@ -336,6 +347,20 @@ def start_pipeline(
             # Add file predictions
             predictions.add_file_predictions(FilePredictions(borehole_predictions_list, file_metadata, filename))
 
+            # Add layers to a csv file
+            if csv:
+                base_path = out_directory / Path(filename).stem
```
stijnvermeeren-swisstopo, line 352: Let's maybe use a new subdirectory `csv` underneath the `out_directory`.

stijnvermeeren-swisstopo (changes requested):

src/stratigraphy/util/file_predictions.py
letao: Agreed, your suggestion is more elegant. I have made changes in the lates commit to address this, the method has been moved to BoreholePredictions class.

src/stratigraphy/main.py
letao: Agreed, this has now been implemented

stijnvermeeren-swisstopo (approved):
LGTM!

Issue #155, opened 2025-03-17
USCS classification: first baseline with substring matching / simple regexes


stijnvermeeren-swisstopo:
After #154 , implement a first simple/naive algorithm for USCS classification.

- Use only substring matching / simple regexes, based on a list of relevant keywords for each USCS class.
- Don't spend more than a day one fine-tuning the matching rules and keywords.
- Check how much accuracy we can achieve with such a simple approach, and use this as a baseline for setting goals and defining follow-up tickets.

The following list of possible USCS classes + labels in different languages can be useful:
[codelist_uscs_type.csv](https://github.com/user-attachments/files/19288397/codelist_uscs_type.csv)

Issue #154, opened 2025-03-17
USCS: initial version of classification script and evaluation


stijnvermeeren-swisstopo:
Setup a script to identify USCS classes from material description, and evaluate the accuracy of the extraction/classification against a ground truth file (currently only available for the GeoQuat dataset, and dependent on #150).

- Should be able to run the independently from the existing borehole data extraction pipeline
- For now, extract the USCS classes directly from the material descriptions as they are defined in the ground truth file, so that we can benchmark the USCS classification independently from the PDF data extraction. 
- Only extract/classify USCS-1 for now, ignore USCS-2.
- Log the results as a new experiment in MLFlow
  - Track F1 score, recall and precision, both overall as well as per language (German, French, English)
- For now, don't include the borehole context (e.g. descriptions of other layers) as an input feature. Rather extract the USCS class directly from each individual material description for each layer. This might limit the achievable accuracy in case we have descriptions such as "dito" or "like 3m-4.2m", but that is something we will only try to deal with later.
- Don't start implementing a real classification algorithm/model yet. For now, we can test the script by using a "dummy" classification that picks the common class "CL-ML" for every layer, regardless of the description.

PR #158, opened 2025-03-21
Feat/issue 154/initial classification pipeline

TicaGit:
Solves issue #154 and partially addresses #155.

This PR introduces a pipeline for classifying the soil description of layers into USCS classes.

For testing the classification workflow, a dummy classifier can be used. Additionally, I have started implementing the Baseline classifier, which performs keyword matching between the USCS class descriptions and the layer's material description.

The description in the README.predictions-json.md was also updated to match the current format.


src/description_classification/main.py
```diff
@@ -0,0 +1,100 @@
...
+    logger.info(f"classification metrics: {classification_metrics.to_json()}")
+    logger.debug(f"classification metrics per class: {classification_metrics.to_json_per_class()}")
+
+    if mlflow_tracking:
+        mlflow.end_run()
+
+
+if __name__ == "__main__":
+    # launch with: python -m src.description_classification.main -f data/geoquat_ground_truth.json
+    # or with: boreholes-classify-descriptions  -f data/geoquat_ground_truth.json
```
stijnvermeeren-swisstopo, line 99: Do you think it would make sense to add an option to only run the classification on description from files that are in a specific dataset (train/test/validation). Once, we have a proper classifier, always running the classification on the full Geoquat data might become quite slow. Also, the split would allow us to check whether we might be overfitting.

stijnvermeeren-swisstopo (changes requested):
Is there a way (e.g. JSON file / MLFLow artifact?) to see the actual predictions for every description?

src/description_classification/main.py
TicaGit: Yes, I think in the future we want to be able to pass the predictions.json file generated. by the extraction pipeline. This requires to map each of the detected layers in the predictions.json to the real layers in the ground truth, so it might require more work.

TicaGit: Regarding your question about outputting the actual predictions for every description, no there is no such option yet. But on MLFLOW, you can see the detail for each predicted class.

src/description_classification/evaluation/evaluate.py
```diff
@@ -0,0 +1,209 @@
...
+    classes = set(predictions) | set(ground_truth)  # Get all unique classes
+    pred_counter = Counter(predictions)
+    gt_counter = Counter(ground_truth)
+
+    metrics_per_class = {}
+    for cls in classes:
+        tp = min(pred_counter[cls], gt_counter[cls])  # Intersection
+        fp = pred_counter[cls] - tp
+        fn = gt_counter[cls] - tp
+        metrics_per_class[cls] = Metrics(tp=tp, fp=fp, fn=fn)
```
stijnvermeeren-swisstopo, line 197: I have some doubts about this computation. Does this even check if the predictions are aligned with the ground truth? If we randomly shuffle the correct classes, wouldn't this still give us a 100% F1 score, even though most descriptions will have an incorrect class?

I would have expected the computation to go more along these lines (ideally we would also avoid the `zip`, and instead already pass each prediction and corresponding ground truth value pairwise together to this method):
```
        tp = sum([1 for pred, gt in zip(predictions, ground_truth) if pred == gt == cls])
        fp = sum([1 for pred, gt in zip(predictions, ground_truth) if pred == cls and gt != cls])
        fn = sum([1 for pred, gt in zip(predictions, ground_truth) if pred != cls and gt == cls])
```

TicaGit: Yes, you're totally right, great spot!

src/description_classification/utils/data_loader.py
```diff
@@ -0,0 +1,145 @@
...
+                uscs_class = map_most_similar_uscs(layer["uscs_1"])
+                layer_descriptions.append(
+                    LayerInformations(
+                        filename,
+                        borehole["borehole_index"],
+                        layer_index,
+                        language,
+                        layer["material_description"],
+                        uscs_class,
+                        None,
```
stijnvermeeren-swisstopo, line 73:
```suggestion
                        prediction_uscs_class=None,
```

src/description_classification/utils/data_loader.py
```diff
@@ -0,0 +1,145 @@
...
+                    )
+                )
+
+    return layer_descriptions
+
+
+def get_data_language_count(layer_descriptions: list[LayerInformations]) -> dict[str:int]:
```
stijnvermeeren-swisstopo, line 80: `dict[str:int]` not correct type?

src/description_classification/utils/data_loader.py
```diff
@@ -0,0 +1,145 @@
...
+        layer_descriptions (list[LayerInformations]): All the layers.
+
+    Returns:
+        dict[str:int]: the count for each language.
+    """
+    language_counts = dict(Counter(layer.language for layer in layer_descriptions))
+    return language_counts
+
+
+def get_data_class_count(layer_descriptions: list[LayerInformations]) -> dict[str:int]:
```
stijnvermeeren-swisstopo, line 93: `dict[str:int]` not correct type?

src/description_classification/utils/data_loader.py
```diff
@@ -0,0 +1,145 @@
...
+        layer_descriptions (list[LayerInformations]): All the layers.
+
+    Returns:
+        dict[str:int]: the count for each class.
+    """
+    class_counts = dict(Counter(layer.ground_truth_uscs_class.name for layer in layer_descriptions))
+    return class_counts
+
+
+def write_predictions(layers_with_predictions: list[LayerInformations], out_dir: Path):
```
stijnvermeeren-swisstopo, line 106: writing predictions does not really relate to "data loader", maybe put this in a separate python file?

src/description_classification/evaluation/evaluate.py
```diff
@@ -0,0 +1,260 @@
...
+    Attributes:
+        global_metrics (dict[USCSClasses, Metrics]): A dictionary containing the classification metrics
+            for each of the USCS classes at a global level.
+        language_metrics (dict[str, dict[USCSClasses, Metrics]]): A dictionary where each key represents
+            a supported language. Each value is another dictionary containing the classification metrics
+            for each of the USCS classes in that language.
+    """
+
+    global_metrics: dict[USCSClasses:Metrics]
+    language_metrics: dict[str : dict[USCSClasses:Metrics]]
```
stijnvermeeren-swisstopo, line 34: Dict type with , instead of :

src/description_classification/evaluation/evaluate.py
```diff
@@ -0,0 +1,260 @@
...
+        precisions = []
+        recalls = []
+        f1s = []
+
+        for m in metric_list:
+            tp, fp, fn = m.tp, m.fp, m.fn
+
+            precision = tp / (tp + fp) if (tp + fp) > 0 else 0.0
+            recall = tp / (tp + fn) if (tp + fn) > 0 else 0.0
+            f1 = 2 * precision * recall / (precision + recall) if (precision + recall) > 0 else 0.0
```
stijnvermeeren-swisstopo, line 61: Why not use the precision, recall and f1 methods from the Metrics class?

src/description_classification/evaluation/evaluate.py
```diff
@@ -0,0 +1,260 @@
...
+        """
+        if not metric_list:
+            return {"micro_precision": 0, "micro_recall": 0, "micro_f1": 0}
+
+        all_aggregated_metric = Metrics.micro_average(metric_list)
+
+        return {
+            "micro_precision": round(all_aggregated_metric.precision, 2),
+            "micro_recall": round(all_aggregated_metric.recall, 2),
+            "micro_f1": round(all_aggregated_metric.f1, 2),
```
stijnvermeeren-swisstopo, line 96: 4 decimal places instead of 2

src/description_classification/evaluation/evaluate.py
```diff
@@ -0,0 +1,260 @@
...
+
+    return all_classification_metrics
+
+
+def per_class_metrics(predictions: list[USCSClasses], ground_truth: list[USCSClasses]) -> dict[USCSClasses, Metrics]:
+    """Compute per-class classification metrics.
+
+    Args:
+        predictions (List[str]): List of predicted class labels.
+        ground_truth (List[str]): List of actual class labels.
```
stijnvermeeren-swisstopo, line 233: types not consistent

stijnvermeeren-swisstopo: Suggestion to avoid the zip:
```
def per_class_metrics(layers: list[LayerInformations]) -> dict[USCSClasses, Metrics]:
    all_classes = set([layer.prediction_uscs_class for layer in layers]) | set([layer.ground_truth_uscs_class for layer in layers])
    metrics_per_class = {}

    for cls in all_classes:
        tp = sum(1 for layer in layers if layer.prediction_uscs_class == layer.ground_truth_uscs_class == cls)
...
```

src/description_classification/evaluation/evaluate.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+def log_metrics_to_mlflow(all_classification_metrics: AllClassificationMetrics, log_per_class=False):
+    """Log metrics to MLFlow with error handling."""
+    # Log overall metrics
+    for name, value in all_classification_metrics.to_json().items():
+        mlflow.log_metric(name, value)
+
+    # Log per-class metrics if requested
+    if log_per_class:
```
stijnvermeeren-swisstopo, line 258: Consensus to totally remove this flag and always log the per-class metrics.

src/description_classification/evaluation/evaluate.py
```diff
@@ -192,69 +182,63 @@ def to_json_per_class(self) -> dict[str, float]:
...
 
 
-def evaluate(layer_descriptions: list[LayerInformations], log_per_class: bool = False) -> AllClassificationMetrics:
+def evaluate(layer_descriptions: list[LayerInformations]) -> AllClassificationMetrics:
     """Evaluates the predictions of the LayerInformations objects against the ground truth.
 
     Args:
-        layer_descriptions (list[LayerDescriptionWithGroundTruth]): the LayerInformations objects
+        layer_descriptions (list[LayerInformations]): the LayerInformations objects
         log_per_class (bool, optional): whether to log per-class metrics. Defaults to False.
```
stijnvermeeren-swisstopo, line 190: arg should be removed from doc string

stijnvermeeren-swisstopo (changes requested):
one more small thing to clean up

stijnvermeeren-swisstopo (approved):

Issue #142, opened 2025-03-03
Fatal error due to date parsing


TicaGit:
**Description**

When running the boreholes-extract-all data extraction pipeline, some pdf's cause the program to fail, for example, geoquat/test/A378.pdf or geoquat/train/A190.pdf.

This is due to some numbers in the pdf being wrongly identified as a date at line 16 of src/stratigraphy/groundwater/utility.py. Proper validation should be implemented.

![Image](https://github.com/user-attachments/assets/b4f0a365-882f-4a10-b25f-6cd99cd66410)

![Image](https://github.com/user-attachments/assets/46a05909-94a2-4b49-85e0-e42adb838458)

**Steps to reproduce**

boreholes-extract-all -i data/geoquat/test/A378.pdf

Issue #151, opened 2025-03-12
Find an "intelligent" way of associating boreholes element to the correct borehole profile


TicaGit:
Issue: Unreliable Assignment of Extracted Data to Borehole Profiles

Context:
This issue was initially raised in Pull Request #150, which aimed to resolve Issue#134.

Problem:
After the extraction process, multiple lists of elements (e.g., elevations, bounding boxes, lists of layers, etc.) are zipped together and assigned to borehole profiles based on their order. However, this approach is not robust, as the extraction order does not guarantee that elements correspond correctly to the intended borehole profiles.

Expected Behavior:
The extracted elements should be reliably mapped to their corresponding borehole profiles based on explicit identifiers rather than relying on list order.

Possible Solution:
- Validate the extracted data with the position of the elements to ensure correctness before association.

stijnvermeeren-swisstopo:
Acceptance criteria:
- Improved accuracy for coordinates for 269126143-bp.pdf
- Scores not worse for any other input files

Issue #118, opened 2025-02-18
Handling ValueError for invalid date strings encountered during date extraction


stijnvermeeren-swisstopo:
[Migrated from LGVISIUM-105.]

The function extract_date in gorundwater/utility.py will currently raise a ValueError when a string gets matched to the date pattern, but fails the format requirements in strptime. (f.e. a string like: `33.12.89`). This causes the pipeline to fail, as the raised ValueError will not be handled.

```
def extract_date(text: str) -> tuple[date | None, str | None]:
    """Extract the date from a string in the format dd.mm.yyyy or dd.mm.yy."""
    date_match = regex.search(r"(\d{1,2}\.\d{1,2}\.\d{2,4})", text)

    if not date_match:
        return None, None

    date_str = date_match.group(1)
    date_format = "%d.%m.%y" if len(date_str.split(".")[2]) == 2 else "%d.%m.%Y"
    return datetime.strptime(date_str, date_format).date(), date_str 
```
Thus to handle potential ValueErrors raised during parsing, the logic of the function should be updated. For example by including a try-expect block to catch Value errors such that the pipeline continues to run even if an invalid date string is encountered.
Maybe add a unit test to verify the functions capability to handle invalid dates.

 
Screenshot of traceback:
![Image](https://github.com/user-attachments/assets/d8990e3e-6325-4bf8-87a1-6b389319c1a2)

PR #159, opened 2025-03-25
Feat/issue 151/multiple borehole smart merge

TicaGit:
Summary
This PR primarily addresses issue #151: Unreliable Assignment of Extracted Data to Borehole Profiles. Since this issue is related to groundwater data extraction, the PR also resolves issue #142 and its duplicate #118.

**Improvements in Data Assignment ("Smart Merge")**
- Borehole elements (elevation, coordinates, groundwater) are now correctly assigned to their respective borehole layers based on their position on the page, rather than relying on extraction order.
- BoreholeInDocument now stores all extracted groundwater data, but no assignments occur during extraction. Instead, all assignments happen in BoreholeListBuilder.build(), ensuring correctness.
- Elevation is no longer passed during the extraction phase. Missing groundwater attributes (depth, elevation) are assigned after matching for better accuracy and consistency.

**Groundwater Date Parsing Fixes**
- Improved regex to correctly match dates containing spaces (e.g., "24. 11.2017").
- Added error handling to prevent crashes when strings are incorrectly identified as dates.

### DIFF
--- a/src/stratigraphy/groundwater/groundwater_extraction.py
+++ b/src/stratigraphy/groundwater/groundwater_extraction.py
@@ -105,7 +105,7 @@ class Groundwater(ExtractedFeature):
             "elevation": self.elevation,
         }
 
-    def set_elevation_info(self, terrain_elevation):
+    def set_elevation_infos(self, terrain_elevation):
         if self.depth is None:  # if was not already set
             self.depth = round(terrain_elevation - self.elevation, 2)
         if self.elevation is None:  # if was not already set
@@ -140,7 +140,7 @@ class GroundwatersInBorehole:
 
     def set_elevation_infos(self, terrain_elevation):
         for entry in self.groundwater_feature_list:
-            entry.feature.set_elevation_info(terrain_elevation)
+            entry.feature.set_elevation_infos(terrain_elevation)
 
 
 @dataclass
--- a/src/stratigraphy/groundwater/groundwater_extraction.py
+++ b/src/stratigraphy/groundwater/groundwater_extraction.py
@@ -119,7 +119,7 @@ class Groundwater(ExtractedFeature):
         if round(terrain_elevation - self.elevation, 2) - self.depth > prec:  # account for approx, up to 0.5m
             logger.warning(
                 f"Extracted groundwater height informations (depth = {self.depth}, elevation = {self.elevation}) "
-                f"do not match the calcul 'terrain_elevation - gw_elevation = gw_depth': {terrain_elevation} "
+                f"do not match the constraint 'terrain_elevation - gw_elevation = gw_depth': {terrain_elevation} "
                 f"- {self.elevation} = {round(terrain_elevation - self.elevation, 2)} != {self.depth} ± {prec}"
             )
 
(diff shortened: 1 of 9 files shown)


stijnvermeeren-swisstopo (changes requested):
Woops, I added these comments on Wednesday but it seems like I failed to submit the review. I don't have time now to verify whether the comments are still valid, or if they might have been fixed in the meantime. Sorry about that! Please have a look yourself, and feel free mark any comments that no longer apply as "resolved".

src/stratigraphy/util/predictions.py
```diff
@@ -71,28 +76,27 @@ def build(self) -> list[BoreholePredictions]:
         """Creates a list of BoreholePredictions after ensuring all lists have the same length."""
         self._extend_length_to_match_boreholes_num_pred()
 
+        # for each element, perform the perfect matching with the borehole layers, and retrieve the matching dict
+        borehole_to_elevation_map = self._match_element_to_borehole(self._elevations_list)
+        borehole_to_coordinate_map = self._match_element_to_borehole(self._coordinates_list)
+        borehole_to_groundwater_map = self._match_element_to_borehole(
+            [gw.groundwater_feature_list for gw in self._groundwater_in_doc.borehole_groundwaters],
+        )
```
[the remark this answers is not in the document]
TicaGit: I implemented option 2. We obtain the same results but the main function is much cleaner now in my opinion, good suggestion.

src/stratigraphy/groundwater/utility.py
```diff
@@ -7,14 +7,26 @@
 
 def extract_date(text: str) -> tuple[date | None, str | None]:
     """Extract the date from a string in the format dd.mm.yyyy or dd.mm.yy."""
-    date_match = regex.search(r"(\d{1,2}\.\d{1,2}\.\d{2,4})", text)
+    date_match = regex.search(r"(\d{1,2})\s*\.\s*(\d{1,2})\s*\.\s*(\d{2,4})", text)
```
stijnvermeeren-swisstopo, line 10: Let's add a unit test that covers this method (in particular the input that failed previously).

stijnvermeeren-swisstopo (changes requested):
Create follow-up ticket to try to improve the following inputs where the accuracy for elevation extraction has decreased:
- 267123078-bp.pdf (Zurich)
- 267123083-bp.pdf (Zurich)
  - Also check why the elevation is still marked as correct, even though it is extracted as 451.65 instead of 451.72; do we allow for some rounding errors?)
- 267125216-bp.pdf (Zurich)
- A361.pdf (Geoquat validation)

src/stratigraphy/groundwater/groundwater_extraction.py
```diff
@@ -346,7 +333,6 @@ def extract_groundwater(
         page_number: int,
         lines: list[TextLine],
         document: fitz.Document,
-        terrain_elevation: FeatureOnPage[Elevation] | None,
```
stijnvermeeren-swisstopo, line 349: Remove this param from the docstring as well.

stijnvermeeren-swisstopo (changes requested):
During the review, I've discovered an issue with the match_predictions_with_ground_truth method. Predictions that are not with associated with any ground truth are ignored, instead of being counted as false positives. I will create a follow-up ticket for that.

stijnvermeeren-swisstopo (approved):

Issue #127, opened 2025-02-18
Deprecation warnings from pydantic


stijnvermeeren-swisstopo:
[Migrated from LGVISIUM-71.]

The pydantic library was introduced along with the API (LGVISIUM-64 / https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/75). The library complains about many deprecation warnings in the code that would need to be addressed.

When running `pytest`, the following warnings are shown:
```
env/lib/python3.11/site-packages/pydantic/_internal/_config.py:295
env/lib/python3.11/site-packages/pydantic/_internal/_config.py:295
  /home/stijn3/swissgeol-boreholes-dataextraction/env/lib/python3.11/site-packages/pydantic/_internal/_config.py:295: PydanticDeprecatedSince20: Support for class-based `config` is deprecated, use ConfigDict instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.10/migration/
    warnings.warn(DEPRECATION_MESSAGE, DeprecationWarning)

env/lib/python3.11/site-packages/pydantic/fields.py:1030: 16 warnings
  /home/stijn3/swissgeol-boreholes-dataextraction/env/lib/python3.11/site-packages/pydantic/fields.py:1030: PydanticDeprecatedSince20: Using extra keyword arguments on `Field` is deprecated and will be removed. Use `json_schema_extra` instead. (Extra keys: 'example'). Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.10/migration/
    warn(

```

PR #160, opened 2025-03-27
Fix/issue 127/pydantic depr warnings

TicaGit:
This PR addresses issue #127. It removes deprecation warnings, that are especially appearing when running pytest.

**Pydantic 2.x Compatibility**
The code has been updated to follow the new Pydantic 2.x format:
  - Replaced the old Config class with model_config = ConfigDict(...) for setting model options.
  - Moved example values from Field(...) into json_schema_extra inside the ConfigDict.

**Ignoring Warnings from Third-Party Libraries**
Some deprecation warnings come from external libraries (botocore and SWIG). Since we can’t fix these in our code, they are now ignored during test runs to keep the output clean.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,6 +97,10 @@ line-length = 119
 convention = "google"
 
 [tool.pytest.ini_options]
+# Some deprecation warnings come from external libraries (botocore and SWIG). Since we can’t fix these in our code,
+# they are now ignored during test runs to keep the output clean. Link to the existing issues:
+# SWIG: https://github.com/pymupdf/PyMuPDF/issues/3931
+# boto3: https://github.com/boto/boto3/issues/3889
 filterwarnings = [
     "ignore:.*Swig.*has no __module__ attribute:DeprecationWarning",
     "ignore:datetime.datetime.utcnow:DeprecationWarning:botocore"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -95,3 +95,9 @@ line-length = 119
 
 [tool.ruff.lint.pydocstyle]
 convention = "google"
+
+[tool.pytest.ini_options]
+filterwarnings = [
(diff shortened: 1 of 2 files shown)


src/app/common/schemas.py
```diff
@@ -215,15 +214,16 @@ class BoundingBoxesRequest(ABC, BaseModel):
...
     )
+
     page_number: int = Field(
         ...,
         description="""Page number within the document where the extraction is to be performed. This is a 1-based 
         index (e.g., 1 for the first page), applicable for multi-page files like PDFs.""",
-        example=1,
     )
 
+    model_config = ConfigDict(json_schema_extra={"example": {"filename": str(Path("1007.pdf")), "page_number": 1}})
```
stijnvermeeren-swisstopo, line 225: why not just
```suggestion
    model_config = ConfigDict(json_schema_extra={"example": {"filename": "1007.pdf", "page_number": 1}})
```

pyproject.toml
```diff
@@ -94,4 +94,10 @@ ignore = [
...
 
 [tool.ruff.lint.pydocstyle]
-convention = "google"
\ No newline at end of file
+convention = "google"
+
+[tool.pytest.ini_options]
+filterwarnings = [
+    "ignore:.*Swig.*has no __module__ attribute:DeprecationWarning",
+    "ignore:datetime.datetime.utcnow:DeprecationWarning:botocore"
```
stijnvermeeren-swisstopo, line 102: add comment with link to the underlying issue (e.g. GitHub issue) from the external libraries, so we can understand why we added these exception in the future, and check if we can at one point remove them again.

stijnvermeeren-swisstopo (changes requested):

stijnvermeeren-swisstopo (approved):
LGTM!

PR #166, opened 2025-04-02
Evaluate predictions from boreholes without matching ground truth as false positives

### LINKED ISSUE #165: Data from additionally extracted boreholes (no ground truth) should be considered as false positives
The method `match_predictions_with_ground_truth` method currently incorrectly ignores predictions that are not with associated with any ground truth, instead of counting them as false positives.

The e.g. has an influence on the scores for 684252058-bp.pdf.

stijnvermeeren-swisstopo:
fixes #165 

Some metrics decrease slightly, but I would argue that the scoring makes more sense like this.


TicaGit (approved):
Looks good

PR #167, opened 2025-04-04
Refactor evaluation methods + show groundwater correctness on visualization

### LINKED ISSUE #124: Find a better way to associate the "correctness" from the evaluation with the extracted groundwater values
[Migrated from LGVISIUM-65.]

This issue originated from the PR of the groundwater level: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/72#issuecomment-2314949349 

The goals is to find a better way to associate the "correctness" from the evaluation with the individual extracted values, so that we can show (red/green) the correctness of each extracted groundwater measurement also in the visualisation.

In particular, solve the TODOs in `groundwater_evaluator.py` and `draw.py` that were added in https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/148.

stijnvermeeren-swisstopo:
Fixes #124

### DIFF
--- a/src/stratigraphy/evaluation/utility.py
+++ b/src/stratigraphy/evaluation/utility.py
@@ -1,32 +1,79 @@
 """Utility functions for evaluation."""
 
-from collections import Counter
+import dataclasses
+from collections.abc import Callable
+from typing import TypeVar
 
 from stratigraphy.evaluation.evaluation_dataclasses import Metrics
 from stratigraphy.layer.layer import LayerDepths
 
 
-def count_against_ground_truth(values: list[any], ground_truth: list[any]) -> Metrics:
-    """Count evaluation metrics by comparing predicted values against ground truth.
+@dataclasses.dataclass
+class EvaluationResults:
+    """Class that contains that data about which extracted entries could be matched with which ground truth values."""
 
-    Metrics are calculated as follows:
-    - True Positives (tp): Number of values that appear in both lists (counting duplicates)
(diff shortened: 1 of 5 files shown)


src/stratigraphy/evaluation/utility.py
```diff
@@ -1,32 +1,79 @@
...
+        """Derive the metrics (true positives, false positives, fales negatives) from the evaluation results."""
+        tp = sum(1 for is_correct in self.extracted_correct if is_correct)
+        return Metrics(
+            tp=tp,
+            fp=len(self.extracted_correct) - tp,
+            fn=len(self.ground_truth_correct) - tp,
+        )
+
+
+T = TypeVar("T")
```
TicaGit, line 29: Aren't T and U always the same type in the current code?

TicaGit (approved):
LGTM!

Maybe it could make sense that extracted and ground truth always have the same type in the future; I believe it is not the case only for coordinates. This would avoid the need for the two typeVars T and U.

Otherwise, the changes look good overall :)

stijnvermeeren-swisstopo:
@TicaGit No that is not entirely true. In particular, the coordinate evaluation in `metadata_evaluator.py` has an extracted `Coordinate` object but the ground truth is a `dict`. We could discuss if it would be better to force these objects to be of the same type, or if it's good to keep this flexibility. But maybe that's a bit out-of-scope for this PR.

letao:
LGTM!

PR #168, opened 2025-04-07
Feat/issue 162/fine tuning bert

TicaGit:
Solves #162 

Implements a classifier that uses the BERT model. 
Also implements a pipeline to load the model and fine-tune it on local data.


src/description_classification/models/train.py
```diff
@@ -0,0 +1,196 @@
...
+    # Load datasets
+    logger.info("Loading datasets (transformers librairy).")
+    train_dataset, eval_dataset = setup_data(bert_model, file_path)
+
+    # Define a custom compute_metrics function
+    def compute_metrics(eval_pred):
+        logits, labels = eval_pred
+        predictions = logits.argmax(axis=-1)
+        metrics = per_class_metric(predictions, labels)
+        return AllClassificationMetrics.compute_micro_average(metrics.values())
```
stijnvermeeren-swisstopo, line 179: This function returns three different metrics in a dict. Where do we define, which is the one that is used for computing the training loss, or does that not relate to this function?

stijnvermeeren-swisstopo (commented):
The accuracy of the fine-tuend BERT-classifier for most classes seems to be a little lower that our simple baseline. How easy do you think it would be to improve the accuracy, and what would be the best way to do so?

src/description_classification/utils/uscs_classes.py
```diff
@@ -9,6 +9,9 @@
 class USCSClasses(Enum):
     """USCS (Unified Soil Classification System) classes."""
 
+    def _generate_next_value_(name, start, count, last_values):
+        return count  # To start from 0 instead of 1
```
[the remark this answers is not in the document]
TicaGit: Technically, there's no strict requirement to start at 0. But in practice, it's a common convention, especially in machine learning, to start class labels from 0. I initially used auto() in the Enum, which starts at 1 by default. That led to a subtle bug where label 0 was unused, and the Trainer created an extra, empty class for it. Took me a couple hours to catch, so switching to 0-based indexing avoids that kind of issue.

src/description_classification/models/train.py
TicaGit: I believe it is only used for logging, the metric used for training is the loss of the model, which is crossentropy.

TicaGit:
Regarding your comment about improving the model accuracy, I would start by applying the tip I mentioned in the pull request description, and also increasing the training time. This is a quick way to potentially boost performance, without immediately diving into a full parameter grid search, which can be time-consuming. I can try to improve it this evening and tomorrtow morning, instead of picking a new issue.

stijnvermeeren-swisstopo:
@TicaGit No, I think that for this first step, the current accuracy is fine. We can discuss on Monday how we prioritize the possible next steps / which methods we focus on for further improvements.

stijnvermeeren-swisstopo:
LGTM!

stijnvermeeren-swisstopo (approved):

PR #170, opened 2025-04-08
optimize baseline, rm synonyms and stop words

stijnvermeeren-swisstopo:
@letao I would probably something like this:
- Multiple possible keyphrases per class, that also give us a better overview of the different descriptions that can be used for each class
- Stop words and synonyms did not improve the scores anymore, so I removed them. The synonyms were not working properly anyway, as we were compared stemmed tokens with synonyms that were not stemmed.
- Correctly predict "kA" if no keyphrases are matched
- Immediately allow matches with gaps, instead of first checking for exact matches

Like this, I currently get a overall macro F1 of 54.1% and a micro F1 of 68.5%.


letao:
@stijnvermeeren-swisstopo 

Good point on the synonyms and stemming, that was overlooked by me previously.
Also, agreed, it makes sense to remove the synonyms, multiple keyphrases per class solves the same issue that synonyms were trying to do - check multiple phrases for the same USCS class. The proposed solution is more elegant.

letao (approved):
LGTM! Great improvements in the accuracy. I can do some basic changes to docstrings and descriptions on the issue-155 branch

PR #171, opened 2025-04-09
Feat/issue 163/per class artifacts

TicaGit:
solves #163 

In the classification pipeline, creates artifacts detailed for each class, to improve data exploration after the classifiaction.


src/description_classification/utils/data_utils.py
```diff
@@ -73,3 +81,233 @@ def write_predictions(layers_with_predictions: list[LayerInformations], out_dir:
...
+    output_file = out_dir / "overview.csv"
+
+    # Define the columns you want in the CSV
+    fieldnames = ["class", "f1", "precision", "recall", "number_ground_truth", "number_prediction"]
+
+    with open(output_file, "w", newline="", encoding="utf-8") as f:
+        writer = csv.DictWriter(f, fieldnames=fieldnames)
+        writer.writeheader()
+
+        for class_ in USCSClasses:
```
stijnvermeeren-swisstopo, line 152: Can we sort these by number of ground truth entries? Now the classes are sorted rather randomly. Usually, the most common classes are the ones that I'm most interested in, so it would make more sense to put these on top of the output.

src/description_classification/utils/data_utils.py
```diff
@@ -73,3 +81,233 @@ def write_predictions(layers_with_predictions: list[LayerInformations], out_dir:
...
+                "f1": next(v for k, v in class_metrics_dict.items() if k.endswith("f1")),
+                "precision": next(v for k, v in class_metrics_dict.items() if k.endswith("precision")),
+                "recall": next(v for k, v in class_metrics_dict.items() if k.endswith("recall")),
+                "number_ground_truth": num_ground_truth,
+                "number_prediction": num_pred,
+            }
+            writer.writerow(row)
+
+
+def get_key_class_functions(
```
stijnvermeeren-swisstopo, line 170: I created a small refactoring proposal: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/172 . Let me know if you like it.

stijnvermeeren-swisstopo (changes requested):
While the output files could certainly be improved a little further, they were already very useful when analyzing and optimizing the baseline classifier from #155 . I would be happy to merge them like this. Based on further experience, we can always create some follow-up tickets to further improve these output files, if necessary.

I just have one minor comments, and one refactoring suggestion...

src/description_classification/utils/data_utils.py
TicaGit: Yes no problem. By the way, you probably knew but just in case, you can manually sort them on mlflow, clicking on the column header :)

stijnvermeeren-swisstopo (approved):
LGTM!

After you merge this into the main branch, could you then also merge main into your fined-tuned-BERT branch? Having these artifacts would make it easier to understand the behaviour of the BERT-based classifier.

PR #172, opened 2025-04-09
refactor grouping by ground truth / predictions

stijnvermeeren-swisstopo:
@TicaGit I would propose this small refactoring. Using a class like this seems a little "safer" to me (less risk of mixing up the elements of a tuple; less boolean values that can easily be set incorrectly). What do you think?

### DIFF
--- a/src/description_classification/utils/data_utils.py
+++ b/src/description_classification/utils/data_utils.py
@@ -5,6 +5,7 @@ import json
 import shutil
 from collections import Counter, OrderedDict
 from collections.abc import Callable
+from dataclasses import dataclass
 from pathlib import Path
 
 from description_classification.evaluation.evaluate import AllClassificationMetrics
@@ -117,7 +118,7 @@ def write_per_language_per_class_predictions(
         # groupby ground truth first
         write_per_class_predictions(
             layers_in_language,
-            groupby_ground_truth_first=True,
+            grouping_functions=group_by_ground_truth_functions,
             metrics_dict=metrics_dict_in_language,
             out_dir=out_dir / language / "group_by_ground_truth",
         )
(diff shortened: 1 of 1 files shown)


src/description_classification/utils/data_utils.py
```diff
@@ -167,55 +168,51 @@ def write_overview(metrics_dict: dict[str, float], layers_with_predictions: list
...
+    metric_used="recall",
+)
+
+group_by_prediction_functions = GroupingFunctions(
+    get_first_key_class=lambda layer: layer.prediction_uscs_class,
+    get_second_key_class=lambda layer: layer.ground_truth_uscs_class,
+    first_key_str="prediction",
+    second_key_str="ground_truth",
+    metric_used="precision",
+)
```
TicaGit, line 204: Looks good, I would just define those in the first function that is called (write_per_language_per_class_predictions). I feel like these variables should not be global to the file.

src/description_classification/utils/data_utils.py
```diff
@@ -167,55 +168,51 @@ def write_overview(metrics_dict: dict[str, float], layers_with_predictions: list
             writer.writerow(row)
 
 
-def get_key_class_functions(
-    groupby_ground_truth_first: bool,
-) -> tuple[Callable[[LayerInformations], USCSClasses], Callable[[LayerInformations], USCSClasses], str, str, str]:
-    """Returns the appropriate functions and depending on whether we are grouping by ground truth or prediction.
+@dataclass
+class GroupingFunctions:
```
TicaGit, line 172: Name feels a bit ambiguous to me. What about something like KeyClassConfig?

TicaGit (approved):
If you think my comments are relevant, you can implement the changes and merge into my branch (or I can also do it).

src/description_classification/utils/data_utils.py
stijnvermeeren-swisstopo: I moved these definitions inline, directly where the are needed

src/description_classification/utils/data_utils.py
stijnvermeeren-swisstopo: sounds good

PR #173, opened 2025-04-09
Fix/issue 146/false positive exact matches

### LINKED ISSUE #146: Optimize parameters for exact matches in AAboveBInterval.matching_blocks()
The method `matching_blocks()` in `AAboveBInterval` (among other things) decides when there is an "exact match" between a depth interval and a material description, i.e. when their relative positioning is so unambiguous, that there is no need to check for alternative alignments.

However, it seems that sometime we get such an exact match, even when this is not correct. For example, in the Nagra profile `Dossier III_Appendix_A_TRU1-1_Lithostratigraphy_1_2500_0000-1310.pdf`, there is an exact match for the description "_DOLOSTONE, dark beige to beige, ..., very fossiliferous, pyritic_"  with the depth interval 1090.46-1119.16, which is not correct (this description is actually linked to the preceding depth interval). Consequently, there is some misalignment in the output of the data extraction pipeline, in particular with a layer whose material description is incorrectly split (989.64-1003.95) and a material description which does not get linked to any depth interval at all (at 1119.16).

Image

When we increase the minimal vertical spacing to other material description lines that is required for an exact match from 5 to e.g. 10, the results for the input PDF improve,

Image

Tasks:
- Refactor the hard-coded values `5` in `matching_block()` (for `distance_above_ok_for_exact` and for `can_end_exact_match`) into a configurable parameter (a single parameter for both usages, defined in a `yml` file under `config/`).
- Test if it out benchmarks (Zurich, Geoquat) improve (or at least don't get worse) if we increase this value a little (e.g. from 5 to 10).
- In case the results get worse, look for alternative ways to improve the logic for exact matches in the `matching_blocks()` method

TicaGit:
Resolves #146

The initial issue was not reproducible, and the problems on Nagra were mostly related to those parasite depths that does not belong to any material descriptions (see: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/146#issuecomment-2786898212).

I remove the magic number and performed a quick parameter search, best value seems to be 10. (see https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/146#issuecomment-2789501775)

### DIFF
--- a/src/stratigraphy/depth/interval.py
+++ b/src/stratigraphy/depth/interval.py
@@ -7,9 +7,6 @@ import fitz
 from stratigraphy.lines.line import TextLine
 from stratigraphy.sidebar.sidebarentry import DepthColumnEntry
 from stratigraphy.text.textblock import TextBlock
-from stratigraphy.util.util import read_params
-
-matching_params = read_params("matching_params.yml")
 
 
 class Interval:
--- a/src/stratigraphy/depth/interval.py
+++ b/src/stratigraphy/depth/interval.py
@@ -7,6 +7,9 @@ import fitz
 from stratigraphy.lines.line import TextLine
 from stratigraphy.sidebar.sidebarentry import DepthColumnEntry
 from stratigraphy.text.textblock import TextBlock
+from stratigraphy.util.util import read_params
+
+matching_params = read_params("matching_params.yml")
 
 
 class Interval:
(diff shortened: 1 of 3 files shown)


src/stratigraphy/depth/interval.py
```diff
@@ -7,6 +7,9 @@
 from stratigraphy.lines.line import TextLine
 from stratigraphy.sidebar.sidebarentry import DepthColumnEntry
 from stratigraphy.text.textblock import TextBlock
+from stratigraphy.util.util import read_params
+
+matching_params = read_params("matching_params.yml")
```
stijnvermeeren-swisstopo, line 12: not used?

stijnvermeeren-swisstopo (changes requested):

stijnvermeeren-swisstopo (approved):
LGTM!

PR #174, opened 2025-04-10
Fix/issue 119/layer identifier sidbar fix

### LINKED ISSUE #119: Improve entry extraction for LayerIdentifierSidebar

TicaGit:
Resolves #119 

This PR adds a check to LayerIdentifierSidebar that rejects sidebars which do not follow a logical progression (e.g., 2 < 3a < 3b < 5 is valid). This check is necessary to prevent the columns of letter "m)" (indicating meters) in the file deepwell Arsch from being mistakenly identified as a sidebar.

Additionally, it rejects some sidebars in the files 12318, 5878, and 9394 in geoquat. All of these rejections are correct and result in slight improvements to the metrics. No other changes have been made for Zurich, deepwell, or Nagra.

The try-except block is used to handle cases where letters are mistakenly extracted as numbers, which would cause crashes during comparison (in B147 and 680). These cases do not impact the desired behavior.

This PR also changes the ordering of identified sidebars: they are now returned sorted by highest score first instead of lowest. This change ensures that when PDFs span multiple pages, the best identified borehole per page is retained. This behavior is desired for deepwell Arsch and geoquat/validation/A114067, and likely beneficial in other cases as well. Note that this last modification slightly lowers the groundwater recall for reasons that remain unclear, but the change is considered correct and intentional. 
This change won't be needed after issue #153 is resolved, as it should merge boreholes that are similar. A list of the pdf's that cause problem can also be found in this issue.

### DIFF
--- a/src/stratigraphy/sidebar/layer_identifier_sidebar_extractor.py
+++ b/src/stratigraphy/sidebar/layer_identifier_sidebar_extractor.py
@@ -69,7 +69,7 @@ class LayerIdentifierSidebarExtractor:
         result = []
         # Remove columns that are fully contained in a longer column
         # Also check that the columns are mostly ascending, both numerically and alphabetically (e.g. 2 < 3a < 3b
-        # < 5 is valid). Some errors are tolerated, as the indicators (especially in Geneva) sometimes carry a 
+        # < 5 is valid). Some errors are tolerated, as the indicators (especially in Geneva) sometimes carry a
         # semantic meaning instead of being strictly ordered (e.g. geoquat/validation/9394.pdf), and we also need to
         # account for human or OCR mistakes. We only reject a column if the error ratio is bigger than the parameter
         # layer_identifier_acceptance_ratio specified in the config file.
--- a/src/stratigraphy/sidebar/layer_identifier_sidebar_extractor.py
+++ b/src/stratigraphy/sidebar/layer_identifier_sidebar_extractor.py
@@ -68,10 +68,11 @@ class LayerIdentifierSidebarExtractor:
 
         result = []
         # Remove columns that are fully contained in a longer column
-        # Also check which columns do not follow a logical progression. Values should ascend numerically and
-        # alphabetically (e.g. 2 < 3a < 3b < 5 is valid). Some errors are tolerated to account for human or OCR
-        # mistakes (e.g. geoquat/validation/9394.pdf). We only reject a column if the error ratio is bigger than the
-        # parameter layer_identifier_acceptance_ratio specified in the config file.
+        # Also check that the columns are mostly ascending, both numerically and alphabetically (e.g. 2 < 3a < 3b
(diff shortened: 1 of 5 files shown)


stijnvermeeren-swisstopo:
@TicaGit I observe a change in the results for 684252058-bp.pdf. It seems related to fact that this borehole is split across several pages. Could you check that?

TicaGit:
I'm not 100% sure, but I think the change in 684252058-bp.pdf in just the groundwater entry switching place.

TicaGit:
Yes, I meant A11406 sorry. It has 3 pages, and on each pages 2 boreholes are detected (where only one should be).

src/stratigraphy/sidebar/layer_identifier_sidebar_extractor.py
```diff
@@ -68,7 +68,10 @@ def from_lines(cls, lines: list[TextLine]) -> list[LayerIdentifierSidebar]:
 
         result = []
         # Remove columns that are fully contained in a longer column
-        # Also remove columns that do not follow a logical progression (e.g. 2 < 3a < 3b < 5 is valid)
+        # Also check which columns do not follow a logical progression. Values should ascend numerically and
+        # alphabetically (e.g. 2 < 3a < 3b < 5 is valid). Some errors are tolerated to account for human or OCR
+        # mistakes (e.g. geoquat/validation/9394.pdf). We only reject a column if the error ratio is bigger than the
+        # parameter layer_identifier_acceptance_ratio specified in the config file.
```
stijnvermeeren-swisstopo, line 74: I would suggest the following phrasing:

> Also check that the columns are mostly ascending, both numerically and alphabetically (e.g. 2 < 3a < 3b < 5 is valid). Some errors are tolerated, as the indicators (especially in Geneva) sometimes carry a semantic meaning instead of being strictly ordered (e.g. geoquat/validation/9394.pdf), and we also need to account for human or OCR mistakes. We only reject a column if the error ratio is bigger than the parameter layer_identifier_acceptance_ratio specified in the config file.

stijnvermeeren-swisstopo (changes requested):

stijnvermeeren-swisstopo (approved):
Approved conditional on CI passing

Issue #161, opened 2025-04-01
USCS classification: chatbot prompting approach


stijnvermeeren-swisstopo:
Can use the Claude model that is available through AWS Bedrock (swisstopo can provide access and sample code, as we already used this in another project).

@letao and @stijnvermeeren-swisstopo coordinate setting up access (identity-based access preferred).

Scope: setting up the pipeline, include prompt in version control to allow future optimization, NO extensive optimization yet.

PR #175, opened 2025-04-10
Feat/issue 161/uscs classification chatbot prompting approach

letao:
Implementation of https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/161

The first version for the Chatbot classifier. The approach uses AWS Bedrock API through boto3. The classifier takes as input material descriptions and appends it to the prompt messages. Afterwards the message is sent to the LLM API. Two response options exist: 

- Simple answer with USCS classification category (used by default)
- Complex answer that uses Chain-of-Thought prompting with both reasoning and USCS classification category. 

At the moment Simple answer is preferred due to cost and faster response times. Accuracy also is for now better with the Simple answer pattern. 

The approach uses environment file with values for model id, API version and region (I have adjusted the example_env file accordingly). The next steps should be implementation of batch requests to improve response times.


letao:
@stijnvermeeren-swisstopo and @TicaGit, 

I have added some changes to the pull request:

- Added asynchronous method for concurrent API calls, however, we are currently blocked by call limit to the API endpoint, using multiple workers leads to rejection due to rate limits.
- Added logging of API call failures for files, boreholes and layers. 
- Added an option to store already processed predictions into separate files so that no predictions are lost due to an API call failure (this can be improved upon further, if needed)

src/description_classification/classifiers/classifiers.py
```diff
@@ -186,3 +197,176 @@ def classify(self, layer_descriptions: list[LayerInformations]):
...
+        Environment variables are used to configure the AWS region, model ID, and Anthropic version.
+        The USCS patterns and classification prompts are read from the configuration files.
+        """
+        self.bedrock_client = boto3.client(service_name="bedrock-runtime", region_name=os.environ.get("AWS_REGION"))
+
+        self.anthropic_version = os.environ.get("ANTHROPIC_VERSION")
+        self.model_id = os.environ.get("ANTHROPIC_MODEL_ID")
+
+        self.uscs_patterns = classification_params["uscs_patterns"]
+        self.classification_prompts = classification_prompts
```
TicaGit, line 217: classification_prompts is already global to the file, do we need to pass it to the class as an argument?  same for uscs_patterns.

src/description_classification/main.py
```diff
@@ -129,9 +130,13 @@ def main(file_path: Path, out_directory: Path, file_subset_directory: Path):
     logger.info(f"Loading data from {file_path}")
     layer_descriptions = load_data(file_path, file_subset_directory)
 
-    classifier: Classifier = BaselineClassifier()
+    classifier: Classifier = AWSBedrockClassifier()
     logger.info(f"Classifying layer description with {classifier.__class__.__name__}")
-    classifier.classify(layer_descriptions)
+
+    if isinstance(classifier, AWSBedrockClassifier):
+        asyncio.run(classifier.classify(layer_descriptions, store_files=True, max_concurrent_calls=1))
```
TicaGit, line 137: It would be nice to keep the same interface for all classify methods, as it is defined by the Protocol Classifier class. 
I would pass all those argument in the constructor of AWSBedrockClassifier instead . 
In the current version of the code (that I just merged from the fine-tune bert branch), there is already if-else statement that checks the Classifier type, so passing different  parameter here should be easy. Also parameter like max_tokens, temperature should be atributes of the classifier in my opinion.
This would also avoid the need for isinstance, which sometimes feels a bit confusing.

TicaGit (changes requested):
Small changes regarding the code, tell me if you think it does not make sense :)

src/description_classification/main.py
```diff
@@ -110,12 +121,12 @@ def common_options(f):
 
 @click.command()
 @common_options
-def click_pipeline(file_path: Path, out_directory: Path, file_subset_directory: Path):
+def click_pipeline(file_path: Path, out_directory: Path, out_directory_beadrock: Path, file_subset_directory: Path):
```
stijnvermeeren-swisstopo, line 124: typo: beadrock/bedrock

src/description_classification/main.py
```diff
@@ -97,6 +101,13 @@ def common_options(f):
         default=DATAPATH / "output_description_classification",
         help="Path to the output directory.",
     )(f)
+    f = click.option(
+        "-ob",
+        "--out-directory-beadrock",
```
stijnvermeeren-swisstopo, line 106: typo: beadrock/bedrock

src/description_classification/classifiers/classifiers.py
```diff
@@ -186,3 +194,178 @@ def classify(self, layer_descriptions: list[LayerInformations]):
...
+            anthropic_version (str): The version of the Anthropic model to use.
+            material_description (str): The material description to classify.
+            language (str): The language of the material description.
+
+        Returns:
+            body (dict): The message body for the Bedrock API.
+        """
+        language_patterns = self.uscs_patterns[language]
+
+        classification_prompts = read_params(os.environ.get("ANTHROPIC_PROMPT_TEMPLATE"))
```
stijnvermeeren-swisstopo, line 232: Where does this env variable come from ? Maybe add to `.env.template`, and ensure that we show a useful error message in case this variable is not set.

pyproject.toml
```diff
@@ -35,7 +35,8 @@ dependencies = [
     "quads>=1.1.0",
     "numpy<2",
     "rtree",
-    "nltk"
+    "nltk",
+    "boto3"
```
stijnvermeeren-swisstopo, line 39: duplicate?

stijnvermeeren-swisstopo (changes requested):
For the input
> Graviers, galets et boulets dans matrice sablo-limoneuse assez grossière, grise, pulvérulente. Quelques fragments de briques observés jusque vers 5.30 (remaniement de la couche lors des terrassements de la fouille blindée). Assise fortement décomprimée, lâche.
> Nappe phréatique entre 4.4 et 4.5m (8-14.9.1992).

I get the prediction "GW" (ground truth: "GM"). The reasoning from the model is as follows.

> Based on the material description, the key elements are:
> - Graviers, galets et boulets (Gravel, pebbles and boulders)
> - Matrice sablo-limoneuse assez grossière (Coarse sandy-silty matrix)
> - Quelques fragments de briques (Some brick fragments)
> - Assise fortement décomprimée, lâche (Highly decompressed, loose structure)
> 
> The presence of gravel, pebbles and boulders in a coarse sandy-silty matrix suggests a well-graded gravel with sand (GW) classification. The presence of some brick fragments indicates potential anthropogenic influence, but this does not change the primary soil classification.

Since the config file / system prompt only defines `GM: ["gravier limoneux"]` and `GW: ["gravier bien gradué"]`, I get the impression here that the language model has acquired significant knowledge about USCS classification already during its training, and in this case is relying more on this built-in knowledge and not so much on the definitions in our prompt.

I think that this is actually useful. It shows that we don't need to completely define the USCS classification in our prompt, but we can instead focus on refining certain class definitions where necessary.

Other explanations indicate that the language model also correctly parses and sometimes relies on our definitions.

Input:
> Marne altérée grise, irrégulièrement feuilletée, dure à très dure, très plastique. Assise très compacte, mais très sensible à l'eau // classes de sol:14/3

Explanation:
> Based on the material description, the key characteristics are:
> - Marne (marl) - a type of sedimentary rock composed of clay and limestone
> - Altered (altérée) - the material has been weathered or altered from its original state
> - Grey color
> - Irregularly layered (feuilletée) - the material has a layered or foliated structure
> - Hard to very hard consistency
> - Very plastic
> - Highly compact (assise très compacte) but sensitive to water
> 
> Given these characteristics, the material seems to be a highly weathered, clay-rich sedimentary rock that exhibits plastic behavior. This suggests it could be classified as a FELS (rocher, marne, marno, molasse, gypse, schistes) category, which encompasses various types of weathered or altered sedimentary rocks.

We should take these insights with us when we work on  https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/180.

As an initial version, this implementation looks good enough to me.

letao:
@stijnvermeeren-swisstopo and @TicaGit 

I have addressed all your comments and integrated the main branch to solve the merge conflicts. As a result I have now moved async processing to classifiers.py file to streamline the main.py file. Additional small changes have been implemented to align with current approach in main.py

pyproject.toml
letao: fixed

src/description_classification/main.py
```diff
@@ -115,9 +127,9 @@ def common_options(f):
     f = click.option(
         "-c",
         "--classifier-type",
-        type=click.Choice(["dummy", "baseline", "bert"], case_sensitive=False),
+        type=click.Choice(["dummy", "baseline", "bert", "bedrock"], case_sensitive=False),
         default="dummy",
-        help="Classifier to use for description classification. Choose from 'dummy', 'baseline', or 'bert'.",
+        help="Classifier to use for description classification. Choose from 'dummy', 'baseline', 'bert' or bedrock.",
```
stijnvermeeren-swisstopo, line 132:
```suggestion
        help="Classifier to use for description classification. Choose from 'dummy', 'baseline', 'bert' or 'bedrock'.",
```

pyproject.toml
```diff
@@ -36,10 +36,12 @@ dependencies = [
     "numpy<2",
     "rtree",
     "nltk",
+    "boto3",
```
stijnvermeeren-swisstopo, line 39: is still a duplicate

.env.template
```diff
@@ -5,3 +5,8 @@ AWS_ACCESS_KEY_ID=your_access_key_id
 AWS_SECRET_ACCESS_KEY=your_secret_access_key
 AWS_ENDPOINT=your_endpoint_url
 AWS_S3_BUCKET=your_bucket_name
+
+ANTHROPIC_VERSION="bedrock_api_version"
+ANTHROPIC_MODEL_ID="your_model_id"
+ANTHROPIC_MODEL_REGION="anthropic_region"
+ANTHROPIC_PROMPT_TEMPLATE="prompt_template_location"
```
stijnvermeeren-swisstopo, line 12: Please provide a better error message in case this env variable is not set.
Currently I get 
> API call failed for '10012.pdf, 7': unsupported operand type(s) for /: 'PosixPath' and 'NoneType'
which is quite confusing because
1. it's not actually the API call that fails, it fails already before that
2. there's nothing in this cryptic message that points me to the fact that an environment variable might be the cause

src/description_classification/main.py
letao: Fixed now

pyproject.toml
letao: Removed the redundant mention

.env.template
letao: @stijnvermeeren-swisstopo, I have addressed this by adding error message in the read_params() function in the  https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/0b404651159e8c37042de012af4543043dd6ebec/src/stratigraphy/util/util.py

This is now the message if an incorrect path is provided:

>   File "/Users/davis.ozols/Repos/swissgeol-boreholes-dataextraction/src/description_classification/classifiers/classifiers.py", line 281, in __init__
    self.classification_prompts = read_params(os.environ.get("ANTHROPIC_PROMPT_TEMPLATE"))
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/davis.ozols/Repos/swissgeol-boreholes-dataextraction/src/stratigraphy/util/util.py", line 118, in read_params
    raise FileNotFoundError(f"Provided parameter file not found: {config_path}")
FileNotFoundError: Provided parameter file not found: /Users/davis.ozols/Repos/swissgeol-boreholes-dataextraction/config/classification_prompts1.yml


This is the message if the variable does not point to a file:

>   File "/Users/davis.ozols/Repos/swissgeol-boreholes-dataextraction/src/description_classification/classifiers/classifiers.py", line 281, in __init__
    self.classification_prompts = read_params(os.environ.get("ANTHROPIC_PROMPT_TEMPLATE"))
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/davis.ozols/Repos/swissgeol-boreholes-dataextraction/src/stratigraphy/util/util.py", line 121, in read_params
    raise ValueError(f"Provided path does not point to a file: {config_path}")
    ValueError: Provided path does not point to a file: /Users/davis.ozols/Repos/swissgeol-boreholes-dataextraction/config
    
I think this should now indicate that we are pointing to an incorrect file for parameters.

stijnvermeeren-swisstopo (approved):
LGTM!

PR #176, opened 2025-04-11
Fix/issue 147/keywords by language

### LINKED ISSUE #147: Split all the keywords in the params file by language

TicaGit:
Resolves #147

Sort the keywords by language in the Yaml file. This is cleanier, and also avoids some false positives.

For example, false positives removed in geoquat/validation from A289 (Kappe not matched as nappe anymore),  1007 (GW not matched, as documents is in french), 2536 (GW not matched, as document is in french).

### DIFF
--- a/config/matching_params.yml
+++ b/config/matching_params.yml
@@ -200,7 +200,6 @@ groundwater_fp_keys:
     - Grundwasser-          
     - Grundwasserfassung   
     - GW/ # makes it possible to avoid false positives like "GW/" from the USCS Nomenclature columns     
-    - GW- # same 
 
 groundwater_unit_keys:
   de:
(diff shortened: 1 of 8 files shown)


stijnvermeeren-swisstopo (approved):
LGTM!

Nice to see that this is not just a refactoring, but actually improves the accuracy as well.

PR #183, opened 2025-04-16
move each classifier into separate files

TicaGit:
Small PR to move each classifier into separate files.


stijnvermeeren-swisstopo (changes requested):
Sorry to be pedantic, but it seems like we have the exact same issue with `"kunst"` vs. `null` with another field in the JSON as well, namely `"ground_truth_uscs_class"`.

stijnvermeeren-swisstopo (approved):
LGTM!

PR #184, opened 2025-04-17
Feat/issue 178/bert fine tuning

### LINKED ISSUE #178: USCS classification: fine-tuning of BERT-based classifier
Time-box: ca. 2 days.

Goal: get an idea of how much we can improve the accuracy by more fine-tuning, but not achieve the full optimization potential yet.

TicaGit:
resolves #178 

This PR introduces several minor enhancements to the BERT training pipeline, including the addition of QOC and improved pipeline parametrization.

The final best model utilizes a custom loss function with class balancing, although the impact on performance was minimal. The training process was divided into two stages:
 - First, the base model was trained for 32 epochs with the classifier and pooler layers unfrozen (learning rate = 3e-3, took about 7hours).
 - Then, the model from epoch 32 was fine-tuned with a smaller learning rate of 1e-4, unfreezing the last layer of the transformer encoder in addition to the classifier and pooler layers. This second stage also took 32 epochs (and about 7hours), but the best-performing model was achieved at epoch 20, which is the version stored in the S3 bucket.

The total inference time for the evaluation set of 2830 sample is 108s, so about 0.038s per sample. Final evaluation metrics are f1_micro=0.8272 and f1_macro=0.6666  (see comment below).

Additionally, the macro average calculation was slightly adjusted to align with scikit-learn’s implementation, which is likely more accurate as it excludes classes not present in the ground truth. This change led to a slight improvement in the results.

Those graphs show the overall training process:

training_weighted_loss
training_weighted_f1


stijnvermeeren-swisstopo (changes requested):
I did not review the training code in detail, nor did I try to reproduce the training. However; I did run the trained model on the validation set and my KPIs are equally good.

This is for me the first approach that achieves accuracies that would be usable in production (>80% F1 on all the larger classes).

However, we will need to discuss if this sufficiently outweighs the disadvantages of this approach (or if we can alleviate these disadvantages in some way), in particular:
- the lack of transparency ("black box")
- the need for a lot of training data (making the approach less usable for new languages / classes without much annotated examples)
- the inference speed / required computational resources.

TicaGit:
The definition of macro average has changed again. It now does not count class labels that don't appear in the whole data, not just the ground truth. This is how sklearn implements it.

TicaGit:
How to improve further?
- correct/investigate the ground truth for the classes OH, OL, GC, kA and SC (in decreasing order of importance).
- perform an in-depth hyper parameter tuning (could gain a few percentage points, but not more  I think).
- try out Domain adaptation as in the [GeoBertje paper](https://arxiv.org/pdf/2407.10991) (page 10)

stijnvermeeren-swisstopo (approved):
The change to the macro average computation makes sense.
I fixed one small typo in the README.
Now looks good to me!

PR #186, opened 2025-04-17
Refactor/issue 120/repo restructuring sidebar depth refactoring

TicaGit:
Reactoring to have a clear dependency direction.

Inside stratigraphy (showing only dependencies to other folder inside stratigraphy):
- base_sidebar_entry:  dependencies on
     - None
    
- interval: dependencies on:
   -  base_sidebar_entry

- sidebar_classes: dependencies on:
   -  interval,
   -  base_sidebar_entry

- sidebar_utils: dependencies on: 
   - sidebar_classes

- sidebar_extractor: dependencies on:
   -  sidebar_classes, 
   - interval, 
   - base_sidebar_entry, 
   - sidebar_utils

- layer : dependencies on:
   -  interval
   -  sidebar_classes


src/borehole_extraction/extraction/stratigraphy/interval/interval.py
```diff
@@ -19,8 +22,68 @@ def __init__(self, start: DepthColumnEntry | None, end: DepthColumnEntry | None)
...
+
+    @classmethod
+    def from_json(cls, data: dict) -> DepthInterval:
+        """Converts a dictionary to an object.
+
+        Args:
+            data (dict): A dictionary representing the layer depths.
+
+        Returns:
+            Interval: the corresponding LayerDepths object.
```
stijnvermeeren-swisstopo, line 55: If we do want to merge the `LayerDepths` and `Interval`/`DepthInterval` classes, then there are some comments like this one that should also be adjusted.

However, I'm not yet convinced that this is the correct idea. I think it was rather intentional to have two different classes here, one with more extraction-related logic (such as the attribute `has_decimal_point` on `DepthColumnEntry`) and one with more evaluation/visualization related logic (JSON serialization, `line_anchor`, `background_rect`). If we use the same class everywhere, then we will also (e.g. when we run the evaluation directly from a `predictions.json` file) have certain attributes such as `has_decimal_point` which not be set to the correct value, but will always have the default value, which I think is a rather dangerous situation. But I do also see that the classes are very similar. I guess it's a bit of a battle between two paradigms, "separation of concerns" vs. "don't repeat yourself".

@TicaGit @letao @lhaibach what do you think?

stijnvermeeren-swisstopo (changes requested):

src/borehole_extraction/extraction/stratigraphy/interval/interval.py
TicaGit: Hmm, yeah, I wasn’t completely sure about the change either. I noticed those two classes had very similar attributes, so I thought merging them might reduce confusion, especially since it can be tricky to decide which one to use. But I see your point, and I agree it could make sense to reintroduce the separation, perhaps with clearer names that better reflect their intended use.

stijnvermeeren-swisstopo (approved):
LGTM!

PR #187, opened 2025-04-22
Fix/issue 164/wrong elevation extraction

### LINKED ISSUE #164: Wrong elevation extraction
The changes for issue #151 introduced incorrect elevation extractions.

Happen's for the following PDFs:
  - 267123078-bp.pdf (Zurich)
  - 267123083-bp.pdf (Zurich)
  - 267125216-bp.pdf (Zurich)
  - A361.pdf (Geoquat validation)

The problem is that the elevation in the topmost layer is detected as the borehole's elevation due to its proximity to the keyword "Kote." The correct elevation is also detected, but during the assignment phase, the wrong one is chosen because it is closer to the borehole bounding box.

Image

For 267123083-bp.pdf (Zurich), also check why the elevation is still marked as correct even though it is extracted as 451.65 instead of 451.72—do we allow for rounding errors?

Possible Solutions:
- Don't allow a distance of 0, meaning that the elevation is contained in the bounding box (prevents misattributions but may exclude valid cases).

- Keep only the best elevation in the extraction phase (prioritizes stronger keywords but complicates cases with multiple boreholes).

- Prefer the highest elevation over the closest (avoids misassignment but may fail when multiple boreholes exist).

TicaGit:
resolves #164 

To be merged in branch of PR#185

This PR modifies the elevation matching logic. When only one borehole is detected, the highest elevation (based on the smallest y-coordinate on the page) is now chosen, instead of the closest to the borehole bounding box.

In most documents, this change has no effect, as the topmost layer elevation and the main elevation are typically the same (differing only in the bounding box). However, for some boreholes, the extraction is now corrected:

- Correct extraction: `267123078-bp.pdf`, `267123083-bp.pdf`, `267125216-bp.pdf` (Zurich), and `3425.pdf` (Geoquat validation)
- Bonus fixes: Proper extraction in `267124282.pdf` (Zurich) and `A361.pdf` (Geoquat validation), where OK Terrain is now extracted instead of OK Rohr

Additionally, elevation entries that are also identified as groundwater are now removed to prevent incorrect elevation extraction. This change is important in the following cases:

- `5115.pdf` and `5111.pdf` (Geoquat validation): The groundwater entries were incorrectly extracted instead of the real elevations, as they appeared above the actual elevations on the page.
- `A7068.pdf`, `A7069.pdf`, and `A361.pdf`: These files are also affected, but the result does not change since the groundwater entries are below the real elevations.


src/borehole_extraction/extraction/predictions/predictions.py
```diff
@@ -153,93 +152,91 @@ def remove_elevations_overlaping_groundwater(self) -> None:
...
+            # map all elements to their closest borehole.
+            borehole_to_many_elements_mapping = self._many_to_one_match_element_to_borehole(
+                element_list, set(borehole_index_to_matched_elem_index.keys())
+            )
+            for borehole_index, element_list_index in borehole_to_many_elements_mapping.items():
+                assert borehole_index not in borehole_index_to_matched_elem_index
+                # only allow to pick from the elements that are not matched yet
+                available_elements_index = [elem for elem in element_list_index if elem not in picked_element_index]
+                if not available_elements_index:
+                    continue
```
stijnvermeeren-swisstopo, line 231: Wouldn't it be more efficient to already restrict to the elements that are still available inside the call to `_many_to_one_match_element_to_borehole`, similar to how we already restrict the considered boreholes?

src/borehole_extraction/extraction/predictions/predictions.py
```diff
@@ -84,6 +82,7 @@ def build(self) -> list[BoreholePredictions]:
 
         # only criterion for the number of boreholes in the document is the number of borehole layers identified.
         self._num_boreholes = len(self._layers_with_bb_in_document.boreholes_layers_with_bb)
+        # Duplicate metadata (e.g. elevation, coordinates) if fewer values were extracted than number of boreholes
         self._extend_lenght_metadata_elements()
```
stijnvermeeren-swisstopo, line 86: Maybe it makes sense to do this at the start of the method `_one_to_one_match_element_to_borehole`? Because if we forget to extend these lists, and call `_one_to_one_match_element_to_borehole` with fewer elements than boreholes, then we will most likely end up in an infinite loop, which is rather dangerous.
If we extend the element list inside `_one_to_one_match_element_to_borehole`, then we won't be able to directly modify the values of `self._elevations_list` and `self._coordinates_list` anymore, and consequently the map returned by `_one_to_one_match_element_to_borehole` will need to contain the actual element objects instead of their indices, but that might even be a good thing.

There's also a typo in the method name by the way, `lenght` -> `length`.

stijnvermeeren-swisstopo (changes requested):
I'm very happy with the method, I just have two suggestions to improve the way it is coded a little bit.

src/borehole_extraction/extraction/predictions/predictions.py
TicaGit: yes, now the dict is mapping the borehole index to the element directly.

stijnvermeeren-swisstopo (approved):
Nice, that's exactly what I wanted :). LGTM!

PR #188, opened 2025-04-25
check for intersections between material descriptions and sidebars

### LINKED ISSUE #130: Dont allow MaterialDescriptionRectWithSidebar to overlap
Currently, in extract.py (process_page()), we remove material description-sidebar pairs (material_descriptions_sidebar_pairs) if their material description bounding box (material_description_rect) intersect with those of another pair. However, we do not check whether the sidebar of one pair intersects with the material description rect of another pair.

We should modify the logic to also prevent sidebars from overlapping with other material description rects to ensure separation. 
Example of where we recieve 2 MaterialDescriptionRectWithSidebars for deepwell 3EK5LEG800_bp_19940601_Arsch93-18.pdf

Image

### LINKED ISSUE #153: Merge MaterialDescriptionRectWithSidebar objects when they belong to the same borehole.
**Descpription**
In extract.py, the process_page function scans the document to extract multiple borehole profiles. However, the same borehole may be detected twice using different reference points. For example, in 13080.pdf, one detection occurs via AaboveBsidebar (depth column) and another via LayerIdentifiersidebar.

Image

The issue likely originates from the following lines:
```        
filtered_pairs = [
    item for index, item in enumerate(material_descriptions_sidebar_pairs) if index not in to_delete
]
```
At this point in the code, filtered_pairs should contain only distinct boreholes. However, in the case of 13080.pdf, it includes duplicate elements:

Image

Ideally, we would merge the two sidebars into one, incorporating as much information as possible from both "source" identifiers (the depth column and the layer identifier). However, this would be more complex, as it would also require additional modifications, such as enabling the display of both pieces of information in the output PDF.

**List of pdfs (non-exhaustive)**
GEOQUAT: `('14225.pdf', 2), ('8366.pdf', 2), ('5872.pdf', 2), ('9394.pdf', 2), ('A1304.pdf', 2), ('A263.pdf', 2), ('A11406.pdf', 2), ('1488.pdf', 2), ('14273.pdf', 2), ('13080.pdf', 2), ('7703.pdf', 2), ('7317.pdf', 2)`
ZURICH: `('684252058-bp.pdf', 2), ('268124125-bp.pdf', 2)`
The number next to the name of the document is the number of boreholes detected, where all of the pdfs only contains one borehole.

**Potential solution idea**
Here is an initial code snippet that I had in mind.

```
def _group_similar_pairs(self, filtered_pairs: list[MaterialDescriptionRectWithSidebar]) -> list[BoreholeLayers]:
    """Groups similar pairs of material description rect with sidebar.

    This is done because the informations about the same borehole could be represented in multiple ways.
    We only want to keep multiple pairs if they belong to different boreholes in the given pdf page.

    Args:
        filtered_pairs (list[MaterialDescriptionRectWithSidebar]): The filtered pairs.

    Returns:
        list[BoreholesLayers]: A list containing individual boreholes with their respective layers.
    """
    treshold = 0.5  # TODO put as a parameter

    intervals_blocks = [self._get_interval_block_pairs(pair) for pair in filtered_pairs]
    depths_per_pairs = []
    for i, pairs in enumerate(intervals_blocks):
        depths_per_pairs.append(set())
        for pair in pairs:
            if not pair.depth_interval:
                continue
            if pair.depth_interval.start:
                depths_per_pairs[i].add(pair.depth_interval.start.value)
            if pair.depth_interval.end:
                depths_per_pairs[i].add(pair.depth_interval.end.value)
    to_merge = []
    for i, depth in enumerate(depths_per_pairs):
        for j, other_depth in enumerate(depths_per_pairs[i + 1 :]):
            similarity_score = max(
                len(depth & other_depth) / len(depth), len(depth & other_depth) / len(other_depth)
            )
            if similarity_score > treshold:
                to_merge.append((i, i + j + 1))

    output: list[BoreholeLayers] = []

    for i, pair in enumerate(filtered_pairs):
        pair = pair
        if i not in to_merge:
            output.append(BoreholeLayers(intervals_blocks[i]))
        else:
            # use new class ??
            # Interval is
            # one Interval, one TextBlock
            # list of IntervalBlockPair
            # BoreholesLayers
            # bounding_boxes = [
            #     BoundingBoxes.from_material_description_rect_with_sidebar(pair, self.page_number)
            #     for pair in filtered_pairs
            # ]
            pass

    return
```

TicaGit:
## This PR resolves #130 

This issue was not reproducible anymore, the overlapping inside deepwell 3EK5LEG800_bp_19940601_Arsch93-18.pdf was already resolved by PR#174, and no similar problems existed in the datasets. I implemented a quick solution anyway, as this problem could appear in some other pdf unseen yet.

## This PR also resolves #153 

Some additional checks are applied to filter out material descriptions with sidebars before validating them as actual boreholes.

The first check removes any pair that contains no depth information (i.e.all depth values are None). This affects the following files:

GeoQuat
- False positives removed (good): 14225.pdf, A11406.pdf, 1488.pdf, 14273.pdf
- True positive removed (bad): A1304.pdf: According to the ground truth, this file contains three boreholes. One is correctly detected. The second borehole is removed at this stage because its depth column was not detected during extraction, leaving it without any depth values (see image below). The third borehole is not detected at all.

Zurich
- False positives removed (good): 695265010-bp.pdf : A column of incorrect numbers is removed. The main borehole isn’t detected in this file, as it is rotated).

Overall, this check seems justified. The only true positive removed is due to an error in the extraction pipeline, which fails to detect a depth column, rather than a flaw in the check itself.

---

The second check finds duplicates in the pairs. It affects the following files:
GeoQuat
- False positives removed (good): 8366.pdf, 13080.pdf, 7317.pdf

This check is only beneficial.

---

Among the list provided in issue #153, the following document remains false positives and here the reason why:
- 5872.pdf: thickness of layers column is detected instead of depth.
- A263.pdf : a series of explanation containing diameter numbers is detected as Sidebar (because it contains 1), ...5) ).
- 7703.pdf: compactness column is detected as a borehole.
- 684252058-bp.pdf: water infiltration is detected instead of depth, this breaks the borehole into 2 incorrect parts.
- 268124125-bp.pdf: humidity column detected

All those issue are related to incorrect depth column extraction and not to duplication, so issue #153 is effectively resolved.

Also note that:
- 9394.pdf was already resolved by  PR#174

image

### DIFF
--- a/src/extraction/features/extract.py
+++ b/src/extraction/features/extract.py
@@ -97,7 +97,7 @@ class MaterialDescriptionRectWithSidebarExtractor:
             item for index, item in enumerate(material_descriptions_sidebar_pairs) if index not in to_delete
         ]
 
-        # remoove pairs that have all depths equal to None (only if there is more than one pair).
+        # remove pairs that have all depths equal to None (only if there is more than one pair).
         to_delete = self._find_no_depths_indices(filtered_pairs)
         filtered_pairs = [item for index, item in enumerate(filtered_pairs) if index not in to_delete]
 
--- a/src/extraction/features/extract.py
+++ b/src/extraction/features/extract.py
@@ -91,13 +91,7 @@ class MaterialDescriptionRectWithSidebarExtractor:
             pair for pair in material_descriptions_sidebar_pairs if pair.score_match >= 0
         ]
 
-        to_delete = []
-        for i, pair in enumerate(material_descriptions_sidebar_pairs):
-            if any(
-                pair.material_description_rect.intersects(other_pair.material_description_rect)
-                for other_pair in material_descriptions_sidebar_pairs[i + 1 :]
-            ):
-                to_delete.append(i)
+        to_delete = self.find_intersecting_indices(material_descriptions_sidebar_pairs)
 
         filtered_pairs = [
             item for index, item in enumerate(material_descriptions_sidebar_pairs) if index not in to_delete
@@ -110,6 +104,46 @@ class MaterialDescriptionRectWithSidebarExtractor:
             self._create_borehole_from_pair(pair)
             for pair in sorted(filtered_pairs, key=lambda pair: pair.score_match, reverse=True)
         ]
+
+    def find_intersecting_indices(
+        self, material_descriptions_sidebar_pairs: list[MaterialDescriptionRectWithSidebar]
+    ) -> list[int]:
+        """Identifies overlapping material descriptions or sidebars.
+
+        This function scans through all material description/sidebar pairs and returns a list of indices
+        that should be removed due to overlaps. If an intersection is found between two elements, only
+        the first (lower-indexed) element is marked for deletion.
+
+        Args:
+            material_descriptions_sidebar_pairs (list[MaterialDescriptionRectWithSidebar]): a list of pairs consisting
+                of a material description rectangle and an optional sidebar.
+
+        Returns:
+            list[int]: The indices of elements that overlap with others and should be removed.
+        """
+        to_delete = []
+
+        def intersects_any(rect, others):
+            return any(rect.intersects(other) for other in others if other is not None)
+
+        for i, pair in enumerate(material_descriptions_sidebar_pairs):
+            mat_rect = pair.material_description_rect
+            sidebar_rect = pair.sidebar.rect() if pair.sidebar else None
+
+            # Build the list of other rectangles
+            remaining_pairs = material_descriptions_sidebar_pairs[i + 1 :]
+            other_mat_rects = [p.material_description_rect for p in remaining_pairs]
+            other_sidebar_rects = [p.sidebar.rect() if p.sidebar else None for p in remaining_pairs]
+
+            # Check all conditions
+            if (
+                intersects_any(mat_rect, other_mat_rects)
+                or intersects_any(mat_rect, other_sidebar_rects)
+                or (sidebar_rect and intersects_any(sidebar_rect, other_mat_rects))
+                or (sidebar_rect and intersects_any(sidebar_rect, other_sidebar_rects))
+            ):
+                to_delete.append(i)
+        return to_delete
 
     def _create_borehole_from_pair(self, pair: MaterialDescriptionRectWithSidebar) -> ExtractedBorehole:
         bounding_boxes = PageBoundingBoxes.from_material_description_rect_with_sidebar(pair, self.page_number)
--- a/src/extraction/features/extract.py
+++ b/src/extraction/features/extract.py
@@ -97,8 +97,7 @@ class MaterialDescriptionRectWithSidebarExtractor:
             item for index, item in enumerate(material_descriptions_sidebar_pairs) if index not in to_delete
         ]
 
-        # remove pairs that have all depths equal to None (only if they is more than one pair).
-        to_delete = self._find_no_depths_indices(filtered_pairs)
+        # remove pairs that have all depths equal to None (only if there is more than one pair).
         filtered_pairs = [item for index, item in enumerate(filtered_pairs) if index not in to_delete]
 
         # remove pairs that are likely duplicates of others.
--- a/src/extraction/features/extract.py
+++ b/src/extraction/features/extract.py
@@ -200,7 +200,7 @@ class MaterialDescriptionRectWithSidebarExtractor:
             [interval.depth_interval for interval in self._get_interval_block_pairs(pair)] for pair in filtered_pairs
         ]
 
-        # create sets with all the depths appeating in the interval list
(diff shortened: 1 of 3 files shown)


src/extraction/features/extract.py
```diff
@@ -91,26 +91,141 @@ def process_page(self) -> list[ExtractedBorehole]:
...
+        to_delete = self._find_intersecting_indices(material_descriptions_sidebar_pairs)
         filtered_pairs = [
             item for index, item in enumerate(material_descriptions_sidebar_pairs) if index not in to_delete
         ]
 
+        # remove pairs that have all depths equal to None (only if there is more than one pair).
+        filtered_pairs = [item for index, item in enumerate(filtered_pairs) if index not in to_delete]
+
+        # remove pairs that are likely duplicates of others.
+        non_duplicated_pairs = [item for index, item in enumerate(filtered_pairs) if index not in to_delete]
```
stijnvermeeren-swisstopo, line 104: Something is very wrong here. We filter out the `to_delete` indices three times, but each time from a different, smaller list. This cannot be on purpose...

src/extraction/features/extract.py
```diff
@@ -91,26 +91,143 @@ def process_page(self) -> list[ExtractedBorehole]:
...
-                to_delete.append(i)
-
+        # remove pairs that have any of their elements (sidebar, material description) intersecting with others.
+        to_delete = self._find_intersecting_indices(material_descriptions_sidebar_pairs)
         filtered_pairs = [
             item for index, item in enumerate(material_descriptions_sidebar_pairs) if index not in to_delete
         ]
 
+        # remoove pairs that have all depths equal to None (only if there is more than one pair).
+        to_delete = self._find_no_depths_indices(filtered_pairs)
```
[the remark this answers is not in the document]
TicaGit: Here, it tries to correct a typo but actually deletes the line below too.

src/extraction/features/extract.py
```diff
@@ -91,26 +91,143 @@ def process_page(self) -> list[ExtractedBorehole]:
...
+        if len(filtered_pairs) <= 1:
+            # only one pair, can't be a duplicate
+            return []
+
+        all_interval_lists = [
+            [interval.depth_interval for interval in self._get_interval_block_pairs(pair)] for pair in filtered_pairs
         ]
 
+        # create sets with all the depths appearing in the interval list
+        all_depths_set = []
```
stijnvermeeren-swisstopo, line 204: The variable name suggests that this is a set, whereas it is actually a list of sets. Maybe something like `all_depth_sets` or `list_of_depth_sets` would be clearer?

src/extraction/features/extract.py
```diff
@@ -91,26 +91,143 @@ def process_page(self) -> list[ExtractedBorehole]:
...
+
+        Returns:
+            list[int]: Indices of pairs where all depth intervals are None.
+        """
+        # if only one pair is detected, we should keep it
+        if len(filtered_pairs) <= 1:
+            return []
+
+        all_interval_lists = [
+            [interval.depth_interval for interval in self._get_interval_block_pairs(pair)] for pair in filtered_pairs
```
stijnvermeeren-swisstopo, line 171: This filters out both pairs that don't have any sidebar, as well as "indicator" sidebars. In some cases however, I think it is worthwhile to keep the indicator sidebars. One example is 1488.pdf, which you list as a "false positive removed", but I would actually argue that this change is not so good, because even thought the F1 score improves, we actually lose some valuable descriptions at the top of the borehole profile.

If I change/simplify the condition to 
```
        to_delete = [
            i
            for i, pair in enumerate(filtered_pairs)
            if not pair.sidebar
        ]
```
then we keep these descriptions. However, it also reverts the improvements for e.g. 14273.pdf. Maybe if we do the duplicate check not just for depth values, but also for the indicator values, then we might get the best of both approaches? Could you try that out?

stijnvermeeren-swisstopo (changes requested):

TicaGit:
Now neither of the two boreholes in 1488.pdf are deleted, but all of the other mentioned files are filtered I believe.

stijnvermeeren-swisstopo (approved):
I like the new behaviour, it's exactly what I intended with my previous comment. As you said, the implementation is not entirely satisfactory yet. There's a lot of code that only covers a small number of known input PDFs, and there's a duplicated call to the `_get_interval_block_pairs` method, which is somewhat confusing and inefficient. However, when thinking about how to solve it in a better way, I did not see an easy fix, but rather I think it will involve some more fundamental changes to how we find/identify boreholes in the first place, and that's certainly out-of-scope for this PR. For that reason, I created a new issue https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/195 and I'm happy to merge this as a temporary solution in the meantime.

Issue #180, opened 2025-04-14
USCS classification: Claude-based classifier: optimize class descriptions in system prompt


stijnvermeeren-swisstopo:
Time-box: 3-4 days

Compared to the baseline classifier, the prompt for Claude could:
- do without extensively listing different synonyms/variations
- benefit from a more verbose description, instead of keywords that are expected to be matched exactly

Potentially reach out to Stefan Volken in case we need help with phrasing the class descriptions for the prompt.

Additional question to look into:
- Can we take advantage of the multilingual capabilities of the language model, by e.g. also using the (more extensively optimized) descriptions in German in the system prompt for material descriptions in English.

PR #190, opened 2025-04-29
Feat/issue 180/claude based classifier optimizations

letao:
Addresses #180 and helps in AWS Bedrock classifier optimisation

Implements small changes to AWS Bedrock classifier:

- Aibility to use classification parameters yml as an environment variable
- Simple versioning or prompts and classification parameters for MLflow (currently for this to work files require _v suffix)
- New prompts for Zero Shot and Reasoning (Chain of Thought).
- Implements option to set delay between API calls to avoid timeouts due to request quota limit.
- Storage of LLM reasoning pattern (if used) in the prediction file


src/classification/classifiers/aws_bedrock_classifier.py
```diff
@@ -38,18 +39,20 @@ def __init__(
...
             store_files (bool): Whether to store the prediction files (default: False)
             max_concurrent_calls (int): Maximum number of concurrent API calls (default: 1)
+            api_call_delay (float): Delay between API calls in seconds (default: 0)
         """
         self.bedrock_client = boto3.client(service_name="bedrock-runtime", region_name=os.environ.get("AWS_REGION"))
         self.anthropic_version = os.environ.get("ANTHROPIC_VERSION")
         self.model_id = os.environ.get("ANTHROPIC_MODEL_ID")
 
-        self.uscs_patterns = classification_params["uscs_patterns"]
+        self.uscs_patterns = read_params(os.environ.get("ANTHROPIC_CLASSIFICATION_PARAMS"))["uscs_patterns"]
```
TicaGit, line 48: Do we still need the classification_params import at line 17? Maybe it would be easier to follow if `read_params(os.environ.get("ANTHROPIC_CLASSIFICATION_PARAMS")` would be done at the top of the file (l.17)

src/classification/main.py
```diff
@@ -65,15 +66,24 @@ def log_ml_flow_infos(
...
+    # Log model and id, prompt and parameter versions if anthropic model used
     if isinstance(classifier, AWSBedrockClassifier):
         mlflow.log_param("anthropic_model_id", os.environ.get("ANTHROPIC_MODEL_ID"))
 
+        prompt_version_match = re.search(r"(v\d+)", os.environ.get("ANTHROPIC_PROMPT_TEMPLATE"))
+        if prompt_version_match:
+            mlflow.log_param("anthropic_prompt_version", prompt_version_match.group(1))
+
+        class_param_version_match = re.search(r"(v\d+)", os.environ.get("ANTHROPIC_CLASSIFICATION_PARAMS"))
+        if class_param_version_match:
```
TicaGit, line 84: Same here, I feel like using `os.environ.get("ANTHROPIC_CLASSIFICATION_PARAMS")` in the code clashes a bit with all the `classification_params = read_params("classification_params.yml")` that are done on top on most files (including this file, main.py).
It is confusing to know which parameters to use where, especially as they are really similar.

TicaGit (changes requested):
Few comments regarding the classification_params file

stijnvermeeren-swisstopo:
I think I agree with @TicaGit , I would probably keep a single file with the "current preferred version" of the params / prompt, and for every relevant version that we want to potentially refer back to later, we can create a commit in Git, then all the important versions are in the git history and can be referred to by their commit hash. What do you think of that idea?

I tried out defining the classes in the system prompt in a more human-readable way, like this:
> ...
> - The class 'CL' is usually be described by phrases such as 'siltiger Ton', 'Lehm, Feinsand'
> - The class 'CL-ML' is usually be described by phrases such as 'toniger Silt', 'tonig-sandiger Silt'
> - The class 'CM' is usually be described by phrases such as 'magerer Ton'
> ...
This did not significantly change the accuracy.

I also wondered what would happen if we only listed the possible classes (e.g. "CL, CL-ML, CM, ...") without proving any example descriptions. Would the model be able to fall back to some build-in knowledge about the USCS classes and not really need the descriptions from our prompt at all. However, the accuracy dropped significantly (no exact numbers; I only ran it on a small sample). So the way we define the possible classes is certainly important for the accuracy. However, it is not immediately clear to me how we can improve the prompt to get significant accuracy gains. Let's discuss tomorrow in the sprint planning what our next steps should be.

src/classification/main.py
letao: This should now be addressed with the new approach using config parameters.

letao:
@stijnvermeeren-swisstopo  and @TicaGit I have refactored the usage of the parameters. Now we can use the bedrock_config.yml file to set the necessary parameters for the classification:
- temperature
- max output tokens 
- prompt version
- class label version

The prompts and class labels are defined inside the classification_patterns_bedrock.yml (for labels) and classification_prompts.yml for prompts. Versioning of prompts and labels can be achieved by adding a specific version value inside the yml file, e.g. v1 which then is used in the bedrock_config.yml file

bedrock_versioning_1


bedrock_versioning_2

stijnvermeeren-swisstopo (approved):

src/classification/classifiers/aws_bedrock_classifier.py
```diff
@@ -14,43 +14,46 @@
...
         self.model_id = os.environ.get("ANTHROPIC_MODEL_ID")
 
-        self.uscs_patterns = read_params(os.environ.get("ANTHROPIC_CLASSIFICATION_PARAMS"))["uscs_patterns"]
-        self.classification_prompts = read_params(os.environ.get("ANTHROPIC_PROMPT_TEMPLATE"))
+        # Bedrock parameters
+        self.uscs_pattern_version = read_params("bedrock/bedrock_config.yml")["uscs_pattern_version"]
+        self.prompt_version = read_params("bedrock/bedrock_config.yml")["prompt_version"]
+        self.temperature = read_params("bedrock/bedrock_config.yml")["temperature"]
+        self.max_tokens = read_params("bedrock/bedrock_config.yml")["max_tokens"]
+        reasoning_mode = read_params("bedrock/bedrock_config.yml")["reasoning_mode"]
```
TicaGit, line 41: I think there is redundant calls to `read_params`, we can do `config = read_params("bedrock/bedrock_config.yml")` and then just access the parameters with `config["prompt_version"]`, ...

same later for `patterns = read_params("bedrock/classification_patterns_bedrock.yml")` and `prompts = ...`

TicaGit (approved):
slight change requested, but overall LGTM!

PR #196, opened 2025-04-30
Feat/issue 181/lithology lexic

### LINKED ISSUE #181: Lithology classification: load and parse Lexic dataset
- Load the RDF dataset from https://github.com/swisstopo/swissgeol-lexic-vocabulary-lithologie.
- Create the appropriate data structure in Python to represent the dataset, so that it will be easy to make use of it in a future lithology classifier.

TicaGit:
Resolves #181: Add Lithology Vocabulary Script and Adapt Classification Pipeline

This PR introduces a script to load the RDF lithology lexicon vocabulary. It generates both a YAML configuration file and a CSV file.

Additionally, this PR adapts the existing classification pipeline to support lithology classification. It also implements a dummy classifier for lithology classification.

This PR does not modify the other classifiers to fully support lithology yet, that work will be handled in separate sub-issues. The only changes made to the existing classifiers are minimal adjustments to ensure continued compatibility with the USCS classification.


src/scripts/load_rdf_to_csv.py
```diff
@@ -0,0 +1,163 @@
...
+    Returns:
+        str: returns the last part of the URL
+    """
+    if isinstance(url, str):
+        return url.split("/")[-1]
+    return url
+
+
+def count_capital_terms(id_value: str) -> int:
+    """Count the number of capital terms in a given string.
```
stijnvermeeren-swisstopo, line 28: What exactly is meant by "capital term" here. Could you give some examples?

src/classification/utils/data_loader.py
```diff
@@ -25,17 +25,19 @@ class LayerInformations:
     layer_index: int
     language: str
     material_description: str
-    ground_truth_uscs_class: None | USCSClasses
-    prediction_uscs_class: None | USCSClasses  # dynamically set
+    data_type: str  # a layer is either classified into USCS or lithology, but never both
```
stijnvermeeren-swisstopo, line 28: maybe it's safer to make this an Enum with a fixed range of possible values, instead of a string that could potentially be anything?

Or maybe even an actual class, where behaviour like
```
    class_str = class_str.lower()
    normalized_str = class_str.replace("-", "_") if data_type == "uscs" else class_str.split(",")[0]
```
can be directly implemented, e.g.

```
class ClassificationType:
    def normalize_label(self, class_str: str) -> str:
        ...

    @prop
    def enum(self) -> ClassEnum:
        ...

    @prop
    def value_for_dummy_classifier(self):
        ...
```
with two instances, one of USCS and one for Lithology. Than can then be more easily extended later when we start classifying even more properties.

src/classification/utils/classification_classes.py
```diff
@@ -0,0 +1,164 @@
...
+        class_str (str): The input string to map.
+        data_type (str): The data_type to search in (uscs or lithology).
+
+    Returns:
+        ClassEnum: The matching enum member, or `kA` if no match is found.
+    """
+    class_str = class_str.lower()
+    normalized_str = class_str.replace("-", "_") if data_type == "uscs" else class_str.split(",")[0]
+
+    classes_enum = USCSClasses if data_type == "uscs" else LithologyClasses
```
stijnvermeeren-swisstopo, line 159: These if-else constructs are not very easy to follow, and it will get much messier once we have more than two possible values. In my comment to data_loader.py I have a suggestion for an alternative approach that could get rid of these if-elses.

src/classification/classifiers/dummy_classifier.py
```diff
@@ -1,23 +1,25 @@
...
-            layer_descriptions (list[LayerInformations]): The LayerInformations object
-
+            layer_descriptions (list[LayerInformations]): List of layer information objects to classify.
         """
         for layer in layer_descriptions:
-            layer.prediction_uscs_class = USCSClasses.CL_ML
+            if layer.data_type == "uscs":
+                layer.prediction_class = USCSClasses.CL_ML
+            else:
+                layer.prediction_class = LithologyClasses.Marlstone
```
stijnvermeeren-swisstopo, line 25: try to avoid such an if-else, see other comments

src/classification/main.py
```diff
@@ -135,6 +135,12 @@ def common_options(f):
         default=None,
         help="Path to the local trained model.",
     )(f)
+    f = click.option(
+        "--data-type",
```
stijnvermeeren-swisstopo, line 139: - Add this to the README
- Log the value as a param to MLFlow

stijnvermeeren-swisstopo (changes requested):

src/classification/evaluation/evaluate.py
```diff
@@ -24,15 +24,15 @@ class AllClassificationMetrics:
...
 
     Attributes:
-        global_metrics (dict[ClassEnum, Metrics]): A dictionary containing the classification metrics
-            for each of the classes at a global level.
-        language_metrics (dict[str, dict[ClassEnum, Metrics]]): A dictionary where each key represents
-            a supported language. Each value is another dictionary containing the classification metrics
+        global_metrics (dict[ClassificationType.EnumClasses, Metrics]): A dictionary containing the
+        classification metrics for each of the classes at a global level.
+        language_metrics (dict[str, dict[ClassificationType.EnumClasses, Metrics]]): A dictionary where each key
+            representsa supported language. Each value is another dictionary containing the classification metrics
```
stijnvermeeren-swisstopo, line 30: typo

README.md
```diff
@@ -183,6 +183,7 @@ boreholes-classify-descriptions -f data/geoquat_ground_truth.json -s data/geoqua
 - You can optionally provide a folder containing a subset of files by using the `-s` or `--file-subset-directory` flag. Only the filenames in this folder will be loaded from the JSON file for classification.
 - Use the `-c` or `--classifier` option to choose the classifier type from `dummy`, `baseline` or `bert`.
 - If you are using the classifier `bert` with a trained local [model](#train-bert-model), specify its folder path using the `-p` or `--model-path` flag. The folder has to contain all files generated when saving a model checkpoint with the transformers library.
+- Use `--class-sys` to specify the system you want to classify the description into. Currently supports `uscs` and `lithology`.
```
stijnvermeeren-swisstopo, line 186: I'm not a huge fan of abbreviating classification system to "class-sys", to me it's not immediately obvious what "class sys" stands for. But we can leave it for now, it's not a big deal.

src/classification/main.py
```diff
@@ -65,6 +66,9 @@ def log_ml_flow_infos(
     for class_, count in get_data_class_count(layer_descriptions).items():
         mlflow.log_param(f"class_{class_}_count", count)
 
+    # Log the classification systems used
+    mlflow.log_param("classification_systems", get_classification_systems(layer_descriptions))
```
stijnvermeeren-swisstopo, line 70: Wouldn't it be more straight-forward and efficient to read this directly from the script parameters, rather than reconstructing it from the classified layer descriptions?

src/scripts/load_rdf_to_csv.py
```diff
@@ -0,0 +1,167 @@
...
+        - Marlstone has 1 uppercased letter (so will be of order 1)
+        - MarlstoneBioclasts has 2 uppercased letter (so will be of order 2)
+
+    Args:
+        id_value (str): The string to count capital letters in.
+
+    Returns:
+        int: The number of capital terms in the string.
+    """
+    terms = re.findall(r"[A-Z][a-z0-9]*", id_value)
```
stijnvermeeren-swisstopo, line 40: Isn't the "[a-z0-9]*" completely redundant here? I mean, it would be relevant if we were interested in the actual terms, but in this case, we are only interested in the count?

stijnvermeeren-swisstopo (changes requested):
A few more small comments.

src/classification/main.py
TicaGit: I did it this way because I think we could want to classify the layer descriptions into different categories in the future. For example, some layers could be classified into uscs, and other (deeper) layers into the lithology system.
I would leave it like that, but I could change it if you think it is better.

stijnvermeeren-swisstopo: True, but I don't really see how that is linked to the question of whether we log this from the script parameters, or from the classified layer descriptions. Also when we read the classification systems from the parameters, this does not stop us from extending this later to multiple classification systems for a single run?

Right now, the most critical point is that for experiments logged to MLFlow, we must be able to understand exactly what was measured (in particular: USCS or lithology classification). I believe that this is best read directly from the parameters.

src/classification/utils/data_utils.py
```diff
@@ -246,15 +260,70 @@ def write_overview(metrics_dict: dict[str, float], layers_with_predictions: list
...
+        stats_first_key, samples_for_first_key, total_first_key = build_class_stats(
+            samples_for_first_key, first_key_class, key_class_config
+        )
+
+        # get the metric for the correct first key (recall if "zoom" is on ground truth, precision otherwise)
+        first_key_metric_micro = next(
+            (
+                v
+                for k, v in metrics_dict.items()
+                if key_class_config.metric_used in k and first_key_class.name == "_".join(k.split("_")[1:-1])
```
stijnvermeeren-swisstopo, line 298: `key_class_config.metric_used in k` is just a substring match, right? What happens is we have a class label that contains the string "recall" or "precision" as a substring? I would prefer a condition here that is guaranteed to work for all possible class labels.

src/classification/main.py
```diff
@@ -172,6 +192,7 @@ def main(
     file_subset_directory: Path,
     classifier_type: str,
     model_path: Path,
+    classification_system: str,
```
stijnvermeeren-swisstopo, line 195: I think that we should check that the classification system and the classifier type are compatible.
Right now, the script can be started as `boreholes-classify-descriptions -f data/geoquat_ground_truth.json -s data/geoquat/validation -c baseline -cs lithology` and then it actually applies USCS classification instead of lithology classification, and eventually runs into a cryptic "StopIteration" error on the following lines, because the classes are not the correct ones:
```
        first_key_metric_micro = next(
            (
                v
                for k, v in metrics_dict.items()
                if key_class_config.metric_used in k and first_key_class.name == "_".join(k.split("_")[1:-1])
            )
        )
```

stijnvermeeren-swisstopo (changes requested):
I pushed a commit with two small fixes, and still have a few other comments remaining.

src/classification/utils/data_utils.py
TicaGit: the metrics_dict look like that:
image

I changed the condition to `k.endswith(key_class_config.metric_used)` which should be enough + I don't think we will ever have a class named "recall" or "prec".
I could refactor the metrics_dict if you want to have a multilevel dict here, but this will take more time.

src/classification/main.py
TicaGit: That was the point of this paragraph in the PR description:
`This PR does not modify the other classifiers to fully support lithology yet, that work will be handled in separate sub-issues. The only changes made to the existing classifiers are minimal adjustments to ensure continued compatibility with the USCS classification.`
It says that lithology is not supported yet with this PR and other classifiers... I will add an explicit check at the start of the pipeline to raise a NotImplementedError.

stijnvermeeren-swisstopo: Yes, that's good enough. I just added a explanatory message to the NotImplementedError.

src/classification/utils/data_utils.py
stijnvermeeren-swisstopo: Indeed, it might be worthwhile to make this dict more structured at a later point, but this is good enough for now.

stijnvermeeren-swisstopo (approved):
I've pushed a small commit, removing an unused method and adding an error message. The PR is good to merge now as far as I'm concerned.

letao (approved):

PR #197, opened 2025-05-01
handle missing values in ground_truth layers.

lhaibach:
Ground truth data now can have NoneType as material description in contrast to solely string as previously. Thus the small adjustment in the loading of the ground truth is needed to handle NoneType inputs in Layers.


stijnvermeeren-swisstopo (changes requested):
The `boreholes-classify-descriptions` script needs a fix as well.

stijnvermeeren-swisstopo (approved):

Issue #195, opened 2025-04-30
More reliable identification of borehole profiles


stijnvermeeren-swisstopo:
In https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/188 we've introduced several "duplicate checks" to avoid detecting false-positive boreholes in the PDFs. Such false positives can arise when other information on the PDF page is incorrectly interpreted as a borehole profile with depths and/or material descriptions, or when data from the same borehole is detected as multiple separate borehole profiles, instead of a single one.

The existing solution is not very satisfactory for several reasons:
- There are several separate duplicate checks, each of which only covers a small number of known cases. It seems like a lot of complexity for a small gain in accuracy.
- For some boreholes, we now do multiple calls to the method `_get_interval_block_pairs`. This is inefficient and potentially confusing.

I think that we should try to tackle these problems already a little earlier on in the pipeline. Instead of identifying and removing duplicated, I think it would be better to try to not identify these false positives as separate boreholes in the first place.

On idea on how to do that, would be to make use of the line detection results to try to identify table-like layouts on the PDF page (rectangular areas with contiguous horizontal and vertical lines). Not every borehole profile is represented in such a table-like structure, but if it is, then each table-structure will only contain at most one borehole. Enforcing this will probably already get rid of most duplicates / false positives from the start, potentially making some of the duplicate checks from https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/188 redundant. Moreover, identifying these table structures could be useful for other purposes, such as more accurately identifying columns of depths and material descriptions, or more accurately associating each borehole with the correct metadata and groundwater data.

PR #200, opened 2025-05-12
Fix/issue 117: Perrin layouts improvement; Stijn's approach

stijnvermeeren-swisstopo:


src/extraction/features/stratigraphy/sidebar/classes/a_to_b_sidebar.py
```diff
@@ -34,17 +38,47 @@ def __repr__(self):
...
-                # (_, big) || (small, _)
-                segments.append(self.entries[segment_start:index])
+        index = 0
+        while index < len(self.entries):
+            # We allow sublayers with depths lower than the end of the current entry, as long as the next layer
+            # has a start depth that exactly matches the current end depth.
+            current_interval = self.entries[index]
+            sublayer_count = number_of_subintervals(current_interval, self.entries[index + 1 :])
+
+            if sublayer_count > 8:
```
TicaGit, line 61: why 8 ?

stijnvermeeren-swisstopo: Just a guess really. Seems like a reasonable number for the maximum number of sublayers that can be found within a single "parent" layer. I had an example where a scale "1:100" was misinterpreted as a parent layer from 1m to 100m, and then the entire borehole profile below was then incorrectly reduced to being sublayers of this incorrectly extracted layer. My hope if that with https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/195 we can interpret the correct start and end of the borehole profile more accurately anyway, and then this check might become redundant.

Issue #121, opened 2025-02-18
Extract vertical lines


stijnvermeeren-swisstopo:
[Migrated from LGVISIUM-107.]

Currently, vertical lines are not retrieved and are excluded from the set of geometric lines on a page. We use the geometric (horizontal/diagonal) lines f.e to split pairs of AAboveBSidebar and material description boxes.

However, vertical lines could potentially help to facilitate finding correct sidebars or restrict material description bounding boxes etc.

Goal is to retrieve vertical lines and visualize them.

Screenshot of the current line detection in zurich/267123021-bp.pdf :

![Image](https://github.com/user-attachments/assets/971e4117-0fda-4683-b445-9193ee79c316)

stijnvermeeren-swisstopo:
It could be that we use the "slope" of lines in some places in the code. Vertical lines will have a slope of "infinity" which might break these parts of the code. They should be changed so that they can deal with vertical lines.

letao:
@stijnvermeeren-swisstopo For now the issue is ready to go to pull request, however two questions remain from my side:

1) Just a clarification, the scope here is to only extract and plot the vertical lines without using them for process_page() step in the extraction script: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/edd68d869ef86a47f1dd562882802947e6a9e146/src/extraction/main.py#L274-L277

2) Currently the line detection for vertical lines is quite basic, it will look for verticality without adjusting for specific length, which results in many of letters and some of numbers having vertical lines drawn for them. Example below from zurich/266126001-bp.pdf:

![Image](https://github.com/user-attachments/assets/8c86e072-f030-4c30-8e63-97117225863b)

Would we want to adjust for this already as part of this issue or leave them for now as is?

PR #201, opened 2025-05-12
Feat/issue 121/extract vertical lines

letao:
Addresses https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/121

Extracts and plots vertical lines as part of the extract and plot the vertical lines as part of the extract_lines() function. The line extracted are currently not available as part of the process_page() function and only serves as illustrations in the artifacts. In future issues the lines can be passed to process_page() function where corresponding methods for their usage need to be developed:

https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/bb4e64f9115ada2c3cc4a64de9e0742bdeabd656/src/extraction/main.py#L274-L278

To keep the compatibility with current approach the naming of the non_vertical lines is kept as geometric_lines in the main.py script

The lines are drawn automatically when the extraction pipeline is executed with the "-l" flag.

### DIFF
--- a/src/extraction/annotations/plot_utils.py
+++ b/src/extraction/annotations/plot_utils.py
@@ -76,19 +76,59 @@ def _convert_line_to_grid(line: Line, scale_factor: float) -> Line:
     return Line(start, end)
 
 
-def plot_lines(page: pymupdf.Page, lines: list[Line], scale_factor: float = 2) -> cv2.COLOR_RGB2BGR:
-    """Given a page object and the lines detected in the page, plot the page with the detected lines.
+def plot_lines(page: pymupdf.Page, non_vertical_lines: list[Line], vertical_lines: list[Line] = None, 
+               scale_factor: float = 2.0) -> np.ndarray:
+    """Plot detected lines on a page image.
 
     Args:
-        page (pymupdf.Page): The page to draw the lines in.
-        lines (ArrayLike): The lines detected in the pdf.
-        scale_factor (float, optional): The scale factor to apply to the pdf. Defaults to 2.
+        page (pymupdf.Page): The page to plot lines on.
+        non_vertical_lines (list[Line]): The non-vertical lines to plot.
+        vertical_lines (list[Line], optional): The vertical lines to plot. Defaults to None.
+        scale_factor (float, optional): The scale factor. Defaults to 2.0.
+
+    Returns:
+        np.ndarray: The image with the lines drawn on it.
     """
-    open_cv_img = convert_page_to_opencv_img(page, scale_factor=scale_factor)
+    pix = page.get_pixmap(matrix=pymupdf.Matrix(scale_factor, scale_factor))
+    img = np.frombuffer(pix.samples, dtype=np.uint8).reshape(pix.h, pix.w, 3)
+    img_copy = img.copy()
 
-    open_cv_img = _draw_lines(open_cv_img, lines, scale_factor=scale_factor)
+    # Draw non-vertical lines in one color (green)
+    for line in non_vertical_lines:
+        cv2.line(
+            img_copy,
+            (int(line.start.x), int(line.start.y)),
+            (int(line.end.x), int(line.end.y)),
+            (0, 255, 0),
+            2,
+        )
 
-    return open_cv_img
+    # Draw vertical lines in a different color (red)
+    if vertical_lines:
+        for line in vertical_lines:
+            cv2.line(
+                img_copy,
+                (int(line.start.x), int(line.start.y)),
+                (int(line.end.x), int(line.end.y)),
+                (0, 0, 255),
+                2,
+            )
+
+    return img_copy
+
+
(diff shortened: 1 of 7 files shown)


src/extraction/features/utils/geometry/linesquadtree.py
```diff
@@ -101,7 +101,7 @@ def _qtree_insert(self, point: Point, line_key: str):
     def _qtree_delete(self, point: Point, line_key: str):
         coordinates = (round(point.x), round(point.y))
         qtree_point = self.qtree.find(coordinates)
-        if qtree_point:
+        if qtree_point in qtree_point.data:
```
stijnvermeeren-swisstopo, line 104: How can an object be contained in its own attribute?

src/extraction/features/utils/geometry/geometric_line_utilities.py
```diff
@@ -14,20 +14,38 @@
...
+    Returns:
+        tuple[list[Line], list[Line]]: A tuple containing (non_vertical_lines, vertical_lines)
+    """
+    # Maintain original functionality for backward compatibility
+    non_vertical_lines = [line for line in lines if np.abs(line.slope) < 1 / threshold]
+    vertical_lines = [line for line in lines if np.abs(line.slope) >= 1 / threshold]
+    return non_vertical_lines, vertical_lines
+
+
 def drop_vertical_lines(lines: list[Line], threshold: float = 0.1) -> ArrayLike:
```
stijnvermeeren-swisstopo, line 35: method not used anymore (only in unit test)?

src/extraction/annotations/plot_utils.py
```diff
@@ -76,21 +76,44 @@ def _convert_line_to_grid(line: Line, scale_factor: float) -> Line:
...
-    open_cv_img = _draw_lines(open_cv_img, lines, scale_factor=scale_factor)
+    open_cv_img = _draw_lines(open_cv_img, non_vertical_lines, scale_factor=scale_factor)
+
+    if vertical_lines:
+        open_cv_img = _draw_lines(open_cv_img, vertical_lines, scale_factor=scale_factor)
 
     return open_cv_img
 
 
+def plot_single_line_set(page: pymupdf.Page, lines: list[Line], scale_factor: float = 2.0) -> np.ndarray:
```
stijnvermeeren-swisstopo, line 103: method not used?

stijnvermeeren-swisstopo (changes requested):
Is there a reason for keeping the vertical and non-vertical lines in separate variables throughout the code? It seems to me that the code would be simpler if you just have one list `geometric_lines` with both vertical and non-vertical lines. Then you could for examples add a property `is_vertical` on the `Line` object, so that you can easily filter out the vertical lines in those places where we explicitly don't want to consider them. Wouldn't that be simpler, or am I missing something?

src/extraction/annotations/plot_utils.py
letao: This has been removed

src/extraction/features/utils/geometry/linesquadtree.py
letao: This was an oversight from my part, the line is missing a crucial element that was accidentally deleted:

if qtree_point and line_key in qtree_point.data:

The aim of this was to address a key error that appears when dealing with both vertical and non_vertical lines in the pipeline call. I believe there is a a chance that the same data gets processed more than once resulting in the key error.

letao:
The aim of the current approach of having separate variable throughout the code was to avoid any interaction with current functionality. Vertical lines were fully isolated from non_vertical. This might have been done in a way that is not too elegant.

I have proposed an alternative approach where the line separation is addressed within the line_detection.py script using the line_type parameter:

https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/bd94c7d5f5dd69d24fe4fc9e40f14913b404546a/src/extraction/features/utils/geometry/line_detection.py#L75-L102

This allows us to select the specific types of lines we want to extract in the main script:

https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/bd94c7d5f5dd69d24fe4fc9e40f14913b404546a/src/extraction/main.py#L273-L278

While also allowing us to select additional lines to plot in the artifact:

https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/c41a60eafbbb37aa3bc75b26c6311dc02817848d/src/extraction/main.py#L304-L310

In my opinion this keeps the functionality while also allowing us to quickly choose the way we separate the lines. If it is still preferable to have a separate property in the Line object it can be changed.

letao:
@stijnvermeeren-swisstopo 

Two new issues addressing the comments have been created:

https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/206 and https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/207.

I was not able to reproduce the issue with 691251009-bp.pdf, it might be that I am missing something but in terms of bounding boxes and metrics the result seems to be identical when running feat/issue-121/extract-vertical-lines vs main

stijnvermeeren-swisstopo (approved):

PR #203, opened 2025-05-14
Feat/issue 193/bedrock lithology

TicaGit:
This PR updates the Bedrock classifier implementation along with configuration files to support both USCS and lithology classification systems. Key changes include:

A new script (transform_yaml_lithology_patterns.py) for converting the baseline YAML to a Bedrock‐compatible format.
Updates in AWSBedrockClassifier and main.py to utilize classifier attributes instead of reading config parameters repeatedly.
Refactors and updates to various configuration files (prompts, patterns, and configs) as well as launch.json to support the new bedrock classification.


src/classification/classifiers/aws_bedrock_classifier.py
```diff
@@ -2,58 +2,77 @@
...
+
 
 class AWSBedrockClassifier:
     """AWSBedrockClassifier class uses AWS Bedrock with underlying Anthropic LLM models."""
 
-    def __init__(self, bedrock_out_directory: Path | None, max_concurrent_calls: int = 1, api_call_delay: float = 0.0):
+    def __init__(
+        self,
+        bedrock_out_directory: Path | None,
+        classification_system_str: str,
```
stijnvermeeren-swisstopo, line 28: Would it make sense to already required a typed classification system here (i.e. an instance of `ClassificationSystem` here) so we don't have to worry so much about dealing with strings that don't correspond to a supported classification system?

stijnvermeeren-swisstopo (approved):
Based on a qualitative analysis of the classification results, I've created the following follow-up tickets:
- https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/213
- https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/214
- https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/215

The implementation looks good to me (except for the one comment that we agreed to deal with later on, after the other PRs have been merged as well), and the results are more than good enough for a first version. So this PR can be merged as far as I'm concerned. :)

letao (approved):

PR #208, opened 2025-05-16
Feat/issue 205/stratigraphy endpoint

### LINKED ISSUE #205: API endpoint for stratigraphy extraction from file
Add a new endpoint to the API for extracting all boreholes with their stratigraphy (depths and corresponding material descriptions) from a PDF file. This endpoint will be used by the boreholes application for https://github.com/swisstopo/swissgeol-boreholes-suite/issues/1963.

**Endpoint**
I suggest creating a new endpoint, rather than adding a new _format_ to the existing `extract_data` endpoint, because in contrast to that endpoint, we are no longer requiring the user to specify page and a bounding box. The endpoint should scan the entire PDF file and return all boreholes that are found in the PDF (with one borehole potentially spanning multiple pages).

Suggested endpoint name: `/api/V1/extract_stratigraphy`

To be decided (check with Geowerkstatt): the application will need PNGs of the PDF pages in order to show the visualization of the extracted data to the end-user. In order to produce these PNGs, should the existing `/api/V1/create_pngs` endpoint be called separately, or should the PNGs be automatically created when calling this new endpoint?

**Request**
Example request body:
```
{
  "filename": "1007.pdf"
}
```

**Response**
Suggested response structure:
```
[
  {
    "page_numbers": [2, 3]
    "layers": [
      {
        "material_description": {
          text: "Kies"
          bounding_boxes: [{ page_number: ..., x0: ..., y0: ...., x1: ..., y1: ... }]
        },
        "start": {
          depth: 0
          bounding_boxes: [{ page_number: ..., x0: ..., y0: ...., x1: ..., y1: ... }]
        },
        "end": {
          depth: 1.2
          bounding_boxes: [{ page_number: ..., x0: ..., y0: ...., x1: ..., y1: ... }]
        }
      },
      ...
    ]
  },
  ...
]
```

The `page_numbers` should list all the (unique) page numbers from all the bounding boxes of that borehole. It seems beneficial to include this as a separate field in the response, so that the application can quickly check which PDF pages are relevant for the extracted borehole, without having to scan through all the bounding boxes in the response.

Bounding boxes will be used for producing a visualization in the application, similar to how this repository creates the PNG visualizations. Bounding boxes should be provided on a line-by-line basis, i.e. most start/end depths will only have one bounding box, but many material descriptions will have several lines and hence several bounding boxes. The bounding boxes object is similar to the one from the /api/V1/bounding_boxes endpoint, but not exactly the same (in this case, we also need a page number for every bounding box, as a single material description theoretically could span across different PDF pages.

Any of the fields `material_description`, `start` and `end` could be `null` for a given layer. I.e. we could extract a material description without any known depths, or we could extract a layer with known depths but no corresponding material description.

TicaGit:
Resolves #205

### DIFF
--- a/src/app/api/v1/endpoints/extract_stratigraphy.py
+++ b/src/app/api/v1/endpoints/extract_stratigraphy.py
@@ -1,0 +1,101 @@
+"""This module defines the FastAPI endpoint for extracting all boreholes with their stratigraphy."""
+
+from app.common.aws import load_pdf_from_aws
+from app.common.schemas import (
+    BoreholeExtractionSchema,
+    BoreholeLayerSchema,
+    ExtractStratigraphyResponse,
+)
+from extraction.features.extract import process_page
+from extraction.features.stratigraphy.layer.duplicate_detection import remove_duplicate_layers
+from extraction.features.stratigraphy.layer.layer import LayersInDocument
+from extraction.features.utils.geometry.line_detection import extract_lines
+from extraction.features.utils.text.extract_text import extract_text_lines
+from fastapi import HTTPException
+from utils.file_utils import read_params
+from utils.language_detection import detect_language_of_document
+
+matching_params = read_params("matching_params.yml")
+line_detection_params = read_params("line_detection_params.yml")
+
+
+def extract_stratigraphy(filename: str) -> ExtractStratigraphyResponse:
+    """Extract all boreholes with stratigraphy from the entire PDF file.
+
+    This endpoint scans all pages of the PDF and returns all boreholes found,
+    each with page numbers and a list of stratigraphy layers (material descriptions,
+    start depths, end depths), including bounding boxes with page references.
+
+    Args:
+        filename (str): filename of the PDF.
+
+    Returns:
+        List[BoreholeExtraction]: List of boreholes with their stratigraphy layers and bounding boxes.
+    """
+    try:
+        document = load_pdf_from_aws(filename)
+    except Exception as e:
+        raise HTTPException(status_code=400, detail="Cannot open PDF file") from e
+
+    language = detect_language_of_document(
+        document, matching_params["default_language"], matching_params["material_description"].keys()
+    )
+
+    layers_with_bb_in_document = LayersInDocument([], filename)
+    for page_index, page in enumerate(document):
+        # TODO should be funcs
+        page_number = page_index + 1
+        text_lines = extract_text_lines(page)
+        geometric_lines = extract_lines(page, line_detection_params)
+        extracted_boreholes = process_page(text_lines, geometric_lines, language, page_number, **matching_params)
+        processed_page_results = LayersInDocument(extracted_boreholes, filename)
+        if page_index > 0:
+            layer_with_bb_predictions = remove_duplicate_layers(
+                previous_page=document[page_index - 1],
+                current_page=page,
+                previous_layers_with_bb=layers_with_bb_in_document,
+                current_layers_with_bb=processed_page_results,
+                img_template_probability_threshold=matching_params["img_template_probability_threshold"],
+            )
+        else:
+            layer_with_bb_predictions = processed_page_results.boreholes_layers_with_bb
+        layers_with_bb_in_document.assign_layers_to_boreholes(layer_with_bb_predictions)
+
+    if not layers_with_bb_in_document.boreholes_layers_with_bb:
+        raise HTTPException(status_code=404, detail="No boreholes found in PDF.")
+
+    extracted_stratigraphy = create_response_object(layers_with_bb_in_document)
+
+    return extracted_stratigraphy
+
+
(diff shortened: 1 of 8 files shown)


letao:
I think the implementation makes sense. I have created a follow up issue for the environment variable checks: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/212

Additionally regarding to this section: 

> To be decided (check with Geowerkstatt): the application will need PNGs of the PDF pages in order to show the visualization of the extracted data to the end-user. In order to produce these PNGs, should the existing /api/V1/create_pngs endpoint be called separately, or should the PNGs be automatically created when calling this new endpoint?

I do not have a straight up answer here, there are multiple ways on how to address this, we can leave the endpoints maximally separated for each action, e.g., PNG creation and stratigraphy are handled separately in a sequential manner via API calls. Alternatively we could also implement something like this: `POST stratigraphy/<file_id>/action` where a specific action (extract, generate_png) is sent in the request body, and file name could be sent as part of the request header. This however, increases complexity of the endpoint and requires more extensive documentation.

For simplicity we can leave it with the current implementation and see if Geowerkstatt sees any benefits from a combined response with png and stratigraphy.

src/app/api/v1/endpoints/extract_stratigraphy.py
```diff
@@ -0,0 +1,98 @@
...
+            current_page_index=page_index,
+            document=document,
+            previous_layers_with_bb=layers_with_bb_in_document,
+            current_layers_with_bb=layers_on_page,
+            img_template_probability_threshold=matching_params["img_template_probability_threshold"],
+        )
+        layers_with_bb_in_document.assign_layers_to_boreholes(layer_with_bb_predictions)
+
+    if not layers_with_bb_in_document.boreholes_layers_with_bb:
+        raise HTTPException(status_code=404, detail="No boreholes found in PDF.")
```
stijnvermeeren-swisstopo, line 62: Maybe it's better to just return an empty list in this case, rather than a 404 error? After all, if there is no borehole in the PDF to start with, the extraction is still successful when it doesn't extract anything. It seems something very different compared to when the API endpoint (or even the PDF file) does not exist.

src/app/api/v1/endpoints/extract_stratigraphy.py
```diff
@@ -0,0 +1,98 @@
...
+        layers_on_page = LayersInDocument(extracted_boreholes, filename)
+
+        layer_with_bb_predictions = remove_duplicate_layers(
+            current_page_index=page_index,
+            document=document,
+            previous_layers_with_bb=layers_with_bb_in_document,
+            current_layers_with_bb=layers_on_page,
+            img_template_probability_threshold=matching_params["img_template_probability_threshold"],
+        )
+        layers_with_bb_in_document.assign_layers_to_boreholes(layer_with_bb_predictions)
```
stijnvermeeren-swisstopo, line 59: Seems like there is quite a bit of duplication between the code here and the code in extraction/main.py. Can we create a/some shared method(s) that reduce this duplication a little?
It would also be cleaner if this file does not need to import methods from so many different files under the extraction package.

stijnvermeeren-swisstopo (changes requested):
I tested different cases, including:
- multiple boreholes per page
- a single boreholes spanning multiple pages
- a borehole without extracted depths per layers

In all cases, the responses look good!

Just two minor comments to look at.

stijnvermeeren-swisstopo (changes requested):
Thanks for the refactoring.
I've committed one minor change: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/208/commits/611b0a9d208df7ed8d2ff84af56666851d0e6ec1
One more duplicated line seems to be better than having a harder-to-understand side-effecting method that mutates the `existing_layers` value inside the method.

I've also fixed a couple of typos and renamed `existing_layers` to `layers_from_previous_pages`. (Because what does "existing" mean anyway? Every layer that we extract should exist, hopefully. :)) 

What remains to be done is what we discussed this morning: verify that the bounding box coordinates are consistently returned across all API endpoints (avoid that we have some endpoints use PNG pixels as the unit and other endpoints PDF points).

src/app/common/schemas.py
```diff
@@ -38,7 +41,12 @@ class PNGRequest(BaseModel):
 
     filename: Path  # This will ensure the filename is a Path object
 
-    model_config = ConfigDict(arbitrary_types_allowed=True)  # This allows using non-standard types like Path
+    model_config = ConfigDict(
+        arbitrary_types_allowed=True,  # This allows using non-standard types like Path
+        json_schema_extra={
+            "example": {"filename": "geoquat/validation/1007.pdf"},
```
stijnvermeeren-swisstopo, line 47: @TicaGit could you still adjust all of these examples according to yesterday's discussion?

PR #217, opened 2025-05-22
Fix/issue 207/geometric lines improvement

### LINKED ISSUE #206: Improve the LSD algorithm after introduction of vertical lines in the extraction pipeline
After the implementation of https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/121 (pull request: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/201) We can observe an artifact as a result of the new line extraction approach where we now also draw vertical lines. Specifically in geoquat/validation/B83.pdf. 

- In case of zurich/B83.pdf, it looks like the line detection is done at an excessive zoom level, leading to many detected lines which should not be used:

Image

To address this we should explore two options applying a preprocessing logic before the LSD algorithm, or by fine-tuning parameters of the Line Segment Detector (Parameters: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/main/config/line_detection_params.yml).

Additionally we should investigate if similar artifacts to zurich/B83.pdf are observed in other files.

### LINKED ISSUE #207: Improve the merge merge_parallel_lines_quadtree method
After the implementation of https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/121 (pull request: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/201) the extraction pipeline has become significantly slower (~ 10% to 20% slower). 

The main suspect for this is the merge_parallel_lines_quadtree method:

https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/edd68d869ef86a47f1dd562882802947e6a9e146/src/extraction/features/utils/geometry/geometric_line_utilities.py#L207-L209

To address that we should investigate the merge_parallel_lines_quadtree method and look at potential ways to make it more efficient (decrease runtime, find more efficient computation) and investigate if we can also increase accuracy.

TicaGit:
Resolves #206
Resolves #207

This PR introduces changes to the parameters of the merging algorithm. The Line Segment Detector (LSD) was detecting many small lines, which slowed down the pipeline and did not contribute meaningful information. These lines are now eliminated if they are shorter than a specified length. However, very short horizontal lines are preserved, as they remain important for our use case (see below, the horizontal lines between the text lines).

fgdasgadf

To ensure adaptability across different document sizes, the length threshold and related parameters now scale with the width of each page

As expected, the evaluation metrics show a slight decrease, since removing or merging lines inevitably leads to some loss of detail. Despite this, the updated pipeline shows a performance improvement of approximately 15–20%. For example: Zurich: 329s vs. 409s and Geoquat/val: 395s vs. 466s

Additionally, a Hough Transform-based algorithm was implemented, using morphological operations to enhance line continuity. Although LSD still performs better for detecting fine details in this use case, I suggest retaining the Hough implementation, as it could be useful for future applications.

Finally, a fix was applied to the are_parallel condition to handle the discontinuity of tan() at π/2. This improves robustness when merging vertical lines


src/extraction/features/utils/geometry/line_detection.py
```diff
@@ -68,9 +118,25 @@ def extract_lines(page: pymupdf.Page, line_detection_params: dict) -> list[Line]
...
+    # return lines
+    tol = merging_params["merging_tolerance"] * scaling_factor
+    angle_tol = merging_params["angle_threshold"]
+    min_line_length = merging_params["min_line_length"]
+    horizontal_slope_tolerance = merging_params["horizontal_slope_tolerance"]
+    merged = merge_parallel_lines_quadtree(lines, tol=tol, angle_threshold=angle_tol)
+    merged = [
+        line
+        for line in merged
+        if line.length >= min_line_length * scaling_factor or abs(line.slope) <= horizontal_slope_tolerance
```
stijnvermeeren-swisstopo, line 140: I think that in some cases we are losing too much valuable information by rejecting shorter non-horizontal lines, for example as A548.pdf.

Instead of 
image
we get 
image

Notice that half of the vertical line that separates the material descriptions & depth from the graphical profile is missing. This however would be an important line to be aware of, especially thinking about future issues such as https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/195. So either we should relax this condition, or (even better) we should improve the merging algorithm to ensure that this entire vertical line is extracted as a single line segment.

By the way, this line is recognized better by the Hough transform, but on the other hand, the current Hough transform implementation has severe troubles with recognizing the horizontal lines:
image

TicaGit:
I never encountered the segmentation fault issue on my machine, but it’s likely related to the profiler. Running the pipeline across many documents triggers a huge number of function calls, so it’s possible that memory usage spikes and eventually crashes the process. I'll remove the @profile decorator, and we should only put it back when running the pipeline on few documents, to have an idea of each functions's runtime.

I’ve already experimented a bit with the LSD parameters (and tried again just now), but didn’t observe any real improvement. The values we currently use are just the default ones from OpenCV. The only parameter that might impact detection quality is the scale—it controls the image resolution passed to the algorithm, and increasing it tends to pick up more fine-grained lines. But doubling the scale value roughly quadruples the runtime, since it results in many more detected lines and merging becomes significantly slower. So overall, the tradeoff doesn’t seem worth it in this case.

And yes, I’m using cv2.HoughLinesP, which applies the probabilistic Hough transform and directly returns segments (with start and end points). That might be the main difference, Renato probably used the standard Hough transform, which just gives the line parameters and not the actual segment endpoints.

src/extraction/features/utils/geometry/line_detection.py
TicaGit: Yes, I agree, finding the right tradeoff is tricky. Depending on how we use the extracted lines, we might need different ones to be preserved. Maybe the Hough transform (for segments) could be a better fit for something like #195, but it’s definitely not reliable when it comes to capturing the smaller lines that separate text blocks.

I also tried tweaking the merging to address cases like the one in A548.pdf, but without any real improvement. I experimented with increasing the tolerance for merging lines that are nearly horizontal or vertical, but that ended up introducing unwanted merges. The challenge is that deciding which lines to merge really depends on context, and it’s hard to capture that with simple rules, it becomes more of a case-by-case decision, which is tough to automate cleanly.

That said, with the current parameters, most of the mergin makes sense (e.g., A548.pdf cleaner, dashed lines being continued, interrupted lines getting rejoined), and I don’t think we’re getting too many incorrect merges overall. To achieve that, I decided to slightly relax the constraint on line length, just to try and preserve more of the lines we care about, especially vertical ones like in A548. This will slightly increase the pipeline’s runtime, but it feels like a reasonable tradeoff.

PR #219, opened 2025-05-22
reduce compling, centralize logic

TicaGit:


config/classifier_config_paths.yml
```diff
@@ -0,0 +1,12 @@
+dummy:
+  uscs: null
+  lithology: null
+baseline:
+  uscs: classification_params.yml
```
stijnvermeeren-swisstopo, line 5: should we rename this to `uscs_classification_params.yml` ?

src/classification/classifiers/baseline_classifier.py
```diff
@@ -15,36 +17,40 @@ class BaselineClassifier:
...
+        self.classification_system = classification_system
 
         self.match_threshold = match_threshold
 
         self.stemmer_languages = {"de": "german", "fr": "french", "en": "english", "it": "italian"}
 
         self.stemmers = {}
 
-    def init_config(self, classification_system: str):
+    def init_config(self, classification_system: type[ClassificationSystem]):
```
stijnvermeeren-swisstopo, line 37: Can we put this into a shared _Classifier_ abstract parent class?

stijnvermeeren-swisstopo (changes requested):

config/bert/bert_config_lithology.yml
```diff
@@ -3,13 +3,13 @@ model_path: "google-bert/bert-base-multilingual-uncased"
...
 classification_system: "lithology"
-train_subset: "nagra_deepwells_custom_split/train.json"
-eval_subset: "nagra_deepwells_custom_split/val.json"
+train_subset: "lithology_splits/train.json"
+eval_subset: "lithology_splits/val.json"
 
 # Training hyperparameters
 batch_size: 32
 num_epochs: 32
-learning_rate: 1e-4
```
stijnvermeeren-swisstopo, line 12: should this have changed? (Same for `# - "layer_11"` below.)

TicaGit: let's leave it like that. The value of lr=3e-3 and keeping the layer_11 frozen is the default setting for the first stage of the training.

PR #223, opened 2025-05-27
Fix/issue 212/api error handling

### LINKED ISSUE #212: Better error handling for the API when no AWS endpoint or S3 Bucket is not provided
Currently we do not specifically handle missing AWS endpoint or S3 Bucket in our API config. 

For the S3 Bucket if missing, a default value is provided, which might not be the intended behavior: 
https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/3930ac8736a71e3d07074e9057a4692465e94119/src/app/common/config.py#L12-L14

Additionally for the AWS endpoint we could check if an endpoint is provided. The check is simply done on the boto client and will result in 500 without an explanation if an incorrect value is provided, or continue if no value is provided, i.e. removing AWS_ENDPOINT from your env file should result in default computer wide connection settings used, which might not be the intended result:
https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/3930ac8736a71e3d07074e9057a4692465e94119/src/app/common/aws.py#L33-L49 

We should investigate what would be the best way to flag missing env variables.

TicaGit:
Resolves #212 

This PR addresses api error handling improvements by updating AWS bucket configuration, testing S3 client connectivity, and refining API error responses.

### DIFF
--- a/src/app/common/config.py
+++ b/src/app/common/config.py
@@ -8,7 +8,6 @@ from pydantic_settings import BaseSettings, SettingsConfigDict
 
 dotenv.load_dotenv(override=True)
 
-TEST_BUCKET_NAME = "test-bucket"
 DEFAULT_BUCKET_NAME = "stijnvermeeren-boreholes-integration-tmp"
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,12 +3,24 @@
 import boto3
 import pytest
 from app.common.aws import get_s3_client
-from app.common.config import TEST_BUCKET_NAME, config
+from app.common.config import config
 from app.main import app
 from fastapi.testclient import TestClient
 from moto import mock_aws
 
-config.bucket_name = TEST_BUCKET_NAME
+TEST_BUCKET_NAME = "test-bucket"
+
+
+@pytest.fixture(autouse=True, scope="session")
+def override_bucket_name():
+    """Automatically override the bucket name before the test session."""
+    config.bucket_name = TEST_BUCKET_NAME
+
+
+@pytest.fixture(scope="session")
+def test_bucket_name():
+    """Returns the name of the test bucket."""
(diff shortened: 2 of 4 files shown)


src/app/common/aws.py
```diff
@@ -94,6 +127,9 @@ def load_data_from_aws(filename: Path, prefix: str = "") -> bytes:
 
     # Check if the document exists in S3
     try:
+        if config.bucket_name == DEFAULT_BUCKET_NAME:
+            # logging is here to avoid circular imports error in the config module
+            logger.warning(f"No bucket name provided, defaulting to {DEFAULT_BUCKET_NAME}")
```
stijnvermeeren-swisstopo, line 132: This log message should not appear if the AWS_S3_BUCKET environment variable is explicitly set to the default value.

stijnvermeeren-swisstopo (changes requested):

src/app/common/aws.py
```diff
@@ -44,10 +51,36 @@ def create_s3_client():
...
+            raise HTTPException(status_code=500, detail=f"AWS Client error: {error_code}") from None
+    except EndpointConnectionError as e:
+        raise HTTPException(
+            status_code=500,
+            detail="Server misconfiguration: could not connect to S3 endpoint URL: "
+            f"{e.kwargs.get('endpoint_url', 'Unknown')}",
+        ) from None
+
+    if config.bucket_name and config.bucket_name not in {bucket["Name"] for bucket in client_response["Buckets"]}:
+        raise HTTPException(status_code=404, detail=f"No bucket named {config.bucket_name}") from None
```
stijnvermeeren-swisstopo, line 81: This message could be somewhat misleading, as the endpoint only lists the endpoint that you have access to with the configured credentials, so it could be that the endpoint exists, but you have no access.

PR #231, opened 2025-06-04
Feat/issue 213/spulprobe documents

TicaGit:


TicaGit:
Our current pipeline does not support pymupdf 1.26, we force the version to 1.25.5.

lhaibach (commented):
Looks good to me :) I have one comment regarding the entry detection for the Spülproben: currently, the lowest depth is used, but in my opinion, it would make more sense to use the highest when constructing the intervals - unless I’m misunderstanding something. Let me know if I got that wrong!

Also, I agree that having open intervals at the beginning and end of a sidebar is useful - especially if we want to handle layer continuity across page gaps in the future.

I ran the pipeline on geoquat/validation (compared to main metrics increase) and also tested the files you mentioned. I didn’t execute the deskew script myself, but I quickly reviewed the code :)

If anything I wrote is unclear or incorrect, please  let me know!

letao (approved):

PR #232, opened 2025-06-05
remove the validation error

TicaGit:
This PR removes the bounding box positive-value validation to eliminate the associated validation error.


stijnvermeeren-swisstopo (approved):
Code looks good to me, but as Davis said, let's only close this after the feedback from Geowerkstatt has been documented in the issue.

Issue #216, opened 2025-05-21
Try out prompt caching for AWS Bedrock with Claude Sonnet 3.7


stijnvermeeren-swisstopo:
AWS documentation: https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html

If we upgrade to the Claude Sonnet 3.7 model, we should be able to use the "prompt caching" feature. If we can cache the long "system prompt" for the USCS and lithology classifiers, then the inference on a larger dataset might become significantly faster, and it might become cheaper as well. Let's put both of these hypotheses to the test (and also check if the upgrade from Sonnet 3.5 to Sonnet 3.7 has any influence on the classification metrics).

PR #234, opened 2025-06-11
Implementation of prompt caching for AWS Bedrock classification

letao:
Addresses: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/216 

Implements prompt caching for AWS Bedrock Classifier. Currently the implementation is handled in message construction by passing "cache_control" argument to together with the classification message:

https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/ec018f9e5561f7eff1a2490df5a766766aa034c7/src/classification/classifiers/aws_bedrock_classifier.py#L99-L109

Currently two cache points are set, for system instructions (system_message) and for general user instructions (user_message_instructions). As a result the current implementation requires to have access to the whole set of classification patterns (english, german, french and italian) from  classification_patterns_bedrock JSON files. 

As a result of this implementation it is possible to use max_concurrent_calls for the AWS Bedrock Classifier in the classifier_factory.py file. The general response time decrease with prompt cache and concurrent calls is significant from around 2.3h for the geoquat to around 38min. 

As a result of using Sonnet 3.7 models we also see a slight increase in our classification metrics for USCS classification, where Bedrock now is around the same accuracy as our baseline model:
USCS-metrics

We also see slight improvement in lithology classification specifically in the macro level metrics:
lithology-metrics

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -32,7 +32,7 @@
                 //"data/geoquat/validation/13080.pdf", //has f1=0, detects 1x the same, and not the right things
                 //"data/zurich/267125439-bp.pdf",
                 //"data/geoquat/validation/A7111.pdf",
-                //"data/geoquat/validation",
+                "data/geoquat/validation",
                 //"data/geoquat/train", #14225.pdfok, A11406.pdfNO, 1488.pdf, 14273.pdfok and A1304
                 // "data/geoquat/validation/937.pdf",
                 //"data/geoquat/test/A8005.pdf",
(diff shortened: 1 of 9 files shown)


stijnvermeeren-swisstopo (changes requested):
I did not see any performance improvement when running lithology extraction on a subset of 12 nagra boreholes. Indeed, the runtime was actually slightly longer. I did not activate any concurrency in the calls. @letao could you maybe verify whether you also see a faster runtime with prompt caching but without concurrency? If you do, then maybe I've misconfigured something...

I can check tomorrow if there has been a significant change in cost. Unfortunately, the AWS cost explorer only updates its data once per day.

stijnvermeeren-swisstopo:
The AWS cost dropped by about 50%, which is a good sign.
Is this number plausible? @letao can you do a quick count of how many tokens are cached and how many tokens are fresh for an average every request?

src/classification/classifiers/aws_bedrock_classifier.py
```diff
@@ -89,25 +89,30 @@ def create_message(
...
+            material_description=material_description
+        )
 
         body = json.dumps(
             {
                 "anthropic_version": anthropic_version,
                 "max_tokens": max_tokens,
                 "temperature": temperature,
-                "system": system_message,
+                "system": [{"type": "text", "text": system_message, "cache_control": {"type": "ephemeral"}}],
```
stijnvermeeren-swisstopo, line 105: Maybe add a small comment here, explaining the advantages of prompt caching and linking to the AWS documentation.

letao: The comment is now added before the body variable.

TicaGit (approved):
LGTM!

PR #237, opened 2025-06-23
feat/issue-230/advanced-line-merging

### LINKED ISSUE #230: Geometric line merging issues
**1) Short, parallel lines are incorrectly merged**

Example: geoquat/validation/A1304.pdf.

Raw line detection results (without merging):
Image

Merged results:
Image

Lines appear in the merged results that don't have a good basis in the raw line detection results. In this case, this even affects the extraction accuracy, because of an incorrectly merged line in this material description, which causes the description to be incorrectly split:

Image

Image

The cause seems to be that we are merging short line segments from letters, line segments that are parallel to each other, but whose distance (measured perpendicular to the direction of the segments) is fairly large relative to the length of the segments (e.g. the line segments corresponding to the letters "l" in the word "lol"). This does not seem very reasonable to merge those.

As a fix, I would propose two changes:
- In `is_point_on_line()`, the tolerance for `line.distance_to(point)` should be significantly smaller compared to the tolerance for how far away the point may be from the end points of the line segment (along the direction of the line).
- In `merge_parallel_lines_quadtree`, check that the new line is still parallel with both of the input lines, and if not, abort the merge.

**2) Method `_are_close()` is not symmetric**

The method `_are_close()` is described as if it behaves symmetrically with respect to the two parameters `line1` and `line2`. However, this is not the case. If `line2` is a subsegment of `line1`, then the method will return `True`, but if `line1` is a subsegment of `line2`, then (contrary to what I would expect) it will return `False`. 

**3) Method `is_point_on_line` should not give a special treatment to horizontal lines** 

The description of the method `is_point_on_line` mentions:
> We assume lines are horizontal, and allow for less margin in the y-coordinate to keep merged lines horizontal. 

This might have made sense for the original use case, but not that we want to use the geometric lines for other purposes that also require accurate information about non-horizontal lines, we should not give any special treatment to horizontal lines anymore.

The different tolerances for horizontal and vertical might be redundant anyway, once we implement a fix for point 1) in this issue.

**4) Unit tests**

The fact that these issues are only being noticed now, highlights the need for more unit tests for these methods.

TicaGit:
Resolves #230 

This PR address all the points mentioned in the issue.

1) **Short, parallel lines are incorrectly merged**

In `geoquat/validation/A1304.pdf`, the text line `Kies und Sand mit Humus` is no longer split due to a parasitic line. The line-merging algorithm has been improved to avoid merging small lines that are merely side by side. It now also scales its closeness tolerance with line length, providing a relative perspective that prevents merging of small lines that are relatively far apart.
As a result, the text `Gemeinde Ostermundigen` (and many others) exhibits significantly less line noise, aside from the small horizontal lines that we are still forced to keep.

image

2) **Method _are_close() is not symmetric**

We now also check line 2 against line 1. Furthermore, _are_close() has been made more selective by using adaptive tolerances, which helps avoid merging short lines that are relatively far apart.

3) **Method is_point_on_line should not give a special treatment to horizontal lines**

Done, the tolerance is now the same in both the Y and X directions. To preserve the useful behavior that was lost due to this change, a new method is_point_near_line() was introduced. It checks whether a point lies close to a given line segment using two tolerances:
One (line_tol) ensures the point is closely aligned with the direction of the line.
The other (tol) allows the point to lie slightly beyond the segment’s endpoints.

This dual-tolerance logic helps avoid merging horizontal lines that are stacked vertically — which are often structural separators, such as between material descriptions.

4) **Unit tests**

The unit tests introduced in a previous PR now pass, and I’ve added new ones to cover the `_are_mergeable` function.

5) **Only consider horizontal lines for material description splitting**

This changes affects A11440.pdf (slightly better) and A135.pdf (slightly worst) in zurich. The split caused by those diagonals indeed makes no sense, so restricting the lines to be horizontal is good for TextBlock splitting.

### DIFF
--- a/src/extraction/features/utils/geometry/geometric_line_utilities.py
+++ b/src/extraction/features/utils/geometry/geometric_line_utilities.py
@@ -40,34 +40,37 @@ def is_point_on_line(line: Line, point: Point, tol: int = 10) -> bool:
     )
 
 
-def is_point_near_line(line: Line, point: Point, tol: int = 10, line_tol: int = 3) -> bool:
+def is_point_near_line(line: Line, point: Point, segment_extension_tol: int = 10, perpendicular_tol: int = 3) -> bool:
     """Check if a point is near a line segment.
 
     This is done by computing the slope and y-intercept of the line, and checking whether the point satisfies
-    the line equation within a tolerance (`tol_line`). Since the line is a finite segment, the function
-    also checks whether the point lies within or near the segment's bounds, using a second tolerance (`tol`).
+    the line equation within a tolerance (`perpendicular_tol`). Since the line is a finite segment, the function
+    also checks whether the point lies within or near the segment's bounds, using a second tolerance
+    (`segment_extension_tol`).
 
     The purpose of using two tolerances is to ensure that:
-    - the point is closely aligned with the line direction (controlled by `tol_line`),
-    - but we still allow the point to lie beyond the segment's endpoints (controlled by `tol`).
+    - the point is closely aligned with the line direction (controlled by `perpendicular_tol`),
+    - but we still allow the point to lie beyond the segment's endpoints (controlled by `segment_extension_tol`).
 
     Args:
         line (Line): A line segment.
         point (Point): The point to check.
-        tol (int, optional): Tolerance for point's position along the segment's direction. Defaults to 10.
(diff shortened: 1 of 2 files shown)


lhaibach (approved):
I checked the line drawings for the files you mentioned and in your pull request and it looks good to me.

PR #249, opened 2025-07-10
Reduce docker image size

TicaGit:
This PR modifies the Dockerfile to significantly reduce image size and improve build performance.

Changes:
By replacing the pip-tools installation approach (pip-compile + pip-sync) with a simpler  `pip install --no-cache-dir .`, I was able to reduce the image size.

Additionally, I tried to remove the heavy packages: transformers, datasets, torch, accelerate

Results:
With pip-tools:
~10 GB image on GitHub Actions, ~3 GB locally

With pip install .:
~6.37 GB on GitHub, ~2.22 GB locally

After removing heavy packages:
~714 MB on GitHub, ~1.12 GB locally

Trade-offs:
The image is much smaller and builds faster. However, we lose some of the reproducibility benefits from pip-sync, such as precise version locking and hash checking.

I'm not sure if this trade-off is good enough or if we should keep the approach with pip-sync.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -5,6 +5,7 @@ AWS_ACCESS_KEY_ID=your_access_key_id
 AWS_SECRET_ACCESS_KEY=your_secret_access_key
 AWS_ENDPOINT=your_endpoint_url
 AWS_S3_BUCKET=your_bucket_name
+AWS_DEFAULT_REGION=your_default_region
 
 ANTHROPIC_VERSION="bedrock_api_version"
 ANTHROPIC_MODEL_ID="your_model_id"
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -18,13 +18,12 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
-
       - name: Set environment variables
         run: |
           echo VERSION=$(cat VERSION).$GITHUB_RUN_NUMBER >> $GITHUB_ENV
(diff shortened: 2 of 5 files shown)


letao:
With the latest changes we add back pip-sync and get back reproducibility benefits. We move the transformers, datasets, torch and accelerate packages to optional dependencies. We also use two stage build and runtime setup for the docker, this can further reduce the image size.As a result our Docker Image is now around 800 MB. 

@stijnvermeeren-swisstopo, since for now we do not have a classification pipeline we can leave the packages as optional, the moment we will need to start doing classification inference the size issue due to the packages will come back (from to my tests the size should be around 6GB in that case). Do you think it makes sense to create a ticket on investigating this issue later? The idea would be to look at a lite version of the specific packages that we use for BERT models (e.g. torch) as that is the biggest culprit for the size increase.

P.S. in order to be able to better test the docker image locally I have changed the env.template file to include AWS_DEFAULT_REGION variable to keep it consistent across all usages (specifically API and Bedrock classification)

stijnvermeeren-swisstopo (approved):
I'm currently not really able to test this, as I'm using AWS Workspaces as a cloud development environment, but using Docker is not really supported there.

The code changes look good though, so I'm happy to merge this, and then we can let Geowerkstatt know that they can test the new image. Very nice that we were able to keep the advantages of pip-sync.

@letao yes, I think it's a good idea to create a follow-up issue like you suggested. The issue can then remain in the backlog until it become relevant.

PR #251, opened 2025-07-10
Feat/issue 228/page rect object

### LINKED ISSUE #228: Potentially incorrect fallback page for depth layers bounding boxes in the API
In the code for the stratigraphy extraction of the API endpoint, the fallback page for the depth layers bounding boxes introduced in #208 may sometimes be incorrect.

The LayerDepthsEntry object currently does not have a page attribute, which makes it difficult to determine the actual page associated with the depth layers bounding box. Currently, it takes the same bounding box of its corresponding material description, which may lead to inconsistencies or bugs when rendering or referencing these elements.

**Tasks:**
We should investigate whether there are cases where the page of the material description differs from that of the bounding box.

If such cases exist, consider two possible solutions:
(Less preferred) Add a page attribute to classes like LayerDepthsEntry (and possibly others). This could lead to excessive bloat across various classes.

(Preferred) Override the Rect class to include a page attribute. This is a more intuitive design, binding the page directly to the bounding box.
However, this change would have broad implications and we need to ensure that a page can be reliably inferred wherever a Rect is used.

Expected outcome: 
Resolve the TODO line 589 in `app/common/schemas.py`

TicaGit:
Resolve #228

This PR introduces a unified RectWithPage object that encapsulates both a rectangle and its associated page number.
To simplify access, classes that hold a RectWithPage now inherit from a new RectWithPageMixin class, which exposes .rect and .page_number as @propertys. This allows continued use of .rect and .page_number directly on the object, without needing to access .rect_with_page.rect or .rect_with_page.page_number.

The PR also resolves the fallback page issue by providing the page_number of the `Layer` object (taken from its `FeatureOnPage[MaterialDescription]` attribute), which is, by construction, the same as the one in the `LayerDepthsEntry` object.

### DIFF
--- a/src/app/common/schemas.py
+++ b/src/app/common/schemas.py
@@ -519,12 +519,12 @@ class LayerMaterialDescriptionSchema(BaseModel):
             text=prediction.feature.text,
             bounding_boxes=[
                 BoundingBoxWithPage(
-                    page_number=line_feature.page,
+                    page_number=line_feature.page_number,
                     x0=line_feature.rect.x0,
                     y0=line_feature.rect.y0,
                     x1=line_feature.rect.x1,
                     y1=line_feature.rect.y1,
-                ).rescale_factor(*pdf_img_scalings[line_feature.page - 1])
+                ).rescale_factor(*pdf_img_scalings[line_feature.page_number - 1])
                 for line_feature in prediction.feature.lines
             ],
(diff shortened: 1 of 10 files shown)


stijnvermeeren-swisstopo (approved):
LGTM!

PR #252, opened 2025-07-15
Feat/issue 131/deskew pages

TicaGit:


stijnvermeeren-swisstopo (approved):
I'm happy to merge this, under the disclaimer that the deskewing script is still very much in a prototype stage (with hard-coded paths, many magic numbers, log messages like "Too many lines, skipped." without any mention of the actual page number, etc.).

However, I think that we first need to clarify if and how we actually want to integrate the deskewing script into our full processing pipeline, before we spend more time on making the code more production-ready. I've created an issue in the swissgeol-ocr repository for this: https://github.com/swisstopo/swissgeol-ocr/issues/28

PR #256, opened 2025-07-24
Feat/issue 195/identification of borehole profiles

letao:
This PR addresses https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/195

We introduce a table detection system by identifying table-like layouts early in the extraction pipeline. The main functions for table detection are available in the `table_detection.py` file under `extraction/features/utils`. The method:

- Filters geometric lines by length and page boundary proximity to identify significant table-forming lines
- Categorizes lines into horizontal and vertical based on configurable angle tolerance
- Assigns confidence score to a table like structure based on area covered, amount of lines and amount of text within a structure. 
-  Focuses detection on the densest line regions

The new method is used in `extract.py` script to extract table like structures and then filter out material description pairs that fall within the table like structure.

Both `drawy.py` and `main.py` are updated to allow for the visualization of the detected tables via the new `--draw-tables` flag. Table structures are integrated into the predictions.json file. The png output is stored in a separate folder, `data/output/tables`. 

Potential limitations:

- Single table detection only, for now we only provide a single table like structure as the final output
- Multiple config parameters used at the moment that increases complexity, reduction of parameters could be helpful
- Relies on the detection of both geometrical and text lines for accuracy

### DIFF
--- a/src/extraction/features/extract.py
+++ b/src/extraction/features/extract.py
@@ -1,6 +1,7 @@
 """Contains the main extraction pipeline for stratigraphy."""
 
 import logging
+
 import pymupdf
 import rtree
 
@@ -32,7 +33,7 @@ from extraction.features.stratigraphy.sidebar.extractor.layer_identifier_sidebar
     LayerIdentifierSidebarExtractor,
 )
 from extraction.features.stratigraphy.sidebar.extractor.spulprobe_sidebar_extractor import SpulprobeSidebarExtractor
-from extraction.features.table_detection import detect_table_structures, filter_extraction_pairs_by_tables
+from extraction.features.utils.table_detection import detect_table_structures, filter_extraction_pairs_by_tables
 from extraction.features.utils.data_extractor import FeatureOnPage
 from extraction.features.utils.geometry.geometry_dataclasses import Line
 from extraction.features.utils.geometry.util import x_overlap, x_overlap_significant_smallest
@@ -55,8 +56,14 @@ class MaterialDescriptionRectWithSidebarExtractor:
     """Class with methods to extract pairs of a material description rect with a corresponding sidebar."""
 
     def __init__(
-        self, lines: list[TextLine], geometric_lines: list[Line], language: str, page_number: int, 
-        page_width: float, page_height: float, **params: dict
+        self,
+        lines: list[TextLine],
+        geometric_lines: list[Line],
+        language: str,
+        page_number: int,
+        page_width: float,
+        page_height: float,
+        **params: dict,
     ):
         """Creates a new MaterialDescriptionRectWithSidebarExtractor.
 
@@ -93,7 +100,7 @@ class MaterialDescriptionRectWithSidebarExtractor:
             geometric_lines=self.geometric_lines,
             page_width=self.page_width,
             page_height=self.page_height,
-            text_lines=self.lines
+            text_lines=self.lines,
         )
 
         logger.info(f"Detected {len(table_structures)} table structures on page {self.page_number}")
@@ -124,7 +131,7 @@ class MaterialDescriptionRectWithSidebarExtractor:
             )
             logger.info(f"After table filtering: {len(material_descriptions_sidebar_pairs)} pairs remaining")
 
-        # Step 4: remove pairs that have any of their elements (sidebar, material description) intersecting with others.
+        # Step 4: remove pairs that have any of their elements (sidebar, material description) intersecting with others
         to_delete = self._find_intersecting_indices(material_descriptions_sidebar_pairs)
         filtered_pairs = [
             item for index, item in enumerate(material_descriptions_sidebar_pairs) if index not in to_delete
@@ -148,10 +155,10 @@ class MaterialDescriptionRectWithSidebarExtractor:
             for pair in sorted(non_duplicated_pairs, key=lambda pair: pair.score_match, reverse=True)
         ]
 
-        logger.info(f"Page {self.page_number}: Extracted {len(boreholes)} boreholes from {len(table_structures)} table structures")
+        logger.info(
+            f"Page {self.page_number}: Extracted {len(boreholes)} boreholes from {len(table_structures)} tables"
+        )
         return [borehole for borehole in boreholes if len(borehole.predictions) >= self.params["min_num_layers"]]
-
-
 
     def _find_intersecting_indices(
         self, material_descriptions_sidebar_pairs: list[MaterialDescriptionRectWithSidebar]
@@ -578,15 +585,12 @@ def extract_page(
     page_rect = page.rect
     page_width = page_rect.width
     page_height = page_rect.height
-    
+
     # Detect table structures on the page
     table_structures = detect_table_structures(
-        geometric_lines=geometric_lines,
-        page_width=page_width,
-        page_height=page_height,
-        text_lines=text_lines
+        geometric_lines=geometric_lines, page_width=page_width, page_height=page_height, text_lines=text_lines
     )
-    
+
     # Extract boreholes
     extracted_boreholes = MaterialDescriptionRectWithSidebarExtractor(
         text_lines, geometric_lines, language, page_index + 1, page_width, page_height, **matching_params
--- a/src/extraction/features/extract.py
+++ b/src/extraction/features/extract.py
@@ -91,7 +91,6 @@ class MaterialDescriptionRectWithSidebarExtractor:
         # Step 1: Detect table structures on the page to guide extraction
         table_structures = detect_table_structures(
             geometric_lines=self.geometric_lines,
-            text_lines=self.lines,
             page_width=self.page_width,
             page_height=self.page_height
         )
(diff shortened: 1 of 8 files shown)


src/extraction/features/extract.py
```diff
@@ -96,7 +123,15 @@ def process_page(self) -> list[ExtractedBorehole]:
             pair for pair in material_descriptions_sidebar_pairs if pair.score_match >= 0
         ]
 
-        # remove pairs that have any of their elements (sidebar, material description) intersecting with others.
+        # Step 3: Apply table-based filtering to reduce duplicates
+        if table_structures:
+            logger.info(f"Before table filtering: {len(material_descriptions_sidebar_pairs)} pairs available")
```
TicaGit, line 128: Maybe those logs could be turned to `debug`

src/extraction/features/extract.py
```diff
@@ -96,7 +123,15 @@ def process_page(self) -> list[ExtractedBorehole]:
...
+            material_descriptions_sidebar_pairs = filter_extraction_pairs_by_tables(
+                material_descriptions_sidebar_pairs, table_structures, proximity_buffer=50
```
TicaGit, line 130: proximity_buffer value should be added to the yml file.

src/extraction/features/utils/table_detection.py
```diff
@@ -0,0 +1,411 @@
...
+    line_density: float
+
+    def contains_rect(self, rect: pymupdf.Rect) -> bool:
+        """Check if a rectangle is within this table structure."""
+        return (
+            self.bounding_rect.x0 <= rect.x0
+            and rect.x1 <= self.bounding_rect.x1
+            and self.bounding_rect.y0 <= rect.y0
+            and rect.y1 <= self.bounding_rect.y1
+        )
```
TicaGit, line 32:
```suggestion
        return (self.bounding_rect & rect) == rect

```

src/extraction/features/utils/table_detection.py
```diff
@@ -0,0 +1,411 @@
...
+        table_rect.y1 + proximity_buffer,
+    )
+
+    # Check for any overlap with expanded table area
+    return not (
+        rect.x1 < expanded_table_rect.x0
+        or rect.x0 > expanded_table_rect.x1
+        or rect.y1 < expanded_table_rect.y0
+        or rect.y0 > expanded_table_rect.y1
+    )
```
TicaGit, line 411:
```suggestion
    return not (rect & expanded_table_rect).is_empty
```

src/extraction/features/utils/table_detection.py
```diff
@@ -0,0 +1,411 @@
...
+    if len(vertical_lines) >= 2:
+        # Sort by length and position to find structural lines
+        v_sorted = sorted(vertical_lines, key=lambda line: line.length, reverse=True)
+        main_verticals = v_sorted[: min(10, len(v_sorted))]  # Top 10 longest
+
+        # Get x-coordinates of main vertical lines
+        x_positions = []
+        for line in main_verticals:
+            x_pos = (line.start.x + line.end.x) / 2
+            x_positions.append(x_pos)
```
TicaGit, line 222:
```suggestion
        x_positions = [(line.start.x + line.end.x) / 2 for line in main_verticals]
```

src/extraction/features/utils/table_detection.py
```diff
@@ -0,0 +1,411 @@
...
+
+        if x_positions:
+            refined_min_x = min(x_positions) - 10  # Small buffer
+            refined_max_x = max(x_positions) + 10.0  # Small buffer
+        else:
+            refined_min_x = initial_rect.x0
+            refined_max_x = initial_rect.x1
```
TicaGit, line 229:
```suggestion
    buffer = 10.0
    refined_min_x = min(x_positions) - buffer if x_positions else initial_rect.x0
    refined_max_x = max(x_positions) + buffer if x_positions else initial_rect.x1
```

TicaGit (changes requested):
Left some comments on code style, readability and optimization :) I did not test the functionality though.

letao:
Currently the PR adds changes suggested from https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/257 and introduces a new method for table structure detection. 

The `_find_table_regions` function implements a connected components algorithm to group lines that belong to the same table structure. It processes lines sequentially and builds groups based on connectivity analysis.

For each identified line, the function, tests connectivity against all existing groups with `_line_connects_to_group()` function that uses _line intersection_, _endpoint proximity_ and _T-junction detection_ methods. The function records which groups the line connects to in `matching_groups[]` and handles three connection scenarios based on connection count:

-  **No Connections (matching_groups = [])**
   - Creates new isolated group with just this line
   - Used for standalone lines not part of any table

- **Single Connection ([len(matching_groups) == 1]**
   - Adds line to the existing connected group
   - Standard case for extending table borders

- **Multiple Connections ([len(matching_groups) > 1]** 
   - Merges all connected groups into one larger group
   - Handles "bridge lines" that connect separate table regions
   - Processes group indices in reverse order to avoid index shifting during removal
   
Final groups are converted to table regions by separating horizontal and vertical lines, applying minimum requirements (≥2 horizontal, ≥2 vertical lines) and returning valid regions. 

The implemented uses changes from https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/257 which introduce `_pair_conflicts_with_tables` function that eliminates pairs with ambiguous positioning relative to tables and `_contained_in_table_index` function which ensures a maximum of one borehole pair per table structure.

src/extraction/features/utils/table_detection.py
```diff
@@ -0,0 +1,457 @@
...
+
+    for group_line in group:
+        # 1. Check line intersection
+        if _lines_intersect(line, group_line):
+            return True
+
+        # 2. Check endpoint proximity with Euclidean distance
+        for p1 in [line.line.start, line.line.end]:
+            for p2 in [group_line.line.start, group_line.line.end]:
+                if ((p1.x - p2.x) ** 2 + (p1.y - p2.y) ** 2) ** 0.5 <= connection_threshold:
```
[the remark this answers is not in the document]
letao: Addressed with the latest code changes

src/extraction/features/utils/table_detection.py
```diff
@@ -0,0 +1,457 @@
...
+    """Helper class for representing horizontal and vertical lines in a table structure."""
+
+    start: float
+    end: float
+    position: float
+    is_vertical: bool
+    line: Line
+
+    def connects_with(self, other: "StructureLine", tolerance: float = 10.0) -> bool:
+        # TODO docstring and unit tests
```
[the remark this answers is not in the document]
letao: Addressed with the latest code changes, unit tests added to test_geometric_line_utillities.py

src/extraction/main.py
```diff
@@ -291,6 +318,25 @@ def start_pipeline(
                 )
                 all_groundwater_entries.groundwater_feature_list.extend(groundwater_entries)
 
+                # Draw table structures if requested
+                if draw_tables:
+                    # TODO ensure that the tables drawn here are not showing up when drawing lines later on
```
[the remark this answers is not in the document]
letao: Currently I do not think that there is a problem of tables drawn showing up when drawing lines later. The implementation creates separate img variables using plot_tables() and plot_lines() functions and saves/logs the images separately as different files. As such there should not be an instance where table structures interfere with draw_lines condition. 

TODO has been removed

src/app/api/v1/endpoints/extract_stratigraphy.py
```diff
@@ -50,10 +51,20 @@ def extract_stratigraphy(filename: str) -> ExtractStratigraphyResponse:
         text_lines = extract_text_lines(page)
         geometric_lines = extract_lines(page, line_detection_params)
 
+        # Detect table structures on the page
+        # TODO avoid code duplication with main.py
```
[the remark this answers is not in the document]
letao: Current implementation moves the calculation of hight and width to the `detect_table_structures()` function. There is still the same calculation happening in the extract_page() and `detect_table_structures()` functions. I am not sure if moving this calculation to a separate function similar to something like this is worthwhile:


```Python
def compute_page_dimensions(page_index: int, document: pymupdf.Document) -> tuple[float, float]:
    """Compute the dimensions of a page in the document.

    Args:
        page_index (int): The index of the page (0-indexed).
        document (pymupdf.Document): The document containing the page.

    Returns:
        tuple[float, float]: The width and height of the page.
    """
    page = document[page_index]

    return page.rect.width, page.rect.height
```

stijnvermeeren-swisstopo:
The improvement on A11276.pdf has disappeared again. This improvement was a lucky accident, coming from the fact that the drawn tables were interfering with the detection of duplicate layers on adjacent pages using the template matching method. Now that the drawn tables are (correctly) not persisted anymore on the PDF page, the interference with the duplicate detection is gone, and the accuracy on this file is back to the status-quo. To properly fix the results for this file, I've created a follow-up ticket: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/262

stijnvermeeren-swisstopo (approved):

PR #263, opened 2025-07-31
improve handling of edge cases in geometry methods

stijnvermeeren-swisstopo:
Get rid of some warning messages from numpy, e.g. when processing Geoquat file A11276.pdf .

The fact that we had some lines of length 0, comes from the rounding of line coordinates in the `line_from_array` method, which I have now removed.

The more accurate line coordinates (floats instead of integers) lead to minor changes in the metrics.

For Zürich, we have minor improvements for 267125339-bp.pdf, 677250019-bp.pdf and 687250019-bp.pdf and a minor drop in accuracy for 680244002-bp.pdf (which is a scan of a boreprofile on [graph paper](https://en.wikipedia.org/wiki/Graph_paper), which understandably is confusing the line detection algorithms, since we've never optimized for this case).
For GeoQuat, we have a minor improvement for A8415.pdf and a minor drop in accuracy for A8236.pdf (where non-straight lines are confusing the algorithm regardless of the precision of the line detection).

Overall, we have a 0.1% improvement in F1 for Zurich and a 0.01% drop in F1 for GeoQuat. None of the individual accuracy drops are directly caused by the changes in this PR, but they rather have unrelated root causes. So I don't see any reason not to use the more accuracy line coordinates.

### DIFF
--- a/src/extraction/features/stratigraphy/sidebar/classes/a_above_b_sidebar.py
+++ b/src/extraction/features/stratigraphy/sidebar/classes/a_above_b_sidebar.py
@@ -130,6 +130,9 @@ class AAboveBSidebar(Sidebar[DepthColumnEntry]):
 
     def make_ascending(self):
         """Adjust entries in this sidebar for an ascending order."""
+        if not self.entries:
+            return self
+
         median_value = np.median(np.array([entry.value for entry in self.entries]))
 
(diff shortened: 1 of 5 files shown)


TicaGit (approved):

letao (approved):
LGTM

PR #275, opened 2025-08-27
alternative approach for fewer false positive borehole profiles

stijnvermeeren-swisstopo:


src/extraction/features/utils/table_detection.py
```diff
@@ -24,19 +25,35 @@ class TableStructure:
...
+def detect_structure_lines(geometric_lines: list[Line]) -> list["StructureLine"]:
+    """Detect significant horizonal and vertical lines in a document.
+
+    Args:
+        geometric_lines (list[Line]): Geometric lines (e.g., from layout analysis).
+
+    Returns:
+        List of detected structure lines
+    """
+    config = read_params("table_detection_params.yml")
```
TicaGit, line 37: I think we could put this at the top of the file (like we did in other files), so that we avoid duplicate calls (l.61)

TicaGit (approved):
LGTM! The approach is better and makes more sense that what I previously did. I just left a few comments; you can merge after considering them (and close the other PR  without merge).

Issue #258, opened 2025-07-29
Follow-up for table-like structures


stijnvermeeren-swisstopo:
Follow-up for https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/195

After https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/256 we have added a method for table structure detection. We need to optimize the method in two ways:

1. We need to investigate the current issues with identifying multiple boreholes on the same page. We should look at possible ways how to extend the table detection logic to identify multiple table-like borehole structures within these files (without reducing overall accuracy) or look for a specific method that handles cases where multiple boreholes without a clear box structure exist on the same page.  

   Currently the following files result in worse performance due to the new method: 

    - 268124232-bp.pdf (worse results; multiple actual boreholes in the same detected table structure)
    - 675245009-bp.pdf (worse results; multiple actual boreholes in the same detected table structure)
    - 681249142-bp.pdf (worse results; multiple actual boreholes in the same detected table structure)
    - 684252058-bp.pdf (KPI improves as one a single borehole is detected (correct) but some good material descriptions are not extracted as a consequence)
    - 690251013-bp.pdf (incorrect borehole removed because two boreholes are in the same table-structure)

   The aim is to improve accuracy on the above mentioned files in scope of the new methods we use.

2. After introduction of table detection we need to investigate potential redundancies  in the `extract.py` file to see if any of the previous checks that have been used are not needed anymore. The main aim of the table detection is to filter at most one borehole structure per table structure which should reduce the need for filtering of material description pairs. 

    The aim is to validate the current extraction logic and simplify the steps where possible without impacting the overall accuracy

Issue #276, opened 2025-08-27
The more useful table-structures should be preserved when checking for overlaps


stijnvermeeren-swisstopo:
While working on https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/275, I discovered to borehole profiles in the Geoquat validation dataset where I would have expected that a table-like structure is detected, but actually none is.

**A1290**:
The issue seems to be that two candidate table-structures are detected, the correct one, and another one that is just a rectangle around the entire scanned page. Because there is an overlap, one of the two table-like structures is rejected, and this happens to be the correct one. Afterwards, there is a condition that rejects structures with fewer than 3 vertical lines, so also the other one is rejected.

Possible solutions:
- Apply the condition for the number of vertical and horizontal lines already *before* checking for overlaps between different table structures.
- Apply a more sensible scoring / ordering of the candidate table-like structures before removing structures with an overlap

**A435.pdf**:
Also in this case, there is an incorrect table-like structure (based on scan artifacts) that is preserved when doing the overlap check, while the table-like structure that would actually be useful is rejected. The incorrect one is actually returned, and looks like this when running the script with `--draw-tables`:
<img width="971" height="2609" alt="Image" src="https://github.com/user-attachments/assets/2a45e561-1f5e-4a72-869b-89e4514e1908" />

PR #277, opened 2025-08-29
Feat/issue 258/follow up table like structures

letao:
Partially addresses https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/258.

We implement a combined filtering method that uses table structures to find potential duplicates and filter them based on score_match parameter, implementing the previous threshold. In cases where no table structures are identified we fall back on the previous logic. It creates simplification by combining the various steps into one logic.

https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/41bfbd73628a03ea277fa142659d72caba5cea77/src/extraction/features/extract.py#L117-L148

We improve on the following boreholes, where we are able to identify multiple structures:

- 268124232-bp.pdf
- 675245009-bp.pdf
- 681249142-bp.pdf 
- 684252058-bp.pdf - partial improvements (better on the material description extraction but we loose depth layer)

We reduce a bit the accuracy on: 
- (Geoquat) 13080.pdf - incorrect sidebar association
- (Geoquat) 14273.pdf - incorrect sidebar association
- (Geoquat) A11403.pdf - keyword extraction issue
- (Geoquat) A1304.pdf - additional table structure
- (Geoquat) A263.pdf - additional table structure

I believe the downsides for geoquat dataset (around -0.4%) comes from our deduplication logic, which now is removed in order to simplify and allow for multiple boreholes structures when a table structure is present. In the long run I think this might be beneficial as it allows for more flexibility. The decreased accuracy inn geoquat can be addressed with new issue or expending the solution for this ticket.

### DIFF
--- a/config/matching_params.yml
+++ b/config/matching_params.yml
@@ -5,6 +5,8 @@ left_line_length_threshold: 7
 img_template_probability_threshold: 0.62
 min_num_layers: 2
 
+material_description_column_width: 0.08
+
 depth_column_params:  # these params should be optimized as soon as there is reliable evaluation data
   noise_count_threshold: 1.25
   noise_count_offset: 2.5
--- a/src/extraction/features/extract.py
+++ b/src/extraction/features/extract.py
@@ -128,13 +128,19 @@ class MaterialDescriptionRectWithSidebarExtractor:
                     if index not in table_groups:
                         table_groups[index] = []
                     table_groups[index].append(pair)
-                else:
+
+                elif pair.score_match >= 0:
                     non_table_pairs.append(pair)
 
             # Keep pairs that meet criteria per table
             for _, pairs in table_groups.items():
-                # Filter by score threshold
-                qualifying_pairs = [p for p in pairs if p.score_match >= 0]
+                qualifying_pairs = [
+                    p
+                    for p in pairs
+                    if p.score_match >= 0
+                    and p.material_description_rect.width
+                    > self.params["material_description_column_width"] * self.page_width
+                ]
                 table_filtered_pairs.extend(qualifying_pairs)
 
             # Add all non-table pairs


letao:
With the latest changes we address some of the issues with the geoquat dataset:

- We add material description column width as an exclusion factor for material pairs (> 8% of total page width). This allows us to eliminate pairs that contain very narrow text instances that are unlikely to be true material description columns:

https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/cf964318ef103889e9331aaa54afef9960ab71e8/src/extraction/features/extract.py#L135-L144

- We add the missing score_match filter on material description pairs that are not part of a table structure:

https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/cf964318ef103889e9331aaa54afef9960ab71e8/src/extraction/features/extract.py#L125-L133

Remaining geoquat issues:

- (Geoquat) A11403.pdf - keyword extraction issue
- (Geoquat) A263.pdf - additional table structure

stijnvermeeren-swisstopo:
@letao On Zurich, I get worse results now for  684252058-bp.pdf and 267123036-bp.pdf. In both cases, it seems like we are identifying two parts of the same borehole profile as two different borehole profiles. So in these cases, neither rejecting or accepting individual borehole profiles can provide a solution, but rather, we should find a way to correctly identify the whole thing as a single borehole profile. Maybe something for a follow-up ticket?

On Geoquat, I also get worse results for A8415.pdf after merging the current main into your branch; the underlying issue seems to be that the identified table-like structures are not very good. With 11403.pdf, the issue seems to be that the table-like structure that is identified only contains part of the actual table; maybe this is related to https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/276 and can be dealt with there?

All-in-all, these changes seem to give slightly worse KPIs now on both benchmarking datasets. However, the implementation still seem to go into the right direction, as the underlying assumptions are more plausible, and that might enable more improvements later on in follow-up tickets, so I would still be ok with merging this, as long as we make sure to deal with the issues mentioned in this comment later on.

stijnvermeeren-swisstopo (approved):
Approving this PR as discussed in the stand-up. Thanks for adding the remaining edge cases to the follow-up issues!

PR #281, opened 2025-09-02
depth string removal alternative

### LINKED ISSUE #266: Remove redundant layer depths from layer descriptions
There are some cases where the depths of an extracted layer are read directly from the layer description itself. In this case, these depths should be removed from the extracted layer description, as keeping them in the layer description would just be redundant information.

For example, the description of the first layer in 10781.pdf (GeoQuat) should simply be "Bitume." (as it is correctly listed in the ground truth), not "0,00-0,07 m : Bitume." (as it is extracted now).

Image

This only applied to the depths that define the upper and lower boundary of the layer! Other depth values in the layer description should still be preserved. This includes values that might be identical to the start or the end depths of the layer, e.g. the 3.80 in this example (12306.pdf):

Image

TicaGit:
Resolves #266 

The desired change—extracting depths and removing them from the string in the same step—is only possible for the LayerIdentifier Sidebar. The difference is that, for this sidebar type, depth is not the sidebar entry, whereas it is for the others. This allows depths to be extracted together with the TextBlock, so both extraction and stripping can happen at the same time.

For the other sidebars, depth extraction does not occur in the same place as TextBlock extraction, so the steps cannot be combined. For A to B sidebars, depths are extracted at line 155 in stratigraphy/sidebar/classes/a_to_b_sidebar.py, while intervals are determined earlier by the Extractor. For A above B sidebars, TextBlocks are assigned around line 246 in a_above_b_sidebar.py, while the depths column is also handled by the Extractor.

In summary:

 - If we also want to remove depths for A to B and A above B Sidebars, we should keep the previous approach and treat all entries uniformly after extraction.

- If we only care about LayerIdentifier, then we can stick with the current approach and perform extraction and depth removal in the same place. If so, I’ll adapt the tests and clean up the code accordingly.

This is the documents affected (left, previous vs right, current PR):
image


src/extraction/features/stratigraphy/interval/a_to_b_interval_extractor.py
```diff
@@ -81,47 +87,87 @@ def from_text(cls, text_line: TextLine, require_start_of_string: bool = True) ->
...
-        input_string = text_line.text.strip().replace(",", ".")
+        input_string = text_line.text
 
         # for every character in input_string, list the index of the word this character originates from
         char_index_to_word_index = []
         for index, word in enumerate(text_line.words):
             char_index_to_word_index.extend([index] * (len(word.text) + 1))  # +1 to include the space between words
 
-        query = r"-?([0-9]+(?:\.[0-9]+)?)[müMN\]*[\s-]+([0-9]+(?:\.[0-9]+)?)[müMN\\.]*"
+        # query = r"-?([0-9]+(?:\.[0-9]+)?)[müMN\]*[\s-]+([0-9]+(?:\.[0-9]+)?)[müMN\\.]*"
```
stijnvermeeren-swisstopo, line 100: remove commented line?

src/extraction/features/stratigraphy/interval/a_to_b_interval_extractor.py
```diff
@@ -81,47 +87,87 @@ def from_text(cls, text_line: TextLine, require_start_of_string: bool = True) ->
...
-            return AToBInterval(
-                DepthColumnEntry.from_string_value(rect_from_group_index(1), match.group(1)),
-                DepthColumnEntry.from_string_value(rect_from_group_index(2), match.group(2)),
+        if depths_match:
+            return (
+                AToBInterval(
+                    DepthColumnEntry.from_string_value(rect_from_group_index(2), depths_match.group(2)),
+                    DepthColumnEntry.from_string_value(rect_from_group_index(3), depths_match.group(3)),
+                ),
+                depths_match.group(1),  # full depth match (without the begining if enable)
```
stijnvermeeren-swisstopo, line 136: Instead of returning a string here, and later reconstructing the words from the string value, could we directly return a collection of TextWord objects here? The `start_word_index` and `end_word_index` from the auxiliary method `rect_from_group_index` could probably help us select the relevant words that can later be removed from the material description.

src/extraction/features/stratigraphy/interval/a_to_b_interval_extractor.py
```diff
@@ -81,47 +87,87 @@ def from_text(cls, text_line: TextLine, require_start_of_string: bool = True) ->
...
+        if depths_match:
+            return AToBInterval(
+                DepthColumnEntry.from_string_value(rect_from_group_index(2), depths_match.group(2)), None
+            ), depths_match.group(1)
+
+        return None, None
 
-        return None
+    @classmethod
+    def strip_begining_depth(cls, line: TextLine, depths_string: str):
```
stijnvermeeren-swisstopo, line 155: Typo begining -> beginning

src/extraction/features/stratigraphy/interval/a_to_b_interval_extractor.py
```diff
@@ -47,11 +47,14 @@ def from_material_description_lines(lines: list[TextLine]) -> list[IntervalBlock
         prev_line = None
         prev_interval = None
         for line in lines:
-            a_to_b_interval = AToBIntervalExtractor.from_text(line, require_start_of_string=False)
+            a_to_b_interval, str_depth_match = AToBIntervalExtractor.from_text(line, require_start_of_string=False)
+            description_line = AToBIntervalExtractor.strip_begining_depth(line, str_depth_match)
```
stijnvermeeren-swisstopo, line 51: The original implementation only strips depths at the beginning of the entire block, not at the beginning of any line. Maybe we can preserve that behaviour? If we manage that, we would get better KPIs again for e.g. 1007.pdf.

stijnvermeeren-swisstopo (changes requested):
I prefer this approach over the original implementation. The logic is more coherent, easier to understand and hopefully also more robust and maintainable.

I don't think that it is a major issue that we only cover the IndicatorSidebar type, as with the other types, there are usually no redundant depths in the material descriptions anyway. The benchmarking seems to confirm this. The documents 10088.pdf, 10626.pdf and 5793.pdf are exceptions, but I'm happy to consider those to be edge cases and out-of-scope for this issue. (Might be good to create a follow-up issue for those. Maybe the best approach for those examples might not be quite the same as for the ones that are covered by this PR anyway.)

Refactoring the code so that the depth extraction is more consistent across different sidebar types might be something useful, but is certainly beyond the scope of this issue.

If the unit tests are fixed and my comments are addressed, then I would be happy to merge this PR.

src/extraction/features/stratigraphy/interval/a_to_b_interval_extractor.py
TicaGit: Yes very good point, I did not pay attention to that.

src/extraction/features/stratigraphy/interval/a_to_b_interval_extractor.py
```diff
@@ -81,50 +88,72 @@ def from_text(cls, text_line: TextLine, require_start_of_string: bool = True) ->
...
 
-        if match:
-            return AToBInterval(
-                DepthColumnEntry.from_string_value(rect_from_group_index(1), match.group(1), page_number),
-                DepthColumnEntry.from_string_value(rect_from_group_index(2), match.group(2), page_number),
+        def remaining_line() -> TextLine:
+            if char_index_to_word_index[depths_match.start(1)] != 0:  # group 1 is the whole depth matching
+                return text_line  # the depths found do not start the line
+            return TextLine(
+                [w for i, w in enumerate(text_line.words) if i > char_index_to_word_index[depths_match.end(1) - 1]]
```
stijnvermeeren-swisstopo, line 134: Isn't this more easily expressed without a list comprehension as 
```
text_line.words[char_index_to_word_index[depths_match.end(1) - 1] + 1:]
```
?

stijnvermeeren-swisstopo (approved):
Thanks for addressing my comments and for the further investigations. I think that this is a great first implementation, and it we could look into further improvements for the remaining edge cases later, but that does not have the highest priority for me anymore, now that this implementation already improves the extraction so much :).

I have one more comment, but I'm already approving the PR, feel free to merge as soon as you've looked at it.

PR #289, opened 2025-09-10
Feat/issue 125/better groundwater

### LINKED ISSUE #125: Evaluate the OpenCV2 implementation of the Template Matching
[Migrated from LGVISIUM-99.]

This is a follow-up ticket of LGVISIUM-63 (recognition of ground water icons using template matching). The solution in that ticket gave an improvement on the accuracy metrics, but also increased the runtime of the data extraction pipeline significantly. For that reason, we left this functionality optional (and disabled by default) for now.

One potential idea it improve the computational performance of the script would be to look into using the OpenCV2 implementation of template matching instead of the scikit-learn one.

Some people seem to claim that the former is more performant (e.g. https://www.reddit.com/r/opencv/comments/g8kdcs/question_the_speed_of_matchtemplate/).

TicaGit:
Resolves #125 

This pull request refactors and improves the groundwater extraction functionality by replacing template matching with groundwater symbol detection based on geometric line patterns. The update simplifies the detection approach and enhances data inference capabilities.

Since the ground truth contains little to no groundwater information, the impact on metrics is negligible. However, the visual results show clear improvements.

### DIFF
--- a/src/extraction/features/groundwater/groundwater_extraction.py
+++ b/src/extraction/features/groundwater/groundwater_extraction.py
@@ -121,7 +121,9 @@ class Groundwater(ExtractedFeature):
             if self.elevation is not None and terrain_elevation is not None:
                 self.depth = round(terrain_elevation - self.elevation, 2)
             else:
+                # TODO https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/293
                 # Optional method to infer groundwater depths when they are not explicitly provided.
+                #
                 # self.depth = self.infer_depth(layers, feature_rect)
                 pass
 
@@ -134,8 +136,9 @@ class Groundwater(ExtractedFeature):
     def infer_depth(self, layers: list[Layer], feature_rect: pymupdf.Rect) -> float | None:
         """Infers the depth of the groundwater feature based on the given layers and feature rectangle.
 
-        TODO This method should be reviewed with a geologist to decide whether to infer missing depths from the
-        y-scale or leave them blank. If unnecessary, it should be removed.
+        TODO https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/293
+        This method should be reviewed with a geologist to decide whether to infer missing depths from the y-scale
+        or leave them blank. If unnecessary, it should be removed.
 
         Args:
             layers (list[Layer]): The list of layers in the borehole.


stijnvermeeren-swisstopo:
@TicaGit I get an error "TypeError: '>' not supported between instances of 'NoneType' and 'datetime.datetime'" when evaluating the extraction for 268123018-bp.pdf (Zurich dataset) on this branch. Could you check if you can reproduce that, and if so, fix it?

src/extraction/features/groundwater/groundwater_extraction.py
```diff
@@ -107,26 +107,67 @@ def to_json(self) -> dict:
...
             self.elevation = round(terrain_elevation - self.depth, 2)
-        # sanity checks here to correct mistake (e.g. wrong depth extracted... )
-        prec = 0.5
-        if round(terrain_elevation - self.elevation, 2) - self.depth > prec:  # account for approx, up to 0.5m
-            logger.warning(
-                f"Extracted groundwater height informations (depth = {self.depth}, elevation = {self.elevation}) "
-                f"do not match the constraint 'terrain_elevation - gw_elevation = gw_depth': {terrain_elevation} "
-                f"- {self.elevation} = {round(terrain_elevation - self.elevation, 2)} != {self.depth} ± {prec}"
+
+    def infer_depth(self, layers: list[Layer], feature_rect: pymupdf.Rect) -> float | None:
```
stijnvermeeren-swisstopo, line 132: Do we really need this method? I don't see any positive impact on the benchmarks from it.

src/extraction/features/groundwater/groundwater_symbol_detection.py
```diff
@@ -0,0 +1,143 @@
...
+    feature_lines = [line for line in lines if line.rect.intersects(search_rect)]
+    # sort, to first treat the lines close to the symbol
+    feature_lines.sort(
+        key=lambda line: ((line.rect.x0 + line.rect.x1) / 2 - (x0 + x1) / 2) ** 2
+        + ((line.rect.y0 + line.rect.y1) / 2 - y) ** 2
+    )
+    return feature_lines
+
+
+def get_groundwater_symbol_upper_lines(
```
stijnvermeeren-swisstopo, line 43: As discussed, let's add a few unit tests for this.

src/extraction/features/groundwater/groundwater_extraction.py
stijnvermeeren-swisstopo: Thanks for the clarification. At first I though that was nothing in the log, but most likely I searched for "inferred" (correct spelling) whereas the logs us the incorrect spelling "infered".

I'm not convinced that having an approximate value is always better than extracting no value at all. At the very least, we should have some way to make it clear to the user when an extracted value comes from a precise numeric value in the document, and when it is just an approximate measurement (related to #138 ). In many cases, when no precise numeric value is available in the borehole profile, it might still be possible to look up the precise value in other documents related to the drilling. If we simply extract an approximate value without any warnings, then our experts might skip this step of looking for more precise values elsewhere, and the borehole database will not end up containing the most accurate date.

The idea of interpolating missing values based on distances on the page / PDF coordinates in itself is not a bad one. There is even potential to apply it to other attributes as well, not just groundwater depths. E.g. in some boreholes profiles, there is not even an explicit depth for all the layer boundaries. However, I think we would first need a more detailed discussion, also involving our geology experts, to decide how we really want to deal with such cases, instead of just inserting it into a PR for an issue where inferring such missing values was not a requirement or even discussed at all.

I would suggest that you remove this initial implementation from this PR, but create a follow-up issue for the inference of missing values, which we can then properly discuss and plan.

stijnvermeeren-swisstopo (approved):
I added explicit links to issue #293 in the comments: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/289/commits/7ca45e39fb4d91fd058532662b587238bd268aba

I'm happy for this PR to be merged now :)

PR #303, opened 2025-10-03
Refactoring of logic for merging boreholes across different pages

stijnvermeeren-swisstopo:


TicaGit (approved):
Nice refactoring! Apart from those few details, I think it’s a good idea to keep all this logic grouped together.

src/extraction/features/stratigraphy/layer/continuation_detection.py
```diff
@@ -0,0 +1,227 @@
...
+                    bounding_boxes=borehole_continuation.bounding_boxes,
+                )
+            _merge_boreholes(borehole_to_extend, borehole_continuation)
+
+            finished_boreholes.extend(other_boreholes_previous)
+            previous_page_boreholes = other_boreholes_current
+            previous_page_boreholes.append(borehole_to_extend)
+        else:
+            finished_boreholes.extend(previous_page_boreholes)
+            previous_page_boreholes = boreholes_on_page
```
TicaGit, line 58: For reference, this is the part I found a bit hard to follow. With quite a few variables involved, I had to focus to keep track of which one was being merged or extended. But I think that’s just the nature of the logic, not much we can simplify further.

Issue #295, opened 2025-09-23
Investigate random behaviour of file geoquat/validation/B83.pdf


TicaGit:
The metrics obtained for the file B83.pdf vary between pipeline runs, especially as code changes modify event seeds.

Sometimes, the layers in the file are not detected at all. This may be related to the order in which geometric lines are merged. The high resolution of this file results in a very large number of detected lines, which likely contributes to the inconsistent behavior.

Action items:
- Investigate the cause of this random behavior.
- Implement a patch to ensure deterministic behavior, for example by sorting the lines in a consistent order, so that layers are always detected reliably.

PR #313, opened 2025-10-14
 Feat/issue-295/random-prediction-geoquat

christianabbet:
PR closes https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/295

# Summary

The pipeline’s behavior for `B83.pdf` was erratic because langdetect is nondeterministic by default. Without a fixed seed, the same input text could yield different languages across runs. When the wrong language was predicted, downstream matching used the wrong keywords and the borehole creation failed due to missing matches.

# Root cause 

- `langdetect` relies on randomized initial state unless a global seed is set (DetectorFactory.seed).
- For the same input, different seeds produced conflicting results:

```python
# From B83.pdf
text = "B   Erde mit Steinen  Kies bis  cm DD        Sand mit Kies bis a cm fil FF  o  D o o     o    I o I I    I  "

DetectorFactory.seed = 0
detect_langs(text)  # -> [de:0.9999977789980602]

DetectorFactory.seed = 606
detect_langs(text)  # -> [en:0.5714274884515167, de:0.42856924905782434]
```

### DIFF
--- a/src/utils/language_detection.py
+++ b/src/utils/language_detection.py
@@ -1,7 +1,7 @@
 """This module contains functionalities for language detection of a document."""
 
 import pymupdf
-from langdetect import detect
+from langdetect import DetectorFactory, detect
 from langdetect.lang_detect_exception import LangDetectException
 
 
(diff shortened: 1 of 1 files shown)


stijnvermeeren-swisstopo (approved):
Thanks for figuring this out! I was able to reproduce the behaviour. Fixing the seed for reproducible results makes sense.

Issue #314, opened 2025-10-14
Non consistent tracking in MatchingParamsAnalytics


christianabbet:
# Problem
When running extraction with --matching-analytics (-ma), the generated `matching_params_analytics.json` varies between runs on the same file. Example on file `B83.pdf`

## Run A

```json
"including_expressions": { "stein": 1, "kies": 2 }
```

## Run B
```json
"including_expressions": { "stein": 1, "kies": 1, "sand": 1 }
```

I would investigate the behavior of the `MatchingParamsAnalytics` class — specifically the `find_matching_expressions` method, which is responsible for tracking matched items. 

## Expected output
```json
"including_expressions": { "stein": 1, "kies": 2, "sand": 1 }
```

PR #315, opened 2025-10-14
Feat/issue-314/matching-analytics-tracking

christianabbet:
close https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/314

# Summary

When running extraction with the --matching-analytics (-ma) option, the generated matching_params_analytics.json file was not deterministic — its content varied between runs on the same input file (e.g., B83.pdf).

This inconsistency was caused by the matching logic quitting after the first successful match, preventing subsequent matches from being tracked.
Relevant code section:

https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/cc4dd10a1b9b1c1d419c3af3cf4e0e9b5446338a/src/extraction/features/utils/text/stemmer.py#L90-L95

## Changes Introduced

- Replaced the previous matching logic with a new helper function `_match_patterns` that returns all matched keywords from the provided list.
- Ensured that all matches are now logged by the analytics, not just the first one.
- Added a dedicated unit test (`test_match_patterns) to validate the new behavior and improve robustness against future regressions.

The code is updated with a new function `_match_patterns` that return all keywords matched from the list such that they can all be logged by the analytics. Moreover a new test function s added for rubustness

## Expected output

After this update, the analytics output becomes deterministic and includes all matched keywords, as expected. For example, the output now correctly shows:

```json
"including_expressions": { "stein": 1, "kies": 2, "sand": 1 }
```

Moreover the overall metrics are unaffected on Zurich and Geoquat

### DIFF
--- a/src/extraction/features/utils/text/stemmer.py
+++ b/src/extraction/features/utils/text/stemmer.py
@@ -67,11 +67,7 @@ def _match_patterns(patterns: list[str], targets: list[str]) -> list[str]:
         list[str]: A list of patterns that were found in the targets.
         The list is empty if no matches are found.
     """
-    matches = []
-    for target in targets:
-        if target in patterns:
-            matches.append(target)
-    return matches
+    return [target for target in targets if target in patterns]
 
 
 def find_matching_expressions(
@@ -112,6 +108,6 @@ def find_matching_expressions(
     target_matches = _match_patterns(patterns, targets_to_check)
 
     # Update match traking
-    [track_match(analytics, m, language, search_excluding) for m in target_matches]
+    [track_match(analytics, match, language, search_excluding) for match in target_matches]
 
     return len(target_matches) != 0


src/extraction/features/utils/text/stemmer.py
```diff
@@ -53,6 +53,27 @@ def _split_compounds(tokens: list[str], split_threshold: float) -> list[str]:
...
+
+    Returns:
+        list[str]: A list of patterns that were found in the targets.
+        The list is empty if no matches are found.
+    """
+    matches = []
+    for target in targets:
+        if target in patterns:
+            matches.append(target)
+    return matches
```
stijnvermeeren-swisstopo: Maybe shorter and easier to read with a list comprehension?
`return [target for target in targets if target in patterns]`

stijnvermeeren-swisstopo (changes requested):
So was the non-determinism here in the end also caused by the non-determinism in the language detection, or was there an additional source of non-determinism directly related to the matching of patterns?

Issue #132, opened 2025-02-20
Avoid extracting layers that can be indirectly identified as duplicates


stijnvermeeren-swisstopo:
[Migrated from LGVISIUM-44.]

From the PR for [LGVISIUM-36](https://jira.swisstopo.ch/browse/LGVISIUM-36) (https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/49#pullrequestreview-2082310549):

The behaviour for A525 page 1-2 is somewhat suboptimal. We have a situation like

```
Page 1  duplicate detected   Page 2
 a
 b
 c      ------------------    C 
 d      ------------------    D
 e                            E
 f                            F
                              G
                              H
```
Layers "e" and "f" are included in the output for both page 1 as well as page 2, even though we could infer from the previously detected duplicates higher up ("c" and "d"), that these layers must be duplicates as well.

![Image](https://github.com/user-attachments/assets/45029671-db0e-49e8-b8db-97aee4abed59)

![Image](https://github.com/user-attachments/assets/9410ff58-b5ae-4224-863e-be6a87fceee7)

 
**Issue in borehole profiles spanning multiple pages:**
For borehole profiles that span multiple pages, the current solution does not handle sidebar alignment optimally. This issue is evident in the example deepwell Jungfraukeil_93.pdf.

In this case:
1. The last entry of the sidebar on page 2 is interpreted as the end of the sidebar, causing the **material description bounding box** to exclude some text.
2. On the subsequent page, the correct entry is located in the **middle of the page**, resulting in all text above it being omitted from the material 
description bounding box.

By using sidebar information from preceding pages, the material description bounding box can be extended to ensure all relevant text is included.

![Image](https://github.com/user-attachments/assets/bc4f0f2f-e6cd-410a-bfee-75d1aaf32292)

![Image](https://github.com/user-attachments/assets/9f928184-5baf-4106-a413-94817ef844f6)

stijnvermeeren-swisstopo:
After the refactoring to the overlap & continuation detection logic from https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/303, this seems like a good time to tackle this issue from the backlog.

[...]

Issue #304, opened 2025-10-06
Optimization-based approach for matching depths with material descriptions


stijnvermeeren-swisstopo:
In https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/285 we implemented a "dynamic programming" approach to find the best match between extracted layers and layers in the ground truth for the corresponding borehole. The idea is to find an order-preserving [zero-or-one]-to-[zero-or-one] matching that maximizes a certain scoring function.

The goals of this issue is to identify is a similar approach can be used to make the logic for splitting material descriptions and matching them with the corresponding depth intervals simpler and more robust.¨

Current, this is a fairly complex logic that looks at many factors, including distances between text lines, indentation, looking for horizontal lines that split the text, etc. The logic is quite convoluted and hard to understand/extend.

My hope is that we can treat all of the relevant features as inputs of an intuitively understandable scoring function, and then have an efficient method that finds the optimal [zero-or-one]-to-[zero-to-many] mapping between depth intervals and material description lines.

PR #317, opened 2025-10-17
Feat/issue 132/layer matching duplicates

christianabbet:
closes https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/132

When a borehole continues onto the next page, the first layers on the new page often duplicate the last layers of the previous page. The previous logic trimmed duplicates but didn’t always keep the most informative version of each layer (i.e., the one with non-empty depth fields).

This PR introduces _find_optimal_split, which scans the overlap bottom-up and shifts the cut so we keep the layer version that contains depth information. It then uses the returned indices in `merge_boreholes` to slice both sides correctly before merging.

# Visualization

Only two documents are affected by this change: Zurich 684252058-bp.pdf and Geoquat A525.pdf.

In`A525.pdf.`, the line “Sand, sauber (stellenweise leicht siltig) …” was previously matched on page 1, where the depth range was incomplete (26.00 → ?). After this update, the same text is matched on page 2, where depth information is present. Due to a local mismatch between the text anchor and the layer boundary, the range is now assigned to 30.00 → 31.00.

## Before (A525 - incomplete range)

|  Start  | End  | Description   |
|---|---|---|
| 25.1  | 26  |  Kies, sandig, siltig, einzelne Steine bis 10 cm ... |
|  - | -  |   Sand, sauber (stellenweise leicht siltig) zahlreiche ... |
| 31  | 31.9  |   Sand, leicht siltig-siltig, Komponenten der ... |

Screenshot 2025-10-17 at 13 07 03

## After (A525 - explicit range)

|  Start  | End  | Description   |
|---|---|---|
| 25.1  | 26  |  Kies, sandig, siltig, einzelne Steine bis 10 cm ... |
| 30 | 31 |  Sand, sauber (stellenweise leicht siltig) zahlreiche ... |
| 31  | 31.9  |   Sand, leicht siltig-siltig, Komponenten der ... |

Screenshot 2025-10-17 at 13 07 51

## Results 684252058-bp

A second affected document illustrates a more complex issue. In this case, two separate boreholes are detected where there should only be one. This segmentation error leads to inconsistent depth ranges and complicates the layer alignment during merging.

|  Start  | End  | Description   |
|---|---|---|
| 8.9  | 15  |  2a) tonige Phase: |
| -| -  |  t =,2.90 6.10 m: Toniger Silt m... |
| -  | -  |   2b) tonig-siltige Phase ... |
| -  | -  |   t = 6.10 - 7.40 m: leic ... |
| -  | -  |   t = 7.40 - 11.40 m: id... |
| 12  | 13  |   3. UEBERGANGSSCH ... |

|  Start  | End  | Description   |
|---|---|---|
| 8.9  | 15  |  2a) tonige Phase: |
| 5.6 | 6.1  |  t =,2.90 6.10 m: Toniger Silt m... |
| 6.1  | 7.4  |   2b) tonig-siltige Phase ... |
| 7.4  | 11.3  |   t = 6.10 - 7.40 m: leic ... |
| 11.3  | 12  |   t = 7.40 - 11.40 m: id... |
| 12  | 13  |   3. UEBERGANGSSCH ... |

## Metrics
Only the layer, depth, and material metrics are affected by this change. Since the modification impacts just two documents (684252058-bp.pdf and A525.pdf), it is not possible to draw statistically significant conclusions about overall model performance. Material F1 slightly decreases, primarily due to the detection of new layer descriptions. While this indicates that additional content is now being captured, it also introduces new false positives relative to the current layer missmatches.

|  Source  | Status  | Depth F1   | Layer F1   | Material F1   |
|---|---|---|---|---|
| Zurich  | Before  |  0.88**79** |  0.830**6** |  0.904**7** |
|    | After  |    0.88**81** |  0.830**0** |   0.904**1** |
| Geoquat  | Before  |  0.614**3** |  0.4537 |  0.6404 |
|    | After  |    0.614**4** |  0.4537 |  0.6404 |

## Recommendation

We recommend postponing further adaptation of the overlap handling for this case until the matching interface is updated to support recovery based on both depth continuity and description similarity.
Once that is available, we can:

* Enforce monotonic depth constraints during merging.
* Use description similarity to detect and unify erroneously split boreholes.
* Improve robustness of layer alignment when range data is partially missing or misordered (https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/330#).

### DIFF
--- a/src/extraction/features/stratigraphy/layer/continuation_detection.py
+++ b/src/extraction/features/stratigraphy/layer/continuation_detection.py
@@ -29,7 +29,7 @@ def _pick_merge_candidates(
             * Extended borehole from the previous page.
             * Unaffected boreholes from the previous page.
             * Extended borehole from the current page that continues it.
-            * Unaffectecd boreholes from the current page.
+            * Unaffected boreholes from the current page.
     """
     # 1) Overlap-based (returns a triple)
     borehole_to_extend, borehole_continuation, last_duplicated_layer_index = select_boreholes_with_overlap(


src/extraction/features/stratigraphy/layer/continuation_detection.py
```diff
@@ -161,6 +178,47 @@ def _is_continuation(
...
+    for i in range(n_overlap):
+        upper_layer = borehole_to_extend.predictions[-(i + 1)]
+        lower_layer = borehole_continuation.predictions[last_duplicate_layer_index - i]
+        # It is possible that the information of the upper layers is cut (), check if we can retrieve information from
+        # the lower layers
+        if not upper_layer.depth_nonempty() and lower_layer.depth_nonempty():
+            offset += 1
+            continue
+        # As soon as information is missing in the lower layer, stop iteration
+        break
```
stijnvermeeren-swisstopo: I would prefer
```suggestion
        if not upper_layer.depth_nonempty() and lower_layer.depth_nonempty():
            offset += 1
        else:
            # As soon as information is missing in the lower layer, stop iteration
            break
```
I would minimize the use of control-flow-affecting keywords such as `continue`  and `break`, as they quickly make the logic harder to understand. In this case for example, my initial reaction was "this `break` statement will always be executed at the end of the first iteration, making the whole `for` loop quite useless", because I had not immediately spotted that there was a `continue` inside an `if` clause higher up.

christianabbet:
* From what i've seen, it is quite common to find boreholes that spread over multiple pages. However I was only able to find a handful of them where the information was missing from the upper page. If only the start was present on the upper page, you could always retrieve both start and end in the lower page. I never encountered on where start was in the upper page and end in the lower page only.
   
* I did ook at the Jungfraukeil_93.pdf and I came to the same conclusion. The error comes form the missmatch bwetween the layer depths and material descriptions. I think then this goes with the current issue https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/304 ?

stijnvermeeren-swisstopo (commented):
I meant to go a little further with the refactoring of the merge logic. I took the liberty of making a commit myself directly here: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/317/commits/80d5ace044420b14169d69ba84935a1f7de4e18a . Let me know what you think about my approach.

Regarding the Jungfraukeil borehole, could you maybe create a new issue to look into the problems associated with that one?

stijnvermeeren-swisstopo (approved):

Issue #282, opened 2025-09-02
Optimize the extract.py script for the new table detection logic


letao:
After the implementation of https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/277 and https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/256  We need to follow up with an investigation into potential redundancies in the `extract.py` script to see if any of the previous deduplication checks / filtering logic is not needed anymore. 

We should think of the table detection system in the following way: the main aim of the table detection is to filter at most one borehole structure per table structure which should reduce the need for filtering of material description pairs.

- Evaluate the current approach in `extract.py`
- Remove redundant steps that do not contribute to the current extraction logic in a meaningful way
- If needed append the current table detection logic to correspond to the definition above

letao:
As the result of changes introduced with https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/277 we get worse results for 684252058-bp.pdf and 267123036-bp.pdf. In both cases, it seems like we are identifying two parts of the same borehole profile as two different borehole profiles. 

The scope of the ticket should also include a way to correctly identify a single borehole profile for documents 684252058-bp.pdf and 267123036-bp.pdf.

stijnvermeeren-swisstopo:
@letao It seems that the linked PR does not yet address the issues with 684252058-bp.pdf and 267123036-bp.pdf that are described in the previous comments. Could you create a separate issue for them?

stijnvermeeren-swisstopo:
This was closed by the PR https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/318

PR #318, opened 2025-10-20
Feat/issue 282/optimize the extractpy script

letao:
Addresses https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/282 

This PR aims to refactor the `extract.py` script and particularly  `process_page()` method into a clear filter chain to eliminate redundant steps that do not boost accuracy (F1 score) of the current test sets.

The `process_page()` is simplified into four steps:

- Find all potential pairs (layer identifiers, depth sidebars, and no-sidebar cases)
- Sort by score (highest first)
- Apply filter chain
- Create boreholes

The main simplification come from removing table grouping logic that created `table_groups` which simplifies the filtering in the following way:

-  Material description pairs are sorted in descending order and filtered by `pair.score_match >= 0` condition
- Material description pairs are filtered by `table criteria` if they are within a table structure a specific width filter is applied. For tables outside the table structures no filtering is applied
- All remaining material descriptions pairs are filtered based on intersection criteria in order to leave only the highest value intersection pair based on `pair.score_match`. This method directly checks if pairs intersect and avoids building lists.
- No special method is applied to handle cases for pairs without sidebars, since applying the previous 3 methods results in approximately the same accuracy metrics (slight improvements around .1 - .2 % against the main branch).

https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/80cdaabb4b9afd4f2f0b12f9dbab53a0337921a7/src/extraction/features/extract.py#L114-L128

The Proposed logic changes one of the goals of the issue 282: "aim of the table detection is to filter at most one borehole structure per table structure". With current filtering approach multiple boreholes can be extracted from the same table: the logic checks if each pair is contained in a table (getting the table_index), but it does not ensure only one pair per unique table. Description pairs are filtered based on specific width requirements and based on intersection. 

Currently we are still allowing for more lenient approach to description pairs that are not part of table structures. Removing this check reduces F1 scores significantly for the Geoquat dataset (~4%)

Additionally the deprecated `merge_blocks_by_vertical_spacing()` function is removed since it is not used in the extract.py script. 

Future improvements could also address simplification of `_find_all_material_description_candidates()` and `_match_sidebars_to_description_rects()` methods with aim to split them into modular components. This is avoided here to not go outside the scope of the issue.

### DIFF
--- a/src/extraction/features/extract.py
+++ b/src/extraction/features/extract.py
@@ -35,7 +35,6 @@ from extraction.features.utils.geometry.geometry_dataclasses import Line
 from extraction.features.utils.geometry.util import x_overlap, x_overlap_significant_smallest
 from extraction.features.utils.strip_log_detection import StripLog
 from extraction.features.utils.table_detection import (
-    StructureLine,
     TableStructure,
     _contained_in_table_index,
 )
@@ -61,7 +60,6 @@ class MaterialDescriptionRectWithSidebarExtractor:
         self,
         lines: list[TextLine],
         geometric_lines: list[Line],
-        structure_lines: list[StructureLine],
         table_structures: list[TableStructure],
         strip_logs: list[StripLog],
         language: str,
@@ -76,7 +74,6 @@ class MaterialDescriptionRectWithSidebarExtractor:
         Args:
             lines (list[TextLine]): all the text lines on the page.
             geometric_lines (list[Line]): The geometric lines of the page.
-            structure_lines (list[StructureLine]): Significant horizontal and vertical lines.
             table_structures (list[TableStructure]): The identified table structures of the page.
             strip_logs (list[StripLog]): The identified strip logs of the page.
             language (str): The language of the page.
@@ -88,7 +85,6 @@ class MaterialDescriptionRectWithSidebarExtractor:
         """
         self.lines = lines
         self.geometric_lines = geometric_lines
-        self.structure_lines = structure_lines
         self.table_structures = table_structures
         self.strip_logs = strip_logs  # added for future usage
         self.language = language
@@ -104,7 +100,7 @@ class MaterialDescriptionRectWithSidebarExtractor:
         Finds all descriptions and depth intervals on the page and matches them.
 
         Returns:
-            ProcessPageResult: a list of the extracted layers and a list of relevant bounding boxes.
+            list[ExtractedBorehole]: The extracted boreholes from the page.
         """
         # Step 1: Find all potential pairs
         material_descriptions_sidebar_pairs = self._find_layer_identifier_sidebar_pairs()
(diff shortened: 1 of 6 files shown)


src/extraction/features/extract.py
```diff
@@ -104,13 +103,10 @@ def process_page(self) -> list[ExtractedBorehole]:
 
         Finds all descriptions and depth intervals on the page and matches them.
 
-        TODO: Ideally, one function does one thing. This function does a lot of things. It should be split into
-        smaller functions.
-
         Returns:
             ProcessPageResult: a list of the extracted layers and a list of relevant bounding boxes.
```
stijnvermeeren-swisstopo: documented return type is incorrect

src/extraction/features/extract.py
```diff
@@ -122,142 +118,80 @@ def process_page(self) -> list[ExtractedBorehole]:
...
 
-        Returns:
-            list[int]: The indices of elements that overlap with others and should be removed.
-        """
-        to_delete = []
+        return kept_pairs
 
-        def intersects_any(rect, others):
-            return any(rect.intersects(other) for other in others if other is not None)
+    def _pairs_intersect(self, pair1, pair2) -> bool:
```
stijnvermeeren-swisstopo: We could potentially simplify this further by not comparing all combinations of sidebar and material description rect, but rather taking a bounding box around the entire pair (i.e. union of material description rect and sidebar rect, if any), and then checking for intersections between those. At first sight, only B169.pdf would be affected, and that file has other problems anyway due to the low scan quality.

Do you think that would be a good idea?

src/extraction/features/extract.py
```diff
@@ -122,142 +118,80 @@ def process_page(self) -> list[ExtractedBorehole]:
...
+                    filtered.append(pair)
+
+        return filtered
+
+    def _filter_by_intersections(
+        self, pairs: list[MaterialDescriptionRectWithSidebar]
+    ) -> list[MaterialDescriptionRectWithSidebar]:
+        """Remove pairs that intersect with higher-scoring pairs."""
+        if not pairs:
+            return []
```
stijnvermeeren-swisstopo: Probably not really necessary, as if `pairs` is an empty list, the return value will be an empty list as well, even without these two lines?

src/extraction/features/extract.py
letao: Agreed that would make the approach simpler. I can propose the union approach in the following way:

```python
  def _pairs_intersect(self, pair1, pair2) -> bool:
      bbox1 = pair1.material_description_rect
      if pair1.sidebar:
          bbox1 = bbox1 | pair1.sidebar.rect

      bbox2 = pair2.material_description_rect
      if pair2.sidebar:
          bbox2 = bbox2 | pair2.sidebar.rect
          
      return bbox1.intersects(bbox2)   
```
I do not see any negative impact on the metrics with this approach so I would agree that we go for this alternative

PR #319, opened 2025-10-21
refactor/issue-304/dp-interval-blocks

### LINKED ISSUE #304: Optimization-based approach for matching depths with material descriptions

TicaGit:
Resolves #304 

### Introduction
This PR introduces a new dynamic programming approach for pairing text lines with interval zones.
Instead of first grouping text lines into blocks and then assigning those blocks to intervals, the new method directly assigns individual lines to intervals in one unified optimization step.

Each text line receives a pairing score with every interval, which reflects  geometric compatibility and depends on the sidebar type detected. In addition, lines have a pairwise affinity term with the line above them, capturing spatial relationships:
- Negative affinity discourages grouping lines. (e.g., a horizontal line separating two lines).
- Positive affinity encourages grouping lines (e.g. lines that slightly overlap vertically).

The dynamic program maximizes the total pairing score + affinity reward while preserving order,
ensuring that contiguous, compatible lines are grouped under the same interval zone in an interpretable and deterministic way.

### Results
The results on the datasets are close, but slighlty worse than before. (New results are in **bold**)
Geoquat/val: 
  - Layer f1: 0.454 -> **0.462**
  - Depth interval f1 :0.614 -> **0.618**
  - Material description f1: 0.640 -> **0.628**
Zurich:
  - Layer f1: 0.831 -> **0.822**
  - Depth interval f1 :  0.888 -> **0.894**
  - Material description f1: 0.905 -> **0.869**

The main problem on geoquat, is that most documents have sidebar that are missing some entries, this causes the matching to be off and lowers the results. Better ocr and better sidebar assignement would increase the metrics  on geoquat
Overall, it was hard to find the balance between loss function (text lines should be in front of the intervals) and affinity (line affinities should be trusted and we should really avoid putting two lines having low affinity in the same block). This fact causes some documents that had a perfect match before to be totally wrong.

### PR details:
Lots of function and methods were deleted, with their logic reimplemented in a way easier to follow. Particularly, the PR:
- replaces the function `match_columns` that was using `sidebar.identify_groups` and `transform_groups` by a DP matching algorithm.
-  replaces `get_description_blocks` by `get_pairs_based_on_line_affinity` which forms blocks based on line affinity.
- deletes `IntervalBlockGroup`, which I found was a confusing class.
- deletes all `matching_blocks` and `identify_groups` implementations in sidebar classes, and condense their logic into the `dp_scoring_fn`method. The method of sidebar `post_processing` also implements part of the logic that was done after the matching.
-  implements `get_interval_zone`, a sidebar method that build zones for each interval (consisting of a start and end). This is needed for sidebars like AtoB, whose intervals only contain the location of the entry (often containing both start and end).
- deletes `merge_blocks_by_vertical_spacing` and `split_blocks_by_textline_length`.
- deletes all `DescriptionBlockSplitter` and `split_based_on_indentation`. Their logic is implemented in the textline_affinity.py module.

### My take on this PR:
I feel like the logic is less intricate and easier to follow, but some forcing features of the previous implementation are lost. For example, we lost the fact that a number of blocks corresponding exactly to the number of interval would match them directly. Explicit logic like those would achieve a better mapping than the DP which tries to maximise a score, thus sometimes weighting more a decision that don't make sense for us.
Maybe a hybrid approach that reintroduce some heuristic could be could.

### Others
- currently, all lines are assigned to intervals. It might be worth looking if allowing non matched at the end could improve results (e.g. 266126001, 267123080). An option could be to allow an open-ended interval at the end, and add a penalty to the pairing scores of lines to this interval, to discourage its use.

### DIFF
--- a/src/extraction/features/stratigraphy/sidebar/classes/spulprobe_sidebar.py
+++ b/src/extraction/features/stratigraphy/sidebar/classes/spulprobe_sidebar.py
@@ -3,7 +3,7 @@
 from typing import ClassVar
 
 from extraction.features.stratigraphy.base.sidebar_entry import SpulprobeEntry
-from extraction.features.stratigraphy.interval.interval import IntervalZone
+from extraction.features.stratigraphy.interval.interval import Interval, IntervalZone
 from extraction.features.stratigraphy.sidebar.classes.sidebar import Sidebar
 from extraction.features.utils.text.textline import TextLine
 
@@ -32,9 +32,15 @@ class SpulprobeSidebar(Sidebar[SpulprobeEntry]):
     def get_interval_zone(self) -> list[IntervalZone]:
         """Get the interval zones defined by the sidebar entries.
 
-        The interval zones are created from the AToBInterval entries, filtering out sublayers and invalid layers.
+        The interval zones are created from the Spulprobe entries, including open-ended intervals at both ends.
 
         Returns:
             list[IntervalZone]: A list of interval zones.
         """
-        return self.get_zones_from_entries(self.entries, include_open_ended=True)
+        open_started = IntervalZone(None, self.entries[0].rect, Interval(None, self.entries[0]))
+        zones = [
+            IntervalZone(entry.rect, next_entry.rect, Interval(entry, next_entry))
+            for entry, next_entry in zip(self.entries, self.entries[1:], strict=False)
+        ]
+        open_ended = IntervalZone(self.entries[-1].rect, None, Interval(self.entries[-1], None))
+        return [open_started] + zones + [open_ended]


src/extraction/features/utils/text/textline_affinity.py
```diff
@@ -0,0 +1,317 @@
...
+
+        Returns:
+            float: The affinity: -1.0 if lines are not compatible, 0.0 otherwise.
+        """
+        return max(-1.0, min(0.0, -(current_line.rect.x1 - previous_line.rect.x1) / current_line.rect.width))
+
+    def _indentation_affinity(self, previous_line: TextLine, current_line: TextLine):
+        """Split the text block based on indentation.
+
+        note: not currently used. is_indented attribute was removed from TextLine, put back if needed.
```
stijnvermeeren-swisstopo: Create follow-up issue to look for a proper was to make use of indentation, or remove these methods if we decide that this is not necessary. @stijnvermeeren-swisstopo check in LGD-288 for examples.

Issue #308, opened 2025-10-08
Improve the strip log confidence score approach


letao:
After the initial solution to https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/280 outlined in https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/301 We need to improve the scoring function for strip-log selection. 

Currently we combine strip log aspect with circle and horizontal line density:

https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/dc11bc37cc35a07432e5e3b5af1669822f33ff24/src/extraction/features/utils/strip_log_detection.py#L232-L244

While it is a good initial result we should aim to use a function that keeps on improving as the circle density grows. As an example we can think of a sigmoid function that we help to differentiate between a good strip log and a really good one, currently the score would simply cap out at the max value for the circle and line density and we loose the ability to differentiate between good and better strip logs.

Additional point to consider Instead of using simple density, we can also reward a larger overall area in the score. If the bottom half of the strip log has a slightly lower density than the top half, then we still want to award a higher confidence to the overall strip log compared to just the top half.

Issue #306, opened 2025-10-07
Improvements to the strip-log detection algorithm


letao:
After the initial solution to https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/280 outlined in https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/301 We need to improve the method of using vertical lines for strip log detection outline ind the `strip_log_detection.py` file.

Current approach is not optimal since we do not control for horizontal alignment between the potential vertical edges of the strip-log. The next couple of lines that we are iterating over could also be above or below the strip log, and the right-hand edge of the strip log could come later on.

https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/0e1b7e4bbbde9a384a8383b9a32d6f1dd94d0e2c/src/extraction/features/utils/strip_log_detection.py#L87-L99

Additionally we should explore the possibility of connecting multiple vertical lines that form one strip-log but that are split due to line extraction algorithm, this would allow to detect the correct strip-log shape and avoid splits like in the case of `267125047-bp.pdf`

While looking at this we should also be mindful of computational complexity that the current method can introduce in files where we have many vertical lines,

PR #328, opened 2025-10-27
Feat/issue 306/striplog detection and scoring improvement

christianabbet:
Closes https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/306 and closes https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/308

 ## Summary

This PR introduces a new end-to-end strip-log detector that builds vertical “striplogs” from page geometry in PDFs. The pipeline:

- detects table/column cells and vertical structure lines from a binarized page,
- scores each cell by visual/text features,
- merges high-confidence, vertically aligned and connected cells into strip logs via graph clustering.

This PR replaces that logic with a robust cell-centric approach that: builds a grid via morphology, labels background cells and vertical line components, uses graph connectivity constrained by vertical alignment, and combines multiple visual/text signals into a confidence score.

Tests on both Zurich and geoquat show promissing results. The absence of validation data makes it difficult to quantify it. However, big change in geoquat detection rate.

bitmap

## Packages

This PR adds scikit-image as a required dependency. It is used to support the region properties method [skimage.measure.regionprops](https://scikit-image.org/docs/0.23.x/api/skimage.measure.html#skimage.measure.regionprops) that is not available in OpenCV.

## Metrics / Scoring signals

Configuration allows toggling each term. Best results are achieved using `pattern` and `text`.

**Crowding score:** density of dilated foreground (local foreground density).
**Pattern score**: Hough circles (area fraction) + LSD line “area” (width × length), combined and clipped. 
**Text**: Check if text is present based on multiple factor such as text length and area. 

## Graph connectivity

We build an undirected adjacency matrix among graph nodes. For each node that also qualifies as a seed, we connect it to other nodes if all are true:

- **aligned**: Section are vertically aligned and show similar width ratio (relative width error and x0 error both < r_tol_width).
- **line-connected**: There exists a vertical structure bbox that connects sections.

We symmetrize the adjacency and cluster via connected components. This enforces coherent vertical stacks with a shared border.

## Drawing / Visualization

Each StripLog is drawn with an outer bounding box (union of sections). Its sections are drawn as filled rectangles. Color encodes the strip-log id (each cluster a different color); sections within a strip share that color with fill.

## Time

The chart shows processing time per dataset and per document. No striplog omits strip-log computation entirely, Striplog (Old) uses the previous approach, and Striplog (New) uses the proposed approach.

| Task                | Dataset | Overall [s] | Striplog [s] | Striplog [s/doc] | 
|---------------|---------|------------|------------|------------------| 
| No striplog     | Zurich    |            183 |                  - |                         - | 
|                        | Geoquat |           196  |                  - |                         - | 
| Striplog Old   | Zurich    |           210  |               27  |                0.181  | 
|                        | Geoquat |           223  |              27  |                0.108  | 
| Striplog New  | Zurich    |           205  |               22 |                  0.137 | 
|                        | Geoquat |            221  |              25 |                 0.100 | 

### Partial detections 
The main issue arises from striplogs that are only partially detected. One identified cause is that when a horizontal line crosses the striplog, the model mistakenly interprets it as two distinct cells. 
As a result, the detected sections become too narrow to be recognized as valid striplogs. However, this may not be the only reason for such failures.
Examples: 267123021, 267123041, 267123094, 267125045, 677250019, 1488, 2537, 979, A11112, A11403, A11436, 267123002, 267125045, 267125047

### False negatives
Some files exhibit false negatives, where the model fails to detect a striplog even though one is clearly visible.
Examples: 267125311, 268124635, 267124097, 267125308, A11068, 267123088, 267123089

### False positivies 
Another common issue is the presence of false positives. This often occurs in files containing large regions with very small cells and/or poor exposure. In such cases, the thresholding 
step may fail to properly adapt to lighting variations, producing spurious patterns in areas that should be mostly empty.
Examples: 267124133, 267124282, 267125223, 677229003, 680244002, 683251058, A108, A11193

bitmap

### DIFF
--- a/src/extraction/annotations/draw.py
+++ b/src/extraction/annotations/draw.py
@@ -4,10 +4,11 @@ import logging
 import os
 from pathlib import Path
 
+import numpy as np
 import pymupdf
 from dotenv import load_dotenv
 
-from extraction.annotations.plot_utils import convert_page_to_opencv_img
+from extraction.annotations.plot_utils import convert_page_to_img
 from extraction.features.predictions.file_predictions import FilePredictions
 from extraction.features.predictions.overall_file_predictions import OverallFilePredictions
 from extraction.features.stratigraphy.layer.layer import Layer
@@ -319,13 +320,16 @@ class PageDrawer:
             )
 
 
-def plot_tables(page: pymupdf.Page, table_structures: list[TableStructure], page_index: int):
+def plot_tables(page: pymupdf.Page, table_structures: list[TableStructure], page_index: int) -> np.ndarray:
     """Draw table structures on a pdf page.
 
     Args:
         page:               The PDF page.
         table_structures:   The identified table structures on the page.
         page_index:         The index of the page in the document (0-based).
+
+    Returns:
+        np.ndarray: The plotted table.
     """
     temp_doc = pymupdf.open()
     temp_doc.insert_pdf(page.parent, from_page=page_index, to_page=page_index)
@@ -335,7 +339,7 @@ def plot_tables(page: pymupdf.Page, table_structures: list[TableStructure], page
     draw_table_structures(shape, temp_page.derotation_matrix, table_structures)
     shape.commit()
 
-    result = convert_page_to_opencv_img(temp_page, scale_factor=2)
+    result = convert_page_to_img(temp_page, scale_factor=2)
 
     temp_doc.close()
 
@@ -430,13 +434,16 @@ def draw_strip_logs(shape: pymupdf.Shape, derotation_matrix: pymupdf.Matrix, str
             )
 
 
-def plot_strip_logs(page: pymupdf.Page, strip_logs: list[StripLog], page_index: int):
+def plot_strip_logs(page: pymupdf.Page, strip_logs: list[StripLog], page_index: int) -> np.ndarray:
     """Draw strip log structures on a pdf page.
 
     Args:
         page: The PDF page.
         strip_logs: The identified strip log structures on the page.
         page_index: The index of the page in the document (0-based).
+
+    Returns:
+        np.ndarray: The plotted striplog.
     """
     temp_doc = pymupdf.open()
     temp_doc.insert_pdf(page.parent, from_page=page_index, to_page=page_index)
@@ -446,7 +453,7 @@ def plot_strip_logs(page: pymupdf.Page, strip_logs: list[StripLog], page_index: 
     draw_strip_logs(shape, temp_page.derotation_matrix, strip_logs)
     shape.commit()
 
-    result = convert_page_to_opencv_img(temp_page, scale_factor=2)
+    result = convert_page_to_img(temp_page, scale_factor=2)
 
     temp_doc.close()
 
--- a/src/extraction/annotations/plot_utils.py
+++ b/src/extraction/annotations/plot_utils.py
@@ -21,8 +21,19 @@ if mlflow_tracking:
 logger = logging.getLogger(__name__)
 
 
-def _draw_lines(open_cv_img, lines, scale_factor=1):
+def _draw_lines(img: np.ndarray, lines: list[Line], scale_factor: int = 1) -> np.ndarray:
+    """Draw lines on image.
+
+    Args:
+        img (np.ndarray): Input image to draw on.
+        lines (list[Line]): Lines to draw
+        scale_factor (int, optional): Scaling factor. Defaults to 1.
+
+    Returns:
+        np.ndarray: Lines plotted on image.
+    """
     grid_lines = [_convert_line_to_grid(line, scale_factor=scale_factor) for line in lines]
+    img = img.copy()
     for line_count, line in enumerate(grid_lines):
         color = (
             (255 - 5 * line_count) % 255,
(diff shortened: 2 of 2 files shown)


stijnvermeeren-swisstopo (changes requested):
Overall, the results look quite good, certainly more robust then with the previous iteration.

I have two main concerns.

**1. Running time**

The new approach is significantly slower, and this does not seem to be sufficiently justified to me. You mention that pages are now rendered in two different resolutions. Why is this necessary? If the different resolution works better for the strip log detection, wouldn't it likely to be better as well for the line & table detection?

Also, did you verify that this is the really cause of the slowdown? I.e. if you process everything at the same dpi, is the running time then the same as before, or could there be other factors that contribute to the slower running time?

**2. Lack of understanding about the "crowding score"**

I find it hard to understand what the crowding score does exactly. It's only documented as "foreground density after dilation" which does not really make things much clearer to me. The crowding score seems to be the factor that decides why certain sections of the strip log below (from Geoquat 13080.pdf) are detected, and others are not. I would like to have a more intuitive explanation for this behaviour, but I'm struggling to pinpoint the reason why some sections are detected and others aren't.

Image

christianabbet:
Here is an update on the latest commits:

The use of two different resolutions was originally due to table detection performing better with specific hyperparameters. After further tuning, both table extraction and section feature extraction now work reliably at the same resolution. Consequently, the processing time for the new approach matches that of the previous one (see updated table in the PR).

The definition of the crowding score has been expanded for clarity. It now reads:
“Computes how visually crowded a binary image region is by measuring the average pixel density after applying a morphological dilation. The dilation expands foreground regions using a circular kernel to capture nearby pixel clusters and fill small gaps. Higher scores indicate denser or more cluttered regions, while lower scores reflect sparser content.”

stijnvermeeren-swisstopo (changes requested):
**Execution**

The execution time of the pipeline still seems slightly slower than before on my virtual machine, but it's no longer a significant/problematic slowdown, so I'm happy to accept this.

**Strip log extraction quality**

The new behaviour seems to be a clear improvement on the old one. I'm particularly pleased with files such as 681249142-bp.pdf, where we now have a nice one-to-one correspondence between the different boreholes and the identified strip logs.

Image

When there are horizontal lines that go all the way across the strip log within a certain section, then the current method is a bit too aggressive in splitting the sections. See e.g. 2537.pdf. However, the strip log as a whole is still identified correctly here, so it's not a major problem for now.

image

There's a few cases where the old method was able to identify a strip log that the new method misses, e.g. A443.pdf and 267124097-bp.pdf. However, these are few and far in between, so I don't think it's necessary to look too much into those at this time.

There's a couple of false positives, but not more than with the old method. An interesting one is 267124133-bp.pdf, where part of a depth column is misidentified as a strip log. Is the penalty for text in the strip log applied correctly here?

Image

It also happens quite often that vertically oriented text was not detected by the OCR, and the strip log detection fails because of that. Usually (e.g. B626.pdf) it's just a single "header" section with a vertically placed word such as "Profil" that is incorrectly included in the strip log. Occasionally it's an entire strip log that is misidentified, such as in A359.pdf. Ideally, this would be fixed already in the OCR. However, maybe we can still make the strip log detection more robust by e.g. looking at how "uniformly" a strip log section is filled? If the section is mostly empty, but with a high concentration of pixels/lines in a smaller area, then we could suspect that this is actually text? I would propose to create a follow-up ticket to look into this idea.

Image

**Requested change**

For non-grayscale PDF files, the colours are incorrect in all the `*_strip_logs.png` visualizations (when running the pipeline with `-sl`. See e.g. `268124297-bp.pdf` for an example. This seems to be already the case on the `main` branch as well, so if the problem is difficult to find/fix, then we could also move it to a follow-up ticket.

Image 

Image

src/extraction/features/utils/strip_log_detection.py
```diff
@@ -1,287 +1,635 @@
...
+    text_area_within = np.sum([(line.rect & rect).get_area() for line in text_within_ocr])
+    text_area_outside = np.sum([(line.rect).get_area() for line in text_within_ocr])
+    ocr_area_factor = text_area_within / text_area_outside
+
+    # Compute portion of area that is overlapping with section
+    score_area = 1 - text_area_within / rect.get_area()
+
+    # make the assumption that if a text overlap with section, it has a linear contribution. For eaxmple, if
+    # "this is a text!" overlap with section at 50% (e.i. ocr_area_factor = 0.5) then only "a text!" is kept.
+    score_length = np.min([np.exp(-len(line.text) * ocr_area_factor / tau) for line in text_within_ocr])
```
stijnvermeeren-swisstopo: Why apply this per line, and the take the minimum? Wouldn't it make more sense to sum all the line lengths inside the exponential, or even better, to also compute and apply the ocr_area_factor for each line separately?

stijnvermeeren-swisstopo (changes requested):

stijnvermeeren-swisstopo (approved):
@christianabbet LGTM now!

Did you consider creating a follow-up ticket for this comment?

> It also happens quite often that vertically oriented text was not detected by the OCR, and the strip log detection fails because of that. Usually (e.g. B626.pdf) it's just a single "header" section with a vertically placed word such as "Profil" that is incorrectly included in the strip log. Occasionally it's an entire strip log that is misidentified, such as in A359.pdf. Ideally, this would be fixed already in the OCR. However, maybe we can still make the strip log detection more robust by e.g. looking at how "uniformly" a strip log section is filled? If the section is mostly empty, but with a high concentration of pixels/lines in a smaller area, then we could suspect that this is actually text? I would propose to create a follow-up ticket to look into this idea.

Issue #310, opened 2025-10-09
Integration between swissgeol-boreholes-dataextraction and swissgeol-assets-dataextraction


letao:
As indicated here: https://github.com/swisstopo/swissgeol-assets-dataextraction/pull/56#pullrequestreview-3310950383, we should integrate swissgeol-boreholes-dataextraction into the assets repository as a dependency for this project. 

The aim should be to reuse as much as possible the code from swissgeol-boreholes-dataextraction without needing to manually copy over code elements and functions (e.g.t able_detection.py or geometric_objects.py). 

With this we should make it  easier to benefit from ongoing improvements such as the extraction of borehole names (https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/300) that are also relevant to the data extraction from assets. 

As part of this issue we should make swissgeol-boreholes-dataextraction and its methods available in the swissgeol-assets-dataextraction context.

stijnvermeeren-swisstopo:
Take a couple of hours to establish a clear long-term vision. Two options:
- Assets data extractions depends on the full boreholes data extraction codebase
- There's a "core" library with common code that is imported by both assets as well as boreholes data extraction

Additionally, it might be better to create a separate issue for the changes required in the swissgeol-assets-dataextraction repo, and use this issue only for the changes required on the swissgeol-boreholes-dataextraction side.

PR #332, opened 2025-10-31
Fix/issue 321/improve affinity split

### LINKED ISSUE #321: Improve affinity-based splitting of material descriptions without depths
### Description
After PR #304, the new DP approach appears to perform worse on layouts without any extracted depths.
This issue is not related to the DP, but rather to how the function `get_pairs_based_on_line_affinity` creates text blocks in `extract.py`.

 It would be good if we can go back to a behaviour that is more similar to the old approach for such cases, and to tweak the block formation based on affinity.

### Example Documents
The issue has been observed in the following files:

267123120-bp.pdf  
A108.pdf  
A19.pdf  
A253.pdf  
B140.pdf  
B363.pdf  
V69.pdf

TicaGit:
Resolves #321 

### Introduction

This PR adds a new affinity metric, gap_with_previous_affinity, to improve text line grouping when no sidebar is detected in the document. It compares the vertical gap between consecutive lines and applies a penalty when the gap significantly increases compared to the previous one.
The idea behind this heuristic is that the gap size between lines in the same text block should remain consistent. Therefore, lines with a gap size similar to the previous one likely belong to the same block and are given a small bonus.

This new affinity is weakly weighted and combined with the vertical spacing to determine block splits.

Additionally:
The PR updates the line comparison logic to use more precise vertical positions (two-thirds and one-third points instead of midpoints).

### Metrics
The impact on the metrics is negligible (less than 1%) across both datasets.

Among the files mentioned in the issue:

- 267123120-bp.pdf improves: lines are grouped correctly.
- A108.pdf improves a lot: lines are grouped correctly.
- A19.pdf slightly decrease, but the grouping make sense.
- A253.pdf slightly decrease, but the grouping is better now, the problem is caused by some parasite text lines (single `-`).
- B140.pdf improves a lot:  lines are grouped correctly.
- B363.pdf  improves: lines are grouped correctly.
- V69.pdf slightly decrease, all lines are close, and grouping is ambiguous

### Difficulty encountered
The main difficulty with this issue is that without sidebar context, it is hard to determine whether lines belong together. For example, in A19 and B363, the distances between lines within a block vary between blocks, so we cannot rely solely on the smallest gap in the document for grouping.

Other specific cases causing issues:
- On 675246002, there are helper lines linking text lines to intervals, but these lines connect the middle of the text line to the middle of the interval, rather than its extremity as is usually the case (thus lines_on_the_left_affinity can sometimes be incorrect).

- On A359, A253, 681249142 (and others), some parasite text lines (like 2 or -) are extracted, which breaks the logic of analyzing previous line gaps.

- On 680244005, the page has a grid background, so relying on horizontal lines to split blocks is not reliable.

### DIFF
--- a/src/extraction/features/utils/text/textline_affinity.py
+++ b/src/extraction/features/utils/text/textline_affinity.py
@@ -169,8 +169,8 @@ class LineAffinityCalculator:
         Returns:
             float: The affinity: -1.0 if lines are not compatible, 0.0 otherwise.
         """
-        previous_line_y_two_thirds = previous_line.rect.y0 + (previous_line.rect.y1 - previous_line.rect.y0) * 2 / 3
-        current_line_y_one_third = current_line.rect.y0 + (current_line.rect.y1 - current_line.rect.y0) / 3
+        previous_line_y_two_thirds = previous_line.rect.y0 + previous_line.rect.height * 2 / 3
+        current_line_y_one_third = current_line.rect.y0 + current_line.rect.height / 3
         for line in self.lines_spanning_description:
             line_y_coordinate = (line.start.y + line.end.y) / 2
             if previous_line_y_two_thirds < line_y_coordinate < current_line_y_one_third:
@@ -187,8 +187,8 @@ class LineAffinityCalculator:
         Returns:
             float: The affinity: -1.0 if lines are not compatible, 0.0 otherwise.
         """
-        previous_line_y_two_thirds = previous_line.rect.y0 + (previous_line.rect.y1 - previous_line.rect.y0) * 2 / 3
-        current_line_y_one_third = current_line.rect.y0 + (current_line.rect.y1 - current_line.rect.y0) / 3
+        previous_line_y_two_thirds = previous_line.rect.y0 + (previous_line.rect.height) * 2 / 3
+        current_line_y_one_third = current_line.rect.y0 + current_line.rect.height / 3
         for line in self.long_horizontals:
             line_y_mid = (line.start.y + line.end.y) / 2
             geom_line_is_in_between = previous_line_y_two_thirds < line_y_mid < current_line_y_one_third


src/extraction/features/utils/text/textline_affinity.py
```diff
@@ -149,11 +169,11 @@ def compute_long_lines_affinity(self, previous_line: TextLine, current_line: Tex
         Returns:
             float: The affinity: -1.0 if lines are not compatible, 0.0 otherwise.
         """
-        last_line_y_coordinate = (previous_line.rect.y0 + previous_line.rect.y1) / 2
-        current_line_y_coordinate = (current_line.rect.y0 + current_line.rect.y1) / 2
+        previous_line_y_two_thirds = previous_line.rect.y0 + (previous_line.rect.y1 - previous_line.rect.y0) * 2 / 3
+        current_line_y_one_third = current_line.rect.y0 + (current_line.rect.y1 - current_line.rect.y0) / 3
```
stijnvermeeren-swisstopo: `current_line.rect.y1 - current_line.rect.y0` is simply `current_line.rect.height`, using the latter would make the code easier to understand.

src/extraction/features/extract.py
```diff
@@ -600,10 +600,11 @@ def get_pairs_based_on_line_affinity(
     """
     pairs = []
     prev_block_idx = 0
-    # reintroduce the heuristic that the presence of >=3 horiz. lines should tighten the vertical spacing constrain
+    # The presence of >=3 horiz. lines should tighten the vertical spacing constrain
     threshold = -0.99 if sum(affinity.long_lines_affinity for affinity in affinities) <= -3.0 else 0.0
     for line_idx, affinity in enumerate(affinities):
-        if affinity.total_affinity() < threshold:  # note: the affinity of the first line is always 0.0
+        # note: the affinity of the first line is always 0.0
+        if affinity.weighted_affinity(1.0, 1.0, 1.0, 1.0, 0.2) < threshold:
```
stijnvermeeren-swisstopo: why these weights?

TicaGit: We previously used `total_affinity`, which is an unweighted sum. Since I wanted to keep all affinities within the range of -1 to 1 while reducing the influence of the last affinity (the new gap affinity), I weighted its impact there. The small weight (0.2) ensures that this affinity doesn’t overpower the vertical distance affinity but still helps resolve tie situations by slightly favoring lines with smaller gaps than the previous.
I tried few other values, and I think this one works the best.

stijnvermeeren-swisstopo: @TicaGit can you add this as a comment in the code? Otherwise these values come "out of the blue" for someone who is reading the code.

TicaGit: The values will be added to a YAML file in the PR that closes issue #310 (currently in progress).

PR #334, opened 2025-11-04
Feat/issue 324/diagonal pairing

### LINKED ISSUE #290: Improve the logic for splitting text blocks based on line length
After analyzing https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/135 we have identified that the incorrect material description identification in A531.pdf is due to the text line splitting logic defined in the `split_blocks_by_textline_length()` function from `extract.py` script. 

Image

The function provides a simple way of splitting text blocks based on the line lengths, using the right edge position (x1 coordinate) of text lines as splitting indicators. The underlying assumption is that shorter lines often indicate natural break points in material descriptions. The logic does not always work perfectly as in the case of A531.pdf. 

https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/543668009d57f7b0b460ab646da6167206e4d107/src/extraction/features/extract.py#L754-L791

We should aim to improve this method in order to better handle issues similar to A531.pdf, incorrectly combining multiple layer descriptions into one due to no clear break points identified. In order to address this we can imagine two approaches:
- A more context aware splitting algorithm, e.g., allowing for larger lines + potential description ending indications to work together in defining a splitting point. 

- Utilizing the diagonal identification lines for material descriptions. If these can be identified they could serve as potential demarcation lines for material description splits:

Image

Alternative solutions can also be tested. In the end we should improve the accuracy metrics for A531.pdf without severely impacting other files.

### LINKED ISSUE #324: Improve scoring for profiles with diagonal misalignments and special borehole layout cases
### Description
After the changes in PR #304, a noticeable loss of accuracy has been observed in certain profiles. The issue seems tocome from how vertical distance scores and affinity bonuses interact in cases with many diagonal connections.

When profiles contain many misaligned intervals and as the misalignment gets bigger, the vertical distance penalty tend to outweigh the affinity bonuses causing lines to be assigned to the wrong interval. Increasing the weight of affinity scores does not resolve the problem, as it negatively impacts other document inputs.

Additionally, the new scoring approach no longer allows extracting additional material description lines that are not linked to a depth interval which was sometimes beneficial, especially when OCR misses depth values. The current DP approach forces lines to be assigned to an interval, even if the pair scoring is very bad.

### Possible Directions

- Detect profiles with significant diagonal misalignment and dynamically rebalance the weight of vertical vs. affinity scores.
- Explore heuristic or adaptive weighting strategies based on alignment consistency.
- Reintroduce (or conditionally enable) extraction of material descriptions not linked to depth intervals.

### Example Documents

The issue has been observed in the following profiles:

267124070-bp.pdf
267125058-bp.pdf
267125439-bp.pdf
268124256-bp.pdf
268124569-bp.pdf
269126084-bp.pdf
675244002-bp.pdf
A359.pdf
A42.pdf
A8561.pdf
A8542.pdf

### Follow-up issue: Special Layout Case

A related situation occurs in A11403.pdf and A11481.pdf.
These represent a nearly new layout type, which the old method handled correctly (because the number of blocks exactly matched the number of intervals), while the new method struggles with them.

It may be worth investigating whether these can be addressed together with the diagonal misalignment examples, or whether custom logic is needed for this borehole profile layout. If it is the case, a new issue can be opened.
(Note: A11404.pdf shares the same layout type but still works correctly due to a differently identified depth column.)

TicaGit:
# Diagonal-line detection and depth–text matching

resolves #324 
resolves #290

This PR adds detection of diagonal connector lines that map depth intervals to text descriptions in borehole profiles. The main improvement is that text descriptions are now more reliably matched to their corresponding depth intervals when connector lines are diagonal (i.e., when the text should be matched to an interval above or below rather than only to adjacent intervals).

## What changed

* **Detect diagonal geometric lines** that end near text descriptions and treat them as connector cues.
* **Vertical-adjustment in scoring**: the scoring mechanism now virtually shifts text line positions vertically when a diagonal connector is present.
* **Affinity extension**: the affinity  computation now includes a **diagonal-line affinity** term as an additional factor.

## Affected documents (from issue #324)

Below are per-document results and notes.

* **267124070-bp.pdf** — *almost perfect*. Remaining errors are due to sidebar entries not identified.

* **267125058-bp.pdf** — *no change*. Connector lines are not diagonal but zig-zag; this PR doesn't improve zig-zag matching. This behaviour is unique to this document. Extracting a vertical shift from the zig-zag line is possible but currently produces too many false positives across the other documents. A connected-component tracing algorithm could follow the zig-zag path precisely, but that is high effort for a single document.

  Connector example

* **267125439-bp.pdf** — *almost perfect*. Allowing open-ended intervals should help this profile.

* **268124256-bp.pdf** — *perfect*.

* **268124569-bp.pdf** — *almost perfect*. Remaining errors are due to sidebar entries not identified.

* **269126084-bp.pdf** — *almost perfect*. Remaining errors are due to sidebar entries not identified.

* **675244002-bp.pdf** — *slightly better*, but many errors remain because sidebar entries are not identified.

* **A359.pdf** — *perfect* (should be 100% — ground truth likely incorrect).

* **A42.pdf** — could be improved; there is a tension between alignment adjustments and affinity penalties that needs more tuning.

* **A8561.pdf** — would benefit from the overlap-affinity introduced in another PR.

* **A8542.pdf** — not related to diagonal mapping; penalizing missed interval will help.

* **A531.pdf** - clear improvement, but many errors remain because sidebar entries or lines are not identified.

**Note:** 
- some strip logs are not detected on the page, and diagonal lines are (incorrectly) detected in those cases. Once strip-log detection improves, those documents should show slight gains (for example `681249002-bp.pdf`).

- opened issue #336 to address the remaining documents mentioned in #324

### DIFF
--- a/src/extraction/features/utils/geometry/line_detection.py
+++ b/src/extraction/features/utils/geometry/line_detection.py
@@ -147,7 +147,7 @@ def extract_lines(page: pymupdf.Page, line_detection_params: dict) -> tuple[list
     return filtered, merged
 
 
-def find_diags_ending_in_zone(lines: list[Line], line_search_zone: pymupdf.Rect):
+def find_diags_ending_in_zone(lines: list[Line], line_search_zone: pymupdf.Rect) -> list[Line]:
     """Find diagonal lines ending in a given zone (note: the end of the line is always on the right).
 
     Args:


src/extraction/features/extract.py
```diff
@@ -511,10 +527,117 @@ def is_valid_pair(
...
+            description_lines (list[TextLine]): The description lines.
+            index_to_diagonal (dict[int, Line]): The indices of the textlines that have a diagonal near them.
+
+        Returns:
+            dict[TextLine, float]: The vertical adjustements for each textline.
+        """
+        if not index_to_diagonal:
+            return {line: 0.0 for line in description_lines}
+
+        MAX_LINE_AFFECTED = diagonals_params["max_line_affected"]
```
stijnvermeeren-swisstopo: Having this as a fixed constant seems a little suspicious to me. Wouldn't it be more robust to e.g. say that we keep applying the adjustment until either there is a significant vertical gap between the text lines, or there is another geometric line nearby?
However, I can't give you any concrete examples where this would make a difference, so we could just put this idea into a follow-up ticket as far as I'm concerned.

src/extraction/main.py
```diff
@@ -367,7 +368,7 @@ def start_pipeline(
                     else:
                         img = plot_lines(
                             page,
-                            geometric_lines,
+                            long_or_horizontal_lines,
```
stijnvermeeren-swisstopo: wouldn't it make more sense to draw all geometric lines here?

src/extraction/features/stratigraphy/sidebar/classes/a_above_b_sidebar.py
```diff
@@ -180,14 +188,21 @@ def break_on_double_descending(self) -> list[AAboveBSidebar]:
...
+        shifted_intervals = self.depth_intervals(with_shift=True)
         return [
             IntervalZone(
-                interval.start.rect if interval.start else None, interval.end.rect if interval.end else None, interval
+                shifted_interval.start.rect if shifted_interval.start else None,
+                shifted_interval.end.rect if shifted_interval.end else None,
+                real_interval,
             )
-            for interval in self.depth_intervals()
+            for shifted_interval, real_interval in zip(shifted_intervals, depth_intervals, strict=True)
```
stijnvermeeren-swisstopo: Wouldn't it be simpler to define
```
    @property
    def shifted_rect(self):
        """Returns the bounding box, shifted by its vertical relative_shift."""
        return pymupdf.Rect(
            self.rect.x0, self.rect.y0 + self.relative_shift, self.rect.x1, self.rect.y1 + self.relative_shift
        )
```
instead of `shifted_copy()` on the `DepthColumEntry` class. Then we don't need the `with_shift` param on `AAboveBSidebar.depth_intervals()` and this method could be simplified to
```
    def get_interval_zone(self):
        depth_intervals = self.depth_intervals()
        return [
            IntervalZone(
                interval.start.shifted_rect if interval.start else None,
                interval.end.shifted_rect if interval.end else None,
                interval,
            )
            for interval in depth_intervals
        ]
```  

Or am I missing something?

src/extraction/features/stratigraphy/sidebar/classes/a_above_b_sidebar.py
```diff
@@ -49,17 +49,25 @@ def strictly_contains(self, other: AAboveBSidebar) -> bool:
...
         return all(self.entries[i].value < self.entries[i + 1].value for i in range(len(self.entries) - 1))
 
-    def depth_intervals(self) -> list[AAboveBInterval]:
+    def depth_intervals(self, with_shift: bool = False) -> list[AAboveBInterval]:
         """Creates a list of depth intervals from the depth column entries.
 
         The first depth interval has an open start value (i.e. None).
 
+        Args:
+            with_shift: whether the interval should include the shift created by the diagonals connectors.
```
christianabbet:
```suggestion
            with_shift (bool): whether the interval should include the shift created by the diagonals connectors.
```

src/extraction/features/utils/geometry/line_detection.py
```diff
@@ -140,7 +141,20 @@ def extract_lines(page: pymupdf.Page, line_detection_params: dict) -> list[Line]
...
     merged = merge_parallel_lines_quadtree(lines, tol=tol, angle_threshold=angle_tol)
-    merged = [
+    filtered = [
         line for line in merged if line.length >= min_line_length or line.is_horizontal(horizontal_slope_tolerance)
     ]
-    return merged
+    return filtered, merged
+
+
+def find_diags_ending_in_zone(lines: list[Line], line_search_zone: pymupdf.Rect):
```
christianabbet:
```suggestion
def find_diags_ending_in_zone(lines: list[Line], line_search_zone: pymupdf.Rect) -> list[Line]:
```

src/extraction/features/utils/geometry/geometry_dataclasses.py
```diff
@@ -158,6 +158,10 @@ def is_horizontal(self, horizontal_slope_tolerance) -> bool:
         """Checks if a line is horizontal."""
         return abs(self.slope) <= horizontal_slope_tolerance
 
+    def is_vertical(self, angle_tolerance) -> bool:
+        """Checks if a line is vertical."""
+        return abs(self.angle) >= 90 - angle_tolerance
```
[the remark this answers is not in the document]
TicaGit: self.angle is defined as 
```
@property
    def angle(self) -> float:
        """Angle of the line with the x-axis in degrees, ranging from -90 (exclusive) to +90 (inclusive)."""
        if self.start.x == self.end.x:
            return 90
        else:
            return math.atan(self.slope) / math.pi * 180
```
with math.atan always returning a value between 90 and -90

src/extraction/features/utils/text/textline_affinity.py
```diff
@@ -276,6 +287,23 @@ def compute_right_end_affinity(self, previous_line: TextLine, current_line: Text
...
+
+        Returns:
+            float: The affinity: -1.0 if lines are separated by a diagonal connection, 0.0 otherwise.
+        """
+        last_line_y_mid = (previous_line.rect.y0 + previous_line.rect.y1) / 2
+        current_line_y_mid = (current_line.rect.y0 + current_line.rect.y1) / 2
+        for g_line in self.diagonals:
+            if last_line_y_mid < g_line.end.y < current_line_y_mid:
+                return -1.0
+        return 0.0
```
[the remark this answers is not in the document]
TicaGit: I did it this way to have early exit as soon as we match, but diagonals should not be that big so it won't make a diff.

PR #337, opened 2025-11-06
Feat/issue 326/interval skip penalty

### LINKED ISSUE #326: Avoid incorrectly skipping intervals when matching depths with material descriptions
From the comments on #319 

> It feels to me like there should be some penalty for "skipping" a depth interval (i.e. assigning no description at all to it, corresponding to the "up" move in the dynamic programming table). There are many cases where this skipping an interval is necessary for getting a higher accuracy (often due to incorrectly identified depths and/or an incorrectly identified material description rect), so the penalty cannot be too strong. Still, there are some cases such as 688259001-bp.pdf, A8542.pdf and A359.pdf that would benefit from such a penalty.

TicaGit:
Resolves #326 

### Summary

This PR introduces a penalty applied to the `up` move in the dynamic programming algorithm. It discourages the algorithm from skipping an interval without assigning any line to it — effectively encouraging the `diag` move to be preferred when filling the DP table.

### Other issue

I also attempted to introduce an open-ended interval, meant to capture cases where material description lines were wrongly identified near the bottom of the page. This was not linked to any specific issue, but it was something we wanted to try.
However, adding this interval creates a side effect: lines whose corresponding sidebar entries are not detected tend to match the interval below. This misalignment then propagates downward, causing the final line to be added to the last open-ended interval. In most cases, this is still less costly than grouping two lines with negative affinity into the same interval.

This problem is illustrated below. In this example, it occurs only in the last two lines, but in other cases, it happens higher up in the document, leading to more incorrect assignments.
image

I experimented with assigning different penalties for placing lines in the open-ended interval:
- Penalties below -4 had no effect.
- When penalties reached the range of the maximum affinity values (line + line_on_the_left + vertical...), some text lines were assigned to the interval, but rarely in a correct way.
- Removing the affinity penalty for assigning a line to the last interval and applying a moderate penalty (around -1) seemed the most promising, but still resulted in degraded and undesirable outcomes.
It seams impossible to find a single penalty value that work for all documents becaause some lines at the bottom have strong negative affinity (horizontal line + line on the left + big gap...) so to compensate for that, the penalty should be as big. On the other end, some documents have no lines, so the biggest penalty they can have is arround -1.

Then, I tried tried to tweak the vertical affinity function in several ways:
- Removed the lower bound (-1) to penalize lines further away from grouping together - this degraded results.
- Applied a quadratic penalty instead of the linear decay to increase the cost of larger distances - also led to worse performance.

I also experimented with the AaboveB scoring function:
- Adjusted close_to_mid_zone_bonus to scale relative to interval width.
- Modified the final score weighting.
None of these changes improved the results.

### Current Implementation

At present, the code still creates the open-ended interval, but assigns each pair a score of -inf, effectively preventing any match. This interval is then removed during post-processing.

### Conclusion

I believe that once the strip log more accurately defines the material description zone, the open-ended interval will no longer be necessary. Additional filtering could also be applied to the extracted lines, such as removing single numbers or isolated dashes.

### Results

Overall performance improved slightly — about +0.5% for layer_f1 and material_description_f1 on the Zurich dataset, and remained unchanged on GeoQuat. The documents mentioned in the issue (A8542.pdf and 688259001-bp) both showed improvements.


src/extraction/features/stratigraphy/sidebar/classes/a_above_b_sidebar.py
```diff
@@ -60,6 +61,7 @@ def depth_intervals(self) -> list[AAboveBInterval]:
         depth_intervals = [] if self.entries[0].value == 0.0 else [AAboveBInterval(None, self.entries[0])]
         for i in range(len(self.entries) - 1):
             depth_intervals.append(AAboveBInterval(self.entries[i], self.entries[i + 1]))
+        depth_intervals.append(AAboveBInterval(self.entries[-1], None))  # include open-ended
```
stijnvermeeren-swisstopo: Since we don't actually use it (due to the penalty of infinity lower down), and are not likely to use it in the near future, I would prefer not adding this to the code at all.

PR #342, opened 2025-11-12
Feat/issue 332/indentation affinity

### LINKED ISSUE #322: Experiment impact of indentation-based affinity on overall scoring
Following PR #304, affinity scores between adjacent lines were introduced.
However, the methods `_indentation_affinity` and `_unindentation_affinity` are currently not used in the global score computation.

Task:
Investigate whether including them in the global score calculation improves results, if if so, chose affinity weights accordingly.
If the inclusion does not improve performance, remove these unused methods.

Additional Context:

It may be helpful to review how indentation was handled before merge #304 to identify which documents or datasets were affected by the change.

@stijnvermeeren-swisstopo check in LGD-288 for examples

TicaGit:
resolves #322

This PR adds support for indentation-based text line affinity to improve text-to-interval matching. The update enables the system to correctly associate indented description lines with the preceding line’s interval.

However, the algorithm still struggles with certain indented lines, as the scoring function’s weighting can dominate when an indented line appears significantly below its intended interval.

### Results

The changes are barely noticeable on geoquat, with a slight decrease in performance. On zurich, however, we observe a small improvement — approximately +0.5% in both layer_f1 and material_description_f1 metrics.

### Document mentioned in #322 
 - 267124180: no changes, the indentation affinity is not enough to overcome the allignement with the interval (i.e. pairing the indented line with the current is too good to the eyes of the algo)
 - 683248004: perfect, all indentation are resolved
 - 701241001: slight improvement, the algorithm still struggles with many indented lines (as the get far away from the desired interval), thus the affinity of 3. to encourage the matching. Some lines seem related to no interval… (even human struggles).
 - 700246002: much better. Some lines seem related to no interval… (even human struggles).
 - 699243001: slightly worse, but the split make sense.
 - 695248002: slight improvement
 - 679249080: clear improvement, but still struggles with some indented lines
 - 679249004: perfect
 - 678249001: clear improvement, but still bad due to missed depth entries.
 - 675246002: clear improvement, but still struggles with some indented lines
 - 675244002: no change
 - 269124191 : messy document, but slight improvement

### other impact:

- 695265009 : depth entries are detected as material description, which introduce some iregularities in the indentation detection, and produce worse results.
- A8384 and A8561 have worse result due to incorrect line grouping.

### DIFF
--- a/config/matching_params.yml
+++ b/config/matching_params.yml
@@ -14,7 +14,7 @@ affinity_params:
       right_end_weight: 1.0
       diagonal_weight: 1.25
       gap_weight: 0.0
-      indentation_weight: 1.0
+      indentation_weight: 0.6
   a_to_b:
     importance: 2.0
     weights:
--- a/src/extraction/features/extract.py
+++ b/src/extraction/features/extract.py
@@ -544,8 +544,10 @@ class MaterialDescriptionRectWithSidebarExtractor:
         # Zone where we will look for diagonal line ends, between the strip logs and material descriptions.
         search_zone = pymupdf.Rect(min_x0 - max_text_height, min_y0, max_x0 + max_text_height / 3, max_y1)
         if self.strip_logs:
-            # shrink left boundary to right edge of the strip
-            search_zone.x0 = max(search_zone.x0, max([sl.bbox.x1 for sl in self.strip_logs]))
+            left_strip_x1s = [sl.bbox.x1 for sl in self.strip_logs if sl.bbox.x0 < search_zone.x0]
+            if left_strip_x1s:
+                # Shrink left boundary to the rightmost edge of intersecting strips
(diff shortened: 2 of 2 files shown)


TicaGit:
I totally agree, I’ve also always had to adjust the other weights to keep their impact roughly balanced. I haven’t found an obvious solution for this; do you have any ideas?

stijnvermeeren-swisstopo:
@TicaGit Not really, apart from the obvious option of changing to a weighted sum instead of a weighted average. But then the "flakiness" just moves to a different point, namely balancing the affinities with the dynamic programming scores... Still, maybe it would be the more robust option. It probably depends on the type of affinity scores. When most inputs have a value of 0 for most affinity scores, then a weighted sum is probably more robust. If most inputs have a non-zero value for most affinity scores, then a weighted average is probably more robust.

PR #344, opened 2025-11-17
Feat/issue 339/en norm bert bedrock

TicaGit:
# PR Summary

This PR adds compatibility for the classification of unconsolidated material descriptions into the EN norm for both **BERT** and **Bedrock** models. It refactored `get_layer_ground_truth_key()` to support nested ground truth structures.

---

## BERT

The BERT model was trained using the ground-truth data in geoquat_ground_truth.json along with the file subset contained in geoquat/train. All hyperparameters from the configuration file were used, with the following adjustments for:

### Stage 1
- Learning rate: `3e-3`  
- Classifier and pooler layers unfrozen

### Stage 2
- Initialized from the best checkpoint of Stage 1  
- Learning rate: `1e-4`  
- Encoder layer 11 unfrozen, in addition to the classifier and pooler layers

The final model is uploaded to S3 under `en_models/`.  
Inference on the validation set completes in a few minutes.

---

## Bedrock

A new utility class, **`BedrockRetryInvoker`**, has been added to handle `ThrottlingException` errors caused by the high number of concurrent requests. Instead of skipping a layer and defaulting to `ns`, requests are now retried several times with increasing delays. With this mechanism and the current parameters, all layers are now classified, whereas many were previously skipped.

Inference on the validation set takes approximately **2 hours 10 minutes**. The current best version is the v5.

The model performs well overall, but it struggles significantly with the class `ba` (backfill, *remblai*). Because this layer is often composed of multiple materials, the model tends to classify the description according to the dominant material rather than identifying it as `ba`. Despite multiple prompt iterations and reformulation, this issue could not be improved. E.g. `Gravier sableux, légèrement limoneux, galets toutes formes, dm. 10 cm, avec débris de construction, compact, sec.` should be `ba` but the model classifies it as `gr`. 

The class `oth` (“other”) is effectively equivalent to `ns`. The current prompt explicitly instructs the model not to assign `oth` and to use `ns` instead, since `oth` does not appear in the ground truth. If `oth` begins to appear in the dataset, the prompt should be updated to clarify what `oth` represents and how it differs from `ns`.

---

### GeoQuat validation
| Model   | Macro F1 | Micro F1 |
|---------|----------|----------|
| BERT    | 0.87    | 0.90     |
| Bedrock | 0.78     | 0.84     |


src/classification/config/bedrock/en_main_classification_prompts.yml
```diff
@@ -0,0 +1,418 @@
...
+      wach siltiger" indique un petit ajout de silt, le **nom** reste Sand et le modificateur granulaire explicite ("Fein-Mittel") prime pour la sous-classe.
+
+
+    user_prompt_description: |
+      This is the material description:
+      <material_description>
+      {material_description}
+      </material_description>
+  
+  v5: # current best
```
TicaGit: only keep this one, others can still be seen from the commit history

PR #349, opened 2026-01-07
Issue 348: replace quads with fastquadtree

### LINKED ISSUE #348: Switch from quads to fastquadtree library
When trying to run data extraction on page 2 of the file `4839_Neubau_2_MFH_Stocken_Oberneunforn_BS3_17.pdf
` from the Thurgau dataset, we get the following error:
```
ValueError: Point <Point: (260, 375)> is not within this node (<Point: (263.1147766113281, 374.89404296875)> - <BoundingBox: (260.0000305175781, 371.3519592285156) to (266.2295227050781, 378.4361267089844)>).
```
The very small difference between `260` (from the point that we are trying to insert) and `260.0000305175781` (from the rect inside the quad tree structure) suggests that there might be some issue with floating point arithmetic inside the quads library that we use for quadtrees. Unforunately, this library was last updated six years ago (https://github.com/toastdriven/quads) so it is probably not worthwhile digging into the internal workings of their quadtree implementation. Rather, we should check if we can switch to a better-maintained implementation of quad trees. The relatively new library https://github.com/Elan456/fastquadtree seems to be a good candidate with great documentation.

stijnvermeeren-swisstopo:
closes #348 

In addition to quads, I noticed that rtree can also be easily replaced by a fastquadtree class (`RectQuadTreeObjects`). I went ahead and did that as well, increasing code consistency and reducing the number of project dependencies.

### DIFF
--- a/src/extraction/evaluation/benchmark/ground_truth.py
+++ b/src/extraction/evaluation/benchmark/ground_truth.py
@@ -36,7 +36,6 @@ class GroundTruth:
                         "depth_interval": layer["depth_interval"],
                     }
                     for layer in layers
-                    if layer["material_description"] is not None and parse_text(layer["material_description"]) != ""
                 ]
                 self.ground_truth[file_name][borehole_index]["metadata"] = borehole_data["metadata"]
                 self.ground_truth[file_name][borehole_index]["groundwater"] = borehole_data["groundwater"]


stijnvermeeren-swisstopo:
@christianabbet Thanks for the suggestion. While I have not used UV before, it sounds promising and could offer us quite a few benefits. The only immediate downside that I see, is the additional installation step, which shouldn't be too bad. Will you create a new issue for switching to UV?

PR #360, opened 2026-01-16
352 evaluate multiple datasets in one run

### LINKED ISSUE #352: Evaluate multiple datasets in one run (extraction)
For the **extraction** pipeline: 
Goals:
- Make it possible to run multiple benchmarks with one command, to avoid forgetting to run certain benchmarks
- Make it possible to compute overall metrics that combine several benchmarking datasets

To be defined:
- How do we specify the datasets elegantly with command line parameters
  - Also affects VSCode json configurations
- How to track the results from individual runs as well as overall metrics in MLFlow?

AgathaSchmidt:
Closes #352 

**Summary**
This PR introduces multi-benchmark evaluation in a single execution for the borehole extraction pipeline.
It allows running and comparing multiple datasets (e.g. Zurich, Geoquat, Thurgau) in one command. The mlflow tracking is structured in a parent-child relationship with a summary of all child runs available on parent level. 
It is not necessary to provide multiple datasets as input, the configuration also works with one dataset. 

**Results/ Outputs**:
The changes proposed in the PR enable: 
--> Run multiple benchmarks in one run instead of launching separate CLI commands
--> Automatic generation of per-benchmark outputs, an aggregates summary JSON as well as a human-readable comparisonCV
--> Results are saved as metric as well as artifact which enables to compare benchmarks side-by-side in a single overview
--> Minimize main.py by moving the start_pipeline function(s) to the runner.py

**Main Changes**
_launch.json_

- add a multi benchmark configuration
- args is a list of benchmark specs that include the
    - name
    - input path
    - ground truth path

_spec.py_

a new file within the src/extraction/evaluation/benchmark folder that
- defines what a benchmark means (benchmark spec which is used in launch json) :
--> defines a class (BenchmarkSpec) that contains name string, input path and ground truth path
--> defines a parser that parses a spec of the form <name>_<input_path>:<ground_truth_path> to a BenchmarkSpec object

_runner.py_

- a new file within the src/extraction folder
- Multi-benchmark orchestration running multiple benchmarks in a single execution
- It creates a parent child relation creating one parent mlflow run which will have the summary as an artifact and it creates n child runs depending on the number of benchmark specs in the configuration list of the launch.json
- process
    - it loops over the list of benchmarkSpecs
    - calls start_pipeline() once per benchmark
    - collects results
    - writes aggregated outputs (json and csv)
    - manages MLflow parent/child runs
- e.g. it enables to run zurich + geoquat + thurgau in one command and compare them cleanly in a summary view

_extraction/main.py_

- make i argument not required but optional to allow multi-benchmarks
- add the —benchmark click option
- add a multi benchmark mode to the click pipeline
- manage temp_directories so that for every benchmark dataset we have an own file instead of overwriting the same multiple times. We need multiple files to create a comprehensive summary.

_score.py_
- add support for multi-benchmark runs so that scoring can aggregate and report results across several benchmarks in one execution

### DIFF
--- a/src/utils/benchmark_utils.py
+++ b/src/utils/benchmark_utils.py
@@ -78,7 +78,7 @@ def _short_metric_key(k: str) -> str:
 
 
 def log_metric_mlflow(
-    summary: BenchmarkSummary | None,
+    summary: BenchmarkSummary,
     out_dir: Path,
     artifact_name: str = "benchmark_summary.json",
 ) -> None:


src/extraction/evaluation/benchmark/runner.py
```diff
@@ -0,0 +1,280 @@
...
+from swissgeol_doc_processing.utils.file_utils import flatten
+
+from .spec import BenchmarkSpec
+
+
+def _flatten_metrics(d: dict[str, Any], prefix: str = "") -> dict[str, float]:
+    """Flatten a nested metrics dict into {"geology/layer_f1": 0.63, ...} format.
+
+    Keeps only numeric values (int/float or strings convertible to float).
+    """
```
christianabbet: If possible try to add arguments in the description. You can use marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring to help you.

def _flatten_metrics(d: dict[str, Any], prefix: str = "") -> dict[str, float]:
    """Short description.

    longer description (if needed).

    Args:
        d (dict[str, Any]): Description arg1.
        prefix (str, optional): Description arg2.

    Returns:
        dict[str, float]: Description return.
    """

src/extraction/evaluation/benchmark/runner.py
```diff
@@ -0,0 +1,280 @@
...
+    draw_lines: bool = False,
+    draw_tables: bool = False,
+    draw_strip_logs: bool = False,
+    csv: bool = False,
+    matching_analytics: bool = False,
+    part: str = "all",
+    mlflow_tracking: bool = False,
+    line_detection_params: dict | None = None,
+    matching_params: dict | None = None,
+):
```
christianabbet: The function is quite long. Maybe it would be worth it to split it into multiple components. For example, one function to setup mlflow tracking, and one to compute the metric in the end?

src/extraction/evaluation/benchmark/score.py
```diff
@@ -30,13 +32,13 @@ def evaluate_all_predictions(
         ground_truth_path (Path | None): The path to the ground truth file.
         temp_directory (Path): The path to the temporary directory.
 
-    Returns:
-        None | pd.DataFrame: the document level metadata metrics
+    Returns a JSON-serializable dict summary that can be used by multi-benchmark runners.
```
christianabbet: Typing was removed for return
```
    Return:
        pd.DataFrame: JSON-serializable dict summary that can be used by multi-benchmark runners.
```

src/extraction/evaluation/benchmark/score.py
```diff
@@ -31,12 +33,13 @@ def evaluate_all_predictions(
         temp_directory (Path): The path to the temporary directory.
 
     Returns:
-        None | pd.DataFrame: the document level metadata metrics
+        dict | None: A JSON-serializable dict summary that can be used by multi-benchmark runners.
```
christianabbet: The return dict | None does not match the one of the function. It is a legacy problem. Can you just change the function return to dict | None instead of pd.DataFrame | None

src/utils/benchmark_utils.py
```diff
@@ -0,0 +1,99 @@
...
+        k (str): The original metric key.
+
+    Returns:
+        str: The shortened metric key.
+    """
+    return k.split("/", 1)[1] if "/" in k else k
+
+
+def log_metric_mlflow(
+    summary: BenchmarkSummary | None,
```
stijnvermeeren-swisstopo: Either you should not allow `summary`  to be `None`, or you should gracefully deal with that case. 

Since one of the two places that call this method already guard against a `None` value, I would suggest to also do that check at the other place (i.e. `start_pipeline_benchmark`) and keep this method as simple as it is.

stijnvermeeren-swisstopo (changes requested):

stijnvermeeren-swisstopo (approved):
LGTM!

PR #373, opened 2026-01-22
Feat/issue 359 evaluate multiple datasets in one run classification

### LINKED ISSUE #359: Evaluate multiple datasets in one run (classification)
For the **classification** pipeline:

Goals:

- Make it possible to run multiple benchmarks with one command, to avoid forgetting to run certain benchmarks
- Make it possible to compute overall metrics that combine several benchmarking datasets

AgathaSchmidt:
Closes  #359  
Related to #360 

**Summary**
This PR introduces multi-benchmark evaluation in a single execution for the borehole classification pipeline.
It allows running and comparing multiple datasets (e.g. Zurich, Geoquat, Thurgau) in one command. The mlflow tracking is structured in a parent-child relationship with a summary of all child runs available on parent level.
It is not necessary to provide multiple datasets as input, the configuration also works with one dataset.
**Currently, we only have classification GT available for one dataset so the current behavior skips benchmarking for all datasets without GT.** 

**Results/ Outputs:**
The changes proposed in the PR enable:
--> Run multiple benchmarks in one run instead of launching separate CLI commands
--> Automatic generation of per-benchmark outputs, an aggregates summary JSON as well as a human-readable comparisonCSV
--> Results are saved as metric as well as artifact which enables to compare benchmarks side-by-side in a single overview
--> Minimize main.py by moving the start_pipeline function(s) to the runner.py
--> moved the utils folder from extraction back to src, because the functions are share by extraction and classification 

**Main Changes**
_launch.json_

- add two multi classification benchmark configuration. for both classification modes: 
1. classification of descriptions in ground truth file 
2. prediction of prediction

_spec.py_

a new file within the src/classification/evaluation/benchmark folder that

defines what a benchmark means (benchmark spec which is used in launch json) :
--> defines a class (BenchmarkSpec) that contains file_path, file_subset_directory and ground_truth_path
--> defines a parser that parses different specs into a BenchmarcSpec object. Supports different specification types. 

_runner.py_

- a new file within the src/classification folder
- Multi-benchmark orchestration running multiple benchmarks in a single execution
- It creates a parent child relation creating one parent mlflow run which will have the summary as an artifact and it creates n child runs depending on the number of benchmark specs in the configuration list of the launch.json

_main.py_
- remove the start_pipeline function 
- add multi benchmark option

### DIFF
--- a/src/classification/runner.py
+++ b/src/classification/runner.py
@@ -266,7 +266,6 @@ def start_pipeline(
         file_path, ground_truth_path, file_subset_directory, classification_system_cls
     )
 
-    # n_documents = count_documents(file_path, file_subset_directory)
     n_documents = len({layer.filename for layer in layer_descriptions})
     if mlflow:
         mlflow.log_metric("n_documents", float(n_documents))


src/utils/dynamic_matching.py
christianabbet: Is there a reason to put this file in `utils` ? I only see usage in `extraction` ?

src/utils/benchmark_utils.py
```diff
@@ -5,50 +5,93 @@
...
+        else:
+            rels.append(str(rel))
 
     return rels
 
 
+T = TypeVar("T")
+
+
+def parent_group_key(
```
christianabbet: I cannot see any use of `parent_group_key` in the code ?

src/utils/benchmark_utils.py
```diff
@@ -5,50 +5,93 @@
...
+    *,
+    prefix: str = "multi",
+) -> str:
+    """Stable group key for the benchmark set, based on a per-benchmark path."""
+    paths = [get_path(b) for b in benchmarks]
+    inputs = " | ".join(sorted(_relative_after_common_root(paths)))
+    return f"{prefix}:{inputs}"
+
+
 def _parent_input_directory_key(benchmarks: Sequence[BenchmarkSpec]) -> str:
```
christianabbet: Small typing issue, the function is fed to `setup_mlflow_tracking` which requires a Path not str. This is minor, we need to add ty in the futur to help

src/classification/runner.py
```diff
@@ -0,0 +1,424 @@
...
+    classifier_type: str,
+    model_path: Path,
+    classification_system: str,
+    mlflow_setup: bool = True,
+) -> ClassificationBenchmarkSummary | None:
+    """Main pipeline to classify the layer's soil descriptions.
+
+    Args:
+    file_path (Path): Path to the json file we want to predict from.
+    ground_truth_path (Path): Path the the ground truth file, if file_path is the predictions.
```
christianabbet: typo

src/classification/runner.py
```diff
@@ -0,0 +1,424 @@
...
+    out_directory (Path): Path to output directory.
```
christianabbet: Missing indentation for args

src/classification/evaluation/benchmark/spec.py
```diff
@@ -0,0 +1,83 @@
...
+
+def parse_benchmark_spec(value: str) -> BenchmarkSpec:
+    """Parse a benchmark spec.
+
+    Args:
+        value: A string in one of the following formats:
+            1) '<name>:<file_path>:<file_subset_directory>'
+            2) '<name>:<file_path>:<file_subset_directory>:<ground_truth_path>'
+            3) '<name>:<predictions_path>:<ground_truth_path>'
+     Returns:
```
christianabbet: Trailing space in front

src/classification/runner.py
```diff
@@ -0,0 +1,424 @@
...
+    model_path: Path | None,
+    classification_system: str,
+    out_directory_bedrock: Path,
+    classification_params: dict | None = None,
+):
+    """Run multiple classification benchmarks in one execution.
+
+    Args:
+    benchmarks: A sequence of BenchmarkSpec, each specifying a benchmark to run.
+    out_directory: The root output directory where subdirectories for each benchmark will be created.
```
christianabbet: Indentation of the doc

src/classification/runner.py
```diff
@@ -0,0 +1,424 @@
...
+logger = logging.getLogger(__name__)
+
+classification_params = read_params("classification_params.yml")
+
+mlflow_tracking = os.getenv("MLFLOW_TRACKING") == "True"  # Checks whether MLFlow tracking is enabled
+if mlflow_tracking:
+    import mlflow
+
+
+def _finalize_overall_summary(
```
christianabbet: Is there really a case when this function is called with `parent_active` being False ?

src/classification/runner.py
```diff
@@ -0,0 +1,424 @@
...
+                mlflow.end_run()
+
+        _finalize_overall_summary(
+            overall_results=overall_results,
+            multi_root=multi_root,
+            parent_active=parent_active,
+        )
+    finally:
+        if mlflow_tracking and parent_active:
+            import mlflow
```
christianabbet: Remove, usew of global import instead

src/classification/runner.py
```diff
@@ -0,0 +1,424 @@
...
+
+    return summary_json_path, summary_csv_path
+
+
+def setup_mlflow_tracking(
+    file_path: Path | None,
+    out_directory: Path,
+    file_subset_directory: Path | None,
+    experiment_name: str = "Layer descriptions classification",
+):
```
christianabbet: Return 

`) -> None:`

christianabbet (changes requested):

src/extraction/utils/utils.py
```diff
@@ -0,0 +1,45 @@
...
+def log_metric_mlflow(
+    summary: BenchmarkSummary,
+    out_dir: Path,
+    artifact_name: str = "benchmark_summary.json",
+) -> None:
+    """Log benchmark metrics + a JSON summary artifact to the *currently active* MLflow run.
+
+    - Does NOT start/end MLflow runs.
+    """
+    import mlflow
```
christianabbet: I would prefer if you use this at the top instead. Then we will refactor it in the next sprint. 

```
mlflow_tracking = os.getenv("MLFLOW_TRACKING") == "True"  # Checks whether MLFlow tracking is enabled
if mlflow_tracking:
    import mlflow
```

src/classification/evaluation/benchmark/spec.py
```diff
@@ -0,0 +1,84 @@
+"""BenchmarkSpec class specification and respective parsing for classification."""
+
+from dataclasses import dataclass
+from pathlib import Path
+
+
+@dataclass(frozen=True)
+class BenchmarkSpec:
```
stijnvermeeren-swisstopo: Though there are some differences with the `BenchmarkSpec` for extraction, in particular the `file_subset_directory` (though TBH I can't immediately remember why we need that at all), I still wonder if we can't make a generic `BenchmarkSpec` class that works for both extraction and classification.

Maybe that's rather something for a follow-up issue though, unless you think that it would be an easy change.

src/classification/runner.py
```diff
@@ -0,0 +1,407 @@
...
+
+            bench_out_bedrock = multi_bedrock_root / spec.name
+            bench_out_bedrock.mkdir(parents=True, exist_ok=True)
+
+            if mlflow_tracking:
+                mlflow.start_run(run_name=spec.name, nested=True)
+                setup_mlflow_tracking(
+                    file_path=spec.file_path,
+                    out_directory=out_directory,
+                    file_subset_directory=None,
```
stijnvermeeren-swisstopo: Should be set to `spec.file_subset_directory`

stijnvermeeren-swisstopo (changes requested):

src/classification/runner.py
```diff
@@ -0,0 +1,373 @@
...
+                    if hasattr(summary, "metrics_flat"):
+                        mlflow.log_metrics(summary.metrics_flat(short=True))
+
+                    else:
+                        # fallback: log raw metrics dict
+                        mlflow.log_metrics({str(key): float(value) for key, value in (summary.metrics or {}).items()})
+
+                mlflow.end_run()
+
+        _finalize_overall_summary(overall_results=overall_results, multi_root=multi_root, mlflow_tracking=mlflow)
```
stijnvermeeren-swisstopo: Redundant param `mlflow_tracking`.

stijnvermeeren-swisstopo (changes requested):
Could you create a follow-up issue for everything that we mentioned regarding more shared code between extraction and classification?

src/classification/runner.py
```diff
@@ -0,0 +1,427 @@
...
+                mlflow.log_metric(short_key, value)
+
+        if pd.notna(total_docs):
+            mlflow.log_metric("total_n_documents", float(total_docs))
+
+        mlflow.log_artifact(str(summary_path), artifact_path="summary")
+        mlflow.log_artifact(str(summary_csv_path), artifact_path="summary")
+
+
+def setup_mlflow_tracking(
```
christianabbet: Can't we share mlflwo setup between the two packages? Most of the arguments should be the same no? If not then just add/remove on both sides

src/extraction/utils/utils.py
```diff
@@ -0,0 +1,31 @@
...
+
+from extraction.evaluation.benchmark.score import BenchmarkSummary
+
+mlflow_tracking = os.getenv("MLFLOW_TRACKING") == "True"  # Checks whether MLFlow tracking is enabled
+if mlflow_tracking:
+    import mlflow
+
+
+def log_metric_mlflow(
+    summary: BenchmarkSummary,
```
christianabbet: Rename BenchmarkSummary to ExtarctionBenchmarkSummary

src/classification/runner.py
```diff
@@ -0,0 +1,427 @@
...
+)
+from core.mlflow_tracking import mlflow
+
+logger = logging.getLogger(__name__)
+
+classification_params = read_params("classification_params.yml")
+
+
+def _finalize_overall_summary(
+    *,
```
christianabbet: Is there no way to share a bt of code with extraction? 

What can be done is the create a BenchmarkSummary object that is shared by both classes ClassificaitonBenchmarkSummary and ExtarctionBenchmarkSummary. Then create a function that create the metric, csv that is package specific. I think most of the writting to mlflow and as csv is shared. Just the content of the dump changes no?

christianabbet (changes requested):
It seems as well that the class from Metrics from extraction.evaluation.evaluation_dataclasses is used in classification (src/classification/evaluation/evaluate.py). It would be nice to share it in Core as well

src/core/benchmark_utils.py
```diff
@@ -0,0 +1,124 @@
...
+    group_key = f"multi:{inputs}"
+    return group_key
+
+
+def configure_logging(level: int = logging.INFO) -> None:
+    """Configure root logging once for the application."""
+    logging.basicConfig(format=DEFAULT_FORMAT, level=level, datefmt=DEFAULT_DATEFMT)
+
+
+def _short_metric_key(k: str) -> str:
```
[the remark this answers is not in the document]
AgathaSchmidt: moved to #389

src/classification/runner.py
AgathaSchmidt: moved to #389

src/classification/runner.py
AgathaSchmidt: moved to #389

src/extraction/utils/benchmark_utils.py
```diff
@@ -9,7 +9,7 @@
 from pathlib import Path
```
stijnvermeeren-swisstopo: Delete this file and import the corresponding method from `core/benchmark_utils.py` in the extraction runner instead.

stijnvermeeren-swisstopo (changes requested):

stijnvermeeren-swisstopo (changes requested):

PR #376, opened 2026-01-27
Feat/issue 82/extraction pipline integration assets

letao:
Related to https://github.com/swisstopo/swissgeol-assets-dataextraction/issues/82 

This pull request refactors and improves the extraction pipeline for borehole PDF documents, it provides an option of using the `extraction` pipeline as a standalone package following the same approach defined for `swissgeol_doc_processing`.

For specific needs of **Swissgeol-assets-dataextraction** project the `src/extraction/minimal_pipeline.py` script has been created for a specific borehole feature extraction logic:

https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/feat/issue-82/extraction-pipline-integration-assets/src/extraction/minimal_pipeline.py

Additional improvements:

* Added `extract_sidebars_with_quality_metrics` method and a new `extract_sidebar_information` function to compute sidebar quality metrics (number of candidates, good sidebars, best score, types found, average noise), providing modular feature extraction. 

https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/12e0a65e7f7759183400f204dcc6cd66baae12f2/src/extraction/features/extract.py#L731-L783

* Refactored extraction pipeline to pass `page` objects instead of `document` objects to extraction functions, simplifying the API.

* Removed `DATAPATH` variable usage from `src/extraction`

* Updated python package imports to the absolute path convention in `src/extraction`

### DIFF
--- a/src/extraction/runner.py
+++ b/src/extraction/runner.py
@@ -11,7 +11,6 @@ import pandas as pd
 import pymupdf
 from tqdm import tqdm
 
-from swissgeol_doc_processing.utils.file_utils import get_data_path
 from extraction.annotations.draw import draw_predictions, plot_strip_logs, plot_tables
 from extraction.annotations.plot_utils import plot_lines, save_visualization
 from extraction.evaluation.benchmark.score import BenchmarkSummary, evaluate_all_predictions
@@ -28,13 +27,13 @@ from extraction.features.predictions.overall_file_predictions import OverallFile
 from extraction.features.predictions.predictions import BoreholeListBuilder
 from extraction.features.stratigraphy.layer.continuation_detection import merge_boreholes
 from extraction.features.stratigraphy.layer.layer import LayersInDocument
+from extraction.utils.benchmark_utils import _parent_input_directory_key, _short_metric_key, log_metric_mlflow
 from swissgeol_doc_processing.geometry.line_detection import extract_lines
 from swissgeol_doc_processing.text.extract_text import extract_text_lines
 from swissgeol_doc_processing.text.matching_params_analytics import create_analytics
-from swissgeol_doc_processing.utils.file_utils import flatten, read_params
+from swissgeol_doc_processing.utils.file_utils import flatten, get_data_path, read_params
 from swissgeol_doc_processing.utils.strip_log_detection import detect_strip_logs
 from swissgeol_doc_processing.utils.table_detection import detect_table_structures
-from extraction.utils.benchmark_utils import _parent_input_directory_key, _short_metric_key, log_metric_mlflow
 
 from .evaluation.benchmark.spec import BenchmarkSpec
 


src/extraction/features/extract.py
```diff
@@ -622,6 +622,57 @@ def _filter_diagonals(
...
+        # Average noise across all pairs with sidebars
+        noise_counts = [pair.noise_count for pair in all_pairs if pair.sidebar is not None]
+        average_sidebar_noise = sum(noise_counts) / len(noise_counts) if noise_counts else 0.0
+
+        return {
+            "number_of_sidebar_candidates": number_of_sidebar_candidates,
+            "number_of_good_sidebars": number_of_good_sidebars,
+            "best_sidebar_score": best_sidebar_score,
+            "sidebar_types_found": sidebar_types_found,
+            "average_sidebar_noise": average_sidebar_noise,
```
stijnvermeeren-swisstopo: I'm not convinced about the usefulness of metrics such as `number_of_sidebar_candidates`, `average_sidebar_noise` and `sidebar_types_found`. Wouldn't it be better to always restrict to "good" sidebars (i.e. always first apply filtering by positive score, table criteria and intersections)? Because if there is a good borehole on the page, then the presence of an incorrectly identified sidebar that is easily filtered out, should not reduce the confidence in the page classification "borehole".

Additionally, this would allow more code sharing between `process_page` and this method, because all of the sidebar detection and filtering can be shared (the only difference being whether to also search for material descriptions without sidebar, which can be made optional).

src/extraction/minimal_pipeline.py
```diff
@@ -0,0 +1,250 @@
...
+            - page_number: Page index
+            - number_of_valid_borehole_descriptions: Count of valid material descriptions
+            - number_of_strip_logs: Count of detected strip logs
+            - number_of_tables: Count of detected tables
+            - number_of_boreholes: Count of detected boreholes
+            - sidebar_information: Sidebar extraction results
+            - number_long_or_horizontal_lines: Count of long/horizontal lines
+            - number_all_geometric_lines: Count of all geometric lines
+            - text_line_count: Number of text lines on page
+            - borehole_name_entries: List of detected borehole names with confidence scores
```
stijnvermeeren-swisstopo: Also here, I would prefer returning a dataclass instead of a dict with "random" types.

src/extraction/minimal_pipeline.py
```diff
@@ -0,0 +1,250 @@
...
+    else:
+        strip_logs = detect_strip_logs(page, text_lines, striplog_detection_params)
+
+    number_of_strip_logs = len(strip_logs)
+
+    # If tables already extracted in context, reuse them
+    if extraction_context is not None and extraction_context.table_structures is not None:
+        table_structures = extraction_context.table_structures
+    else:
+        table_structures = detect_table_structures(page, long_or_horizontal_lines, text_lines, table_detection_params)
```
stijnvermeeren-swisstopo: Wouldn't it be simpler to make the attributes of the `ExtractionContext` non-optional (i.e. types that do not include `None`), and then move the logic for initializing the different attributes outside of this method?

Or do we have use cases where we might only want to compute some, but not all of the attributes in the `ExtractionContext`?

stijnvermeeren-swisstopo (changes requested):

stijnvermeeren-swisstopo (approved):
LGTM!

Issue #362, opened 2026-01-19
Incremental Output Writing


christianabbet:
Followup of https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/351#issue-3792285577

# Problem

Extracted data is currently written only at the end of the pipeline, and PNG visualizations are generated last. When processing thousands of borehole profiles, a late error causes all previously computed outputs to be lost.

# Proposal
- Write JSON outputs per file as soon as processing completes
- Generate PNGs incrementally, not only at the end
- Validate the pipeline is working as expected using tests (files are generated given arguments)
- (?) Allow the pipeline to resume after interruption, skipping already processed files

PR #377, opened 2026-01-28
Incremental output writing

christianabbet:
Closes https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/362
## Summary
Implements incremental output writing for the extraction pipeline to prevent data loss when processing large batches of borehole profiles. If the pipeline crashes or is interrupted, it can now resume from where it left off.

## Changes
- Incremental writing: Predictions are saved to a `.tmp` file after each processed PDF, then copied to the final output on completion
- MLflow run resumption: MLflow run IDs are saved locally as `.tmp`, allowing tracking to continue across interrupted runs
- Documentation: Improved docstrings with clear parameter descriptions.
- Testing: Added comprehensive I/O tests to validate file generation
- Extraction: Bytes IO is now supported for the core function `extract`
- To resume extraction use the flag `-r` or `--resume`

## Limitations

* Resuming matching analytics is not supported (analytics restart from scratch on resume)

### DIFF
--- a/src/extraction/features/predictions/overall_file_predictions.py
+++ b/src/extraction/features/predictions/overall_file_predictions.py
@@ -25,7 +25,7 @@ class OverallFilePredictions:
             filename (str): Filename to check.
 
         Returns:
-            bool: True if `filename` is contained `file_predictions_list`, else False.
+            bool: True if `file_predictions_list` contains `filename`, else False.
         """
         return any(file.file_name == filename for file in self.file_predictions_list)
 


tests/test_extraction_io.py
```diff
@@ -0,0 +1,240 @@
...
+    # Check inference of all vs not all
+    infer_part(part="metadata")
+    assert not predictions_path.exists()
+
+    infer_part(part="all")
+    assert predictions_path.exists()
+
+
+def test_start_pipeline_nested(tmp_path: Path, borehole_pdf: Path) -> None:
+    """Test that pipeline resumes from temporary file and skips already processed files."""
```
AgathaSchmidt: Args:
        tmp_path (Path): Path to temporary folder (pytest handled).
        borehole_pdf (Path): Path to borehole PDF file.

tests/test_extraction_io.py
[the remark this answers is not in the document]
christianabbet: I think it comes from the testing process. As the test function are run on the github account MLFLOW_TRACKING should be false, and then should not generate any runs. To mimic that locally jsut turn off MLFLOW tracking in your env. 

Still I think you made a good point that this is not clear. I'll try to see if we can enforce the behaviour MLFLOW_TRACKING=False such that it always work as expected

src/extraction/utils/benchmark_utils.py
```diff
@@ -14,6 +14,11 @@
 DEFAULT_FORMAT = "%(asctime)s %(levelname)-8s %(name)s: %(message)s"
 DEFAULT_DATEFMT = "%Y-%m-%d %H:%M:%S"
 
+mlflow_tracking = os.getenv("MLFLOW_TRACKING") == "True"  # Checks whether MLFlow tracking is enabled
+
+if mlflow_tracking:
+    import mlflow
+
```
stijnvermeeren-swisstopo: I'm not a huge fan of these global variables, but I guess that as a pragmatic approach it's still ok. I would try to reduce the code duplication though. Maybe we could create a file like `src/utils/mlflow_tracking.py` that has something like

```python
import os

mlflow_tracking = os.getenv("MLFLOW_TRACKING") == "True"  # Checks whether MLFlow tracking is enabled

mlflow = None

if mlflow_tracking:
    import mlflow
```

and then elsewhere we can just do `from utils.mlflow_tracking import mlflow` and
```
if mlflow:
    mlflow.log_...()
``` 

It looks a little weird though, re-exporting an imported library in an "optional" way... Maybe there are some good reasons not to do it like this. But it seems to work, and it would significantly reduce code duplication.

christianabbet: Good point! Hopefuly the issue https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/383#issue will make this cleaner as well

src/extraction/features/predictions/overall_file_predictions.py
```diff
@@ -18,6 +18,17 @@ def __init__(self) -> None:
...
         self.file_predictions_list: list[FilePredictions] = []
 
+    def contains(self, filename: str) -> bool:
+        """Check if filename is contained `file_predictions_list`.
+
+        Args:
+            filename (str): Filename to check.
+
+        Returns:
+            bool: True if `filename` is contained `file_predictions_list`, else False.
```
stijnvermeeren-swisstopo:
```suggestion
            bool: True if `file_predictions_list` contains `filename`, else False.
```

stijnvermeeren-swisstopo (approved):
Looks good to me 👍

PR #397, opened 2026-02-24
save PNG images to draw directory with correct colorspace and compression

stijnvermeeren-swisstopo:
The PNG files in the draw directory had incorrect colors, while the colors for the MLFlow artifact PNG images were correct. This was due to the fact that we use cv2.imwrite for saving the images to the draw directory, and CV2 expects a BGR colorspace instead of RGB.

Moreover, to save some disk space, this PR also adds PNG compression.

The PR also adds some type hints for the affected method, and removes a log statement that does not seem very useful.

### DIFF
--- a/src/extraction/annotations/plot_utils.py
+++ b/src/extraction/annotations/plot_utils.py
@@ -170,10 +170,12 @@ def save_visualization(
     img: np.ndarray, filename: str, page_number: int, visualization_type: str, draw_directory: Path
 ):
     """Save visualization image to file and/or MLflow."""
+    filename_visualization = f"{Path(filename).stem}_page{page_number}_{visualization_type}.png"
+
     if draw_directory:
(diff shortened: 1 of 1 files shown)


src/extraction/annotations/plot_utils.py
```diff
@@ -166,14 +166,14 @@ def draw_blocks_and_lines(
     return open_cv_img
 
 
-def save_visualization(img, filename, page_number, visualization_type, draw_directory):
+def save_visualization(
+    img: np.ndarray, filename: str, page_number: int, visualization_type: str, draw_directory: Path
+):
     """Save visualization image to file and/or MLflow."""
     if draw_directory:
         img_path = draw_directory / f"{Path(filename).stem}_page{page_number}_{visualization_type}.png"
```
christianabbet: Minor comment: I was wondering if there is a reason why we use once `Path(filename).stem` for local and `{filename}` for mlflow ? Maybe we can homogenize it

PR #402, opened 2026-03-03
Feat/issue 370/cluster depth values with skew

### LINKED ISSUE #370: Fix missing top depth layers - Thurgau
Depth extraction often misses the first (top) layers, particularly in borehole profiles from “progeo”. These scans are frequently slightly skewed, which likely causes the top portion of the stratigraphy table/layering to be partially missed (`1000`, `1004`, `1023`). 

An edge case is `1046` with a lot of pages. Only page 7 is a borehole, but tables in other pages are detected as boreholes.

Image

stijnvermeeren-swisstopo:
closes #370 

We now get perfect results for the first page of the three boreholes in the ticket...

1000_EWS_Kreuzlingen_8286.pdf
image

1004_EWS_Kreuzlingen_3087.pdf
image

1023_EWS_Kreuzlingen_6492.pdf:
image

The main issue was that the clustering logic in src/extraction/features/stratigraphy/sidebar/utils/cluster.py was very inflexible, having no tolerance for skewed images, and being more restrictive if the first value is very narrow (e.g. single-digit) relative to later (multi-digit) values.

The new approach does not just try to build a cluster relative to the x-coordinates of the first element, but considers pairs of first and last elements, that are allowed to have a certain skew between them. This introduces some additional computations complexity, but thanks to some optimisations, this is limited to a ca. 10% increase in the overall runtime. I think that this is acceptable for now, and it is likely that further optimisations are possible.

The more flexible clusters of depth values also brought to light a lack of robustness in several other places in the code. I've also tried to make these more robust. In particular:
- We no longer try to fix OCR mistakes where a "1" has been incorrectly read as a "4". This is very rare, I could only find 3 examples in the Zurich and Geoquat datasets where this makes a difference (680244005-bp.pdf, A11486.pdf, B169.pdf). The "fix" actually misbehaves on two other inputs in those datasets (267123083-bp.pdf, 268124094-bp.pdf). Without the fix, all boreholes are still detected, it's just an individual depth value that is omitted.
- The Pearson correlation coefficient showed some unexpected behaviour for several inputs. In most places, I've replaced it with a mean-squared-error loss function for a linear regression, which seems more robust for our purposes.
- The logic for identifying and excluding regular scales (e.g. 1m, 2m, 3m, 4m, 5m....) has changes. Instead of detecting and removing these values in each cluster individually, we now remove such detected values for the whole page, and then re-apply the clustering without those values.

With all of these changes, the impact on the Benchmarking datasets is rather limited. The Layer F1 score improves by 0.6% on Geoquat and goes down by 0.3% on Zurich. I've looked at most of the inputs that are affected by the changes, and I did not see anything dramatic. Many of the affected inputs can be classified as edge cases or low-quality documents anyway.

The heuristic rules used in some methods (especially those that control how exactly values are clustered together and when a cluster is considered "valid") can seem a bit arbitrary. Indeed, it is currently not possible to really understand why these rules have been chosen in this way, and/or if an alternative approach is feasible, without doing extensive benchmarking and meticulously checking the impact of every small change on individual borehole profiles. The logic could definitely benefit from more explanations and unit tests, but since these changes are already more extensive than expected, I would suggest to postpone such improvements to follow-up issues.

### DIFF
--- a/src/extraction/features/stratigraphy/sidebar/extractor/a_above_b_sidebar_extractor.py
+++ b/src/extraction/features/stratigraphy/sidebar/extractor/a_above_b_sidebar_extractor.py
@@ -77,7 +77,7 @@ class AAboveBSidebarExtractor:
         Args:
             all_words (list[TextWord]): All words in the page.
             line_rtree (rtree.index.Index): Pre-built R-tree for spatial queries.
-            table_structures (list[TableStructure]): List of identified table-like sturctures on the page.
+            table_structures (list[TableStructure]): List of identified table-like structures on the page.
             used_entry_rects (list[pymupdf.Rect]): Part of the document to ignore.
             sidebar_params (dict): Parameters for the AAboveBSidebar objects.
 


src/extraction/features/stratigraphy/sidebar/extractor/a_above_b_sidebar_extractor.py
```diff
@@ -15,6 +17,52 @@
 class AAboveBSidebarExtractor:
     """Class that finds AAboveBSidebar instances in a borehole profile."""
 
+    @staticmethod
+    def _arithmetic_progression_entries(entries: list[DepthColumnEntry]) -> list[DepthColumnEntry]:
+        # First, try to find an artihmetic progression in the entries without decimal point
```
christianabbet:
```suggestion
        # First, try to find an arithmetic progression in the entries without decimal point
```

stijnvermeeren-swisstopo:
@AgathaSchmidt the issue in 267124070-bp.pdf comes from the OCR, where the numbers were not correctly detected.

PR #420, opened 2026-03-12
Improve borehole name extraction

### LINKED ISSUE #366: Improve borehole name extraction - Thurgau
In Thurgau, borehole name extraction is inconsistent across files and does not align with the ground truth (GT). In some documents, the borehole name is missing entirely, while in others there is no clear keyword or prefix (e.g. “Borehung”) to identify the name. Additionally, the GT uses the filename as the borehole name, which differs from the actual name detected in the document.

- `1000`, `1008`: Borehole name never mentioned in the document.
- `1011`, `1094`: No explicit keyword/prefix indicating a borehole name.
- GT vs detection mismatch:
	- `1032`: GT = EWS Kreuzlingen 557, detected = ESB
	- `1050`: GT = Wohnueberbauung Seevruga Ermatingen, detected = S2
	- `2724`: GT = SB2 Frauenfeld 61705, detected = B2

Should the GT be reviewed to match the borehole definition as in Zurich / Geoquat.

christianabbet:
Closes #366 

# Notes on Issue

In some cases, borehole title detection could not be improved due to the following limiting factors:

* **Missing name**: The borehole name is not present in the document (`TG:1000`, `TG:1008`)
* **Partially detected**: The borehole name is only partially detected by the algorithm, as in `TG:1032`. The trailing `1` is not detected because `ESB` is treated as the original name. Extending the logic to capture the following `1` would likely harm detection in other documents. We therefore keep the existing logic of looking on the same line only when no text is detected.

# Improvements

Modifications to the name extraction module to improve name detection:

* **New strict keywords**: `KB`, `RB`, `BS` and `B` are now recognised as valid keywords for borehole names. These use strict matching, meaning the detected text must contain and solely represent the keyword. For example, `BS 1` is valid, but `KERBS 1` is not. (`TG:1099`, `TG:1122`)
* **Improved evaluation**: Name matching now aligns with the rules of `_clean_borehole_name`. Spacing, keywords, parentheses, scale, and punctuation are ignored when comparing two names, For example, `KB1` == `KB.1  (1:100)` (`GQ:B260`).
* **Pattern matching**: We investigated imposing a minimum of one digit in borehole names. However, this would break valid cases where names are defined as `RB II`, `RB IV` or `KB A`. Creating a rule to cover all such cases would likely be counterproductive. The constrain on name length solve this.
* **Name length**: For some documents, the borehole name is too long to be captured. The internal threshold limits it to 10 characters. The limit was increased to 20. On top of that, we add a limit on words length within the name. Any words (alphabetical) longer than 3 characted will cause the name to be flagged as false. Solves `TG:2532`.
* **Page range**: Metadata is now required to fall within the borehole's page range. This is particularly important for documents containing multiple boreholes, where features spanning multiple pages could otherwise be incorrectly assigned to a different borehole. This constraint removes many false positives in long documents (`TG:1061`, `TG:1122`, `TG:1163`).
* **Language**: An edge case in language detection has been fixed. Previously, if majority voting resulted in a tie between two languages, the assignment was not deterministic.
A significant number of undetected titles in Thurgau are linked to cases where no boreholes are detected at all (files: `TG:1587`, `TG:111`, `TG:1265`, `TG:1648`, `TG:1839`). These are expected to improve as borehole false negatives are reduced. 
* **Case**: For inner keywords we impose case matching as they represent abbreviations (KB = Kernbohrung)

Since thurgau includes multiple files with multiple boreholes, it may be worth exploring a way to indicate which name is matched to which borehole, either color or marker-based.

# Metrics

A positive influence is also observed on other metadata such as coordinates, which are now detected within the borehole page range (e.g. `TG:1102`, `TG:2307`, `TG:2371`, `TG:25`).

## Name

* `zurich`: `71.90` → `74.78` (+2.88%) (affected files positively: `ZH:269126143, ZH:268124232, ZH:267123080, ZH:267124097`, `ZH:268125303`, `ZH:268124112`, negative: `ZH:684252058`, `ZH:691251009` - distinct borehole, different names )

* `geoquat`: `51.12` → `61.40` (+10.38%) (affected files positively: `GQ:A68`, `GQ:A84`, `GQ:A167`, `GQ:A223`, `GQ:A253`, `GQ:A277`, `GQ:A356`, `GQ:A435`, `GQ:A418`, `GQ:A437`, `GQ:A551`, `GQ:B140`,  `GQ:B156`, `GQ:B167`, `GQ:B205`, `GQ:B207`,`GQ:B260`, ... and more, neutral: `GQ:A11452`, negative:  `GQ:A525`, `GQ:A263`, `GQ:1488` - borehole detected as two distinct entities, with the second missing coordinate/elevation on its page.)

* `thurgau`: No metric available, but improvements related to the items described above can be observed eg: `TG:1099`, `TG:1107`, `TG:2494`.

## Elevation

* `zurich`: `83.54` → `83.18`
* `geoquat`: `62.87` → `62.67`  

## Coordinates

* `zurich`: `70.30` → `70.55`
* `geoquat`: `48.50` → `48.64`

### DIFF
--- a/src/swissgeol_doc_processing/utils/language_filtering.py
+++ b/src/swissgeol_doc_processing/utils/language_filtering.py
@@ -55,8 +55,6 @@ def match_any_keyword(
 ) -> re.Match | None:
     """Search for the first occurrence of any keyword from a predefined list in a text.
 
-    Keywords are treated as **raw regex patterns**.
-
     Args:
         text (str): The text to search within.
         keywords (list[str]): A list of regex patterns to look for.


src/swissgeol_doc_processing/config/name_detection_params.yml
```diff
@@ -27,7 +31,11 @@ matching_keywords: [
     "bohrung",
     "schacht",
     "sondierung",
-]
+    "ews\\-anzahl",
```
stijnvermeeren-swisstopo: I also had to check this with my colleagues, but unfortunately, it turns out that this is not correct.

Image

This refers to the fact that two geothermal probes were inserted next to each other, but only for one of those two boreholes, a protocol was written down. Since it is not an identifier of a specific borehole, it is not appropriate to extract this number 2 as a borehole name.

Issue #378, opened 2026-01-29
Reduce borehole FN - Thurgau


christianabbet:
Detect boreholes that are currently missed entirely. It seems that a lot of Borehole profiles that are presented as tables are missed. One soltuion could be to implement a new sidebar type in the same fashion as `a_above_b_sidebar` or `a_to_b_sidebar`. 

Note that similar structure can be observe in `Zurich` (eg: `10626`). Where there are no striplogs present, It might be a start for the investigation.

<img width="475" height="269" alt="Image" src="https://github.com/user-attachments/assets/9fc90120-fccd-478e-a48d-2f0b68da11b9" />

We can observe this new potential layout for `Thurgau` dataset in `359`, `5629`, or `5910`

<img width="706" height="228" alt="Image" src="https://github.com/user-attachments/assets/dcdca11a-4a7e-459a-bd3c-aadebdad5d4a" />
<img width="802" height="99" alt="Image" src="https://github.com/user-attachments/assets/8ac93314-1c8f-4cf7-9bd8-6a74d1f5b19c" />
<img width="724" height="276" alt="Image" src="https://github.com/user-attachments/assets/8cc9a48b-32ae-422e-a279-640c126d838d" />

PR #422, opened 2026-03-16
Feat/issue 417 create new protocolsidebar

### LINKED ISSUE #417: Create new ProtocolSidebar
Based on #378 #403 

The current implementation of the AaboveBSidebar is failing for a special borehole type called "Bohrprotokoll". 
The reason for this is a missing striplog and more importantly: no proportionality between the vertical position of the depth value and their value. 
There can probably be some share logic between the new sidebar type and the AAboveBSidebar, but with the following notable differences:

- The new sidebar should have no requirements about correlation between vertical position and depth value (no Pearson correlation coefficient, no linear fit loss...)
- The alignment between the depth values and the material description should be simpler, like with the AToBSidebar.
- To avoid false positives, we should probably require that we only create depth value clusters inside a table-like structure and below a certain keyword (e.g. "Tiefe"). The list of such keywords should be configurable.
- The protocol sidebar could be more relaxed for arithmetically progressing sidebars. e.g. 1347, 1360, 1363,...
- Less layers required, sinde there are multiple boreholes with only 2 layers. E.g. 1244, 1271, 1305, .... 

Image

Image

AgathaSchmidt:
# Created new protocol sidebar 
Has the following characteristics: 
- min layers: 2 
- no dependency on vertical position 
- expects a table like structure 
- has only strictly increasing values
- matching with material description similar to AtoB sidebar
- every depth interval has to be matched to one material description
- when a sidebar has the end depth duplicated e.g. [12, 155, 155] we trim the last value so it becomes a valid sidebar and is not rejected because it does not fulfill the condition of steady increase

Not sidebar dependent changed: 
- added measuring of the angles of the left and right edges of the bounding boxes of the depth column entries to better suit left or right aligned text/table columns

# Effect on main dataset 
In our FN subset (n=80) we are detecting additionally to the improvements from #378  the following 38 files now: 
- 1244, 1271, 1305, 1325, 1326, 1327, 1330,  1351, 1359, 1360, 1363, 1377, 1392, 1402, 1415, 1422, 1425, 1466, 1498, 1521, 1543, 1608, 1646, 1762, 1770, 1839,  2046, 264, 4115, 4120, 4122, 4123, 4124, 4127, 4139, 4142 detected 
- 1416 detected (not completely correct)
- 1347 different parts detected 

depth_intervall_f1 from 0.165 -->0.506

# Side effects on other datasets
(Version of main 17.03 15:00)

Negative effect: 
- 268124125 material description lost on page 2 due to FP in the bottom of the page
Screenshot 2026-03-30 at 08 06 17

<google-sheets-html-origin>
Zurich | layer_f1 | depth_intervall_f1 | Material_description_f1 | elevation_f1 | coordinates_f1 | Borehole_name_f1
-- | -- | -- | -- | -- | -- | --
main | 0.841 | 0.893 | 0.879 |  0.828 | 0.701 | 0.715
branch | 0.841 | 0.893 | 0.879 |  0.828 | 0.701 | 0.715

**Geoquat/val**
Positiven (?) effect: 
- 1506.pdf first two depth entries detected additionally 
This branch: 
Screenshot 2026-03-30 at 08 09 30

Main: 
Screenshot 2026-03-30 at 08 11 37

Negative effect: 
-  A437.pdf for page 2 scale sidebar (because we have every interval matched with one material description )
Screenshot 2026-03-30 at 08 08 45

<google-sheets-html-origin>
Geoquat/val | layer_f1 | depth_intervall_f1 | Material_description_f1 | elevation_f1 | coordinates_f1 | Borehole_name_f1
-- | -- | -- | -- | -- | -- | --
main | 0.4794 | 0.631 | 0.629 | 0.617 | 0.475 | 0.501
branch | 0.479 | 0.6296 | 0.628 | 0.613 | 0.475 | 0.501


stijnvermeeren-swisstopo:
@AgathaSchmidt The examples that you mentioned, requiring more tolerance on the skew angle, are actually not skewed at all, just left-aligned instead of center-aligned. However, currently, the angle is computed from the midpoints of the first and last bounding boxes, so this leads to a rather diagonal line, even though the numbers are actually nicely (left-)aligned.

image

How about we keep the threshold at 5 degrees, but we also compute the angle of the line that connects the left-edge of both bounding boxes (any maybe even the rightmost edge as well, to deal with right-aligned values) instead of just the line that connects the midpoints? And as soon as one of these lines is less than 5 degrees from vertical, then we accept?

tests/test_protocolsidebar.py
```diff
@@ -0,0 +1,26 @@
...
+    """Test the trim_trailing_duplicate_depths method of the ProtocolSidebar class."""
+    sidebar = ProtocolSidebar(
+        [
+            DepthColumnEntry(rect=pymupdf.Rect(), value=5, page_number=0),
+            DepthColumnEntry(rect=pymupdf.Rect(), value=155, page_number=0),
+            DepthColumnEntry(rect=pymupdf.Rect(), value=155, page_number=0),
+        ]
+    )
+    sidebar = sidebar.trim_trailing_duplicate_depths()
+    assert ProtocolSidebar(
```
christianabbet: The assertion only check the object is created ? I guess here we want to check if the triming is successful. Shouldn't it be something like :

```python
    sidebar = ProtocolSidebar(
        [
            DepthColumnEntry(rect=pymupdf.Rect(), value=5, page_number=0),
            DepthColumnEntry(rect=pymupdf.Rect(), value=155, page_number=0),
            DepthColumnEntry(rect=pymupdf.Rect(), value=155, page_number=0),
        ]
    )
    sidebar_trimmed = ProtocolSidebar(
        [
            DepthColumnEntry(rect=pymupdf.Rect(), value=5, page_number=0),
            DepthColumnEntry(rect=pymupdf.Rect(), value=155, page_number=0),
        ]
    )
    sidebar = sidebar.trim_trailing_duplicate_depths()
    assert sidebar== sidebar_trimmed
```

christianabbet (changes requested):
Good results with the new sidebar. I think the the small drop in performance is more than acceptable compared to the benefits.

stijnvermeeren-swisstopo (changes requested):
Nice work! I just have a few ideas for simplification / reducing duplicated code, and a couple of small questions.

stijnvermeeren-swisstopo (changes requested):

stijnvermeeren-swisstopo (approved):
I'm happy with the PR now :). No impact on ZH and Geoquat, and a significant positive impact on the Thurgau false negatives. Code quality is great now as well.

For TG, I see quite a few examples where the borehole is generally correctly identified, but the alignment between depths and descriptions is off-by-one (usually the first depth interval has no description, and the last one has two lines). E.g. here Thurgau ID 1360: 
Image

Do you know why this happens? If you don't have a quick solution, then we can create a follow-up ticket as well. If there's a follow-up issue, then I'd also be happy to merge this PR as-is.

christianabbet (approved):

PR #425, opened 2026-03-23
Improve code sharing between extraction and classification

### LINKED ISSUE #389: Improve code sharing between extraction and classification
The pull request for multi benchmarking for classification #373  and extraction #360 have the potential to share more code. 

**runner.py's**
There's a lot of potential to share more code between extraction/runner.py and this file classification/runner.py. For example, with the exception of the default experiment name, the methods look  identical. A cleaner solution would be to have a parent Runner class for all the shared functionality, and then create a child class for extraction and classification respectively, only containing custom implementations. This would make the code more maintainable and easier to comprehend (as you would no longer need to worry about potential differences between classification and extraction).

**classification parameters**
Since we have two different modes for classification, namely classify-descriptions and classify-predictions, we habe different benchmark spec for running classification pipeline vs running the extraction pipeline. 

For the first case that means that -f file_path is input and ground truth at the same time
But optionally, if we do not want to classify all descriptions given in the ground truth we can specify a subset in -s file_subset_directory

In the second case -f file_path is the input which are the material description predictions from our model and the ground truth is given by the -g flag

Maybe we can unify it somehow so for the first case we also use -f and -g like this:
-f ground_truth.json
-g ground_truth.json
if we want to classify all descriptions in the gt
Or
-f file_subset_directory
-g ground_truth.json
if we only want to classify the subset. 

This could enable us to use the same BenchmarkSpec class for both, extraction and classification 
Or we could think of a BenchmarkSpec and BenchmarkSummary parent class

AgathaSchmidt:
# Changes
**One classification mode only**  
_Before_ we had a description classification mode and a prediction classification mode, where either the ground truth descriptions or the predicted material descriptions are classifies respectively. 
This also came along with different input parameters. 
We had -f file path, -g ground truth path and -s subdirectory path. 
_Now_  it is all unified with only -f file path and -g ground truth which makes it analog to the extraction parameters enabling more code sharing. 
The -f file path can be anything now: 
- a ground truth path so we extract the description and classify them
- a subset of ground truth 
- a prediction.json of either the full dataset or a subset

To account for that, I needed to add new helper function load_and_format_input_data () in src/classification/utils/data_formatter.py which reads the file_path and formats the data respectively to the type/mode. 

The launch.json template was updated as well.

**Shared functions and classes**
The shared utilities are defined in **src/core/benchmark_utils.py** and then imported from there in the runner.py files of classification and extraction. 
- finalize_overall_summary(): before, each had their own _finalize_overall_summary. Now they share one and when calling it, only `aggregate_label` is different. 
- parent_input_key(): Generate a stable parent input key for a group of child inputs.
- prepare_pipeline_temp_paths(): new function to share temp path handling
- class BenchmarkSummary: new parent class for ExtractionBenchmarkSummary and ClassificationBenchmarkSummary defining metrics_flat and class attributes
- class Metrics: was actually used by both pipelines before but was defined in extraction.evaluation.evaluation_dataclasses. Now it was moved to core.

**src/core/mlflow_utils.py**

- setup_mlflow_tracking()
- setup_mlflow_parent_run()
Which are both called in both runner.py's

**src/core/pipeline_runner.py**
- class PipelineRunResult: container for pipeline prediction output --> extracted boreholes or classification predictions 
- class PipelineRunner: base for pipeline execution with abstract methods for running predictions logic and evaluation
- class MultiBenchmarkRunner: base for orchestrating multiple benchmarks with shared MLflow tracking with abstract methods for running the single benchmark and summarizing


christianabbet (commented):
Great work! Not easy to properly review as it modify a lot of files. I think the overall logic seems good. Tested on my side and everthing seems to be working. Just some minor comments / typing issues. 

PS: Is there a reason to use the bare asterisk in function calls to impose explicit arguments ? (eg prepare_pipeline_temp_paths, finalize_pipeline_run, start_or_resume_mlflow_run, ...)

src/core/benchmark_utils.py
```diff
@@ -57,18 +67,277 @@ def configure_logging(level: int = logging.INFO) -> None:
...
+    Returns:
+        Group key string.
+    """
+    inputs = " | ".join(sorted(relative_after_common_root(paths)))
+    return f"multi:{inputs}"
+
+
+def finalize_overall_summary(
+    *,
+    overall_results: list[tuple[str, Any | None]],
```
stijnvermeeren-swisstopo: can we use the type `BenchmarkSummary` here instead of `any`?

src/extraction/evaluation/benchmark/score.py
```diff
@@ -23,72 +22,43 @@
...
+class ExtractionBenchmarkSummary(BenchmarkSummary):
+    """Summary of a single extraction benchmark."""
 
-    ground_truth_path: str
-    n_documents: int
     geology: dict[str, float]
     metadata: dict[str, float]
 
-    def metrics(self, short: bool = False) -> dict[str, float]:
+    def metrics_flat(self, prefix: str = "metrics", short: bool = False) -> dict[str, float]:
```
stijnvermeeren-swisstopo: I don't see the purpose of this new `prefix` arg, I can't find any usages of this method that define any value for this arg.

Also, I don't fully understand why we need the `metrics/` prefix at all, instead of just `f"{category}/{metric}"`. The only place that I could find where the `metrics/` prefix ends up is in the file `overall_summary.csv`. Is there a pressing reason to have the prefix in the column headers of that CSV file?

src/classification/runner.py
```diff
@@ -263,205 +96,148 @@ def run_predictions(
...
+    def setup_parent_run(self, runid: str | None) -> str:
+        return setup_mlflow_parent_run(
+            run_id=runid,
             experiment_name="Layer descriptions classification",
+            parent_input_key=parent_input_key([Path(b.file_path) for b in self.benchmarks]),
+            benchmarks=self.benchmarks,
+            input_tag_name="json_file_path",
+            out_directory=self.multi_root.parent,
+            include_git_info=False,
         )
```
stijnvermeeren-swisstopo: The only difference between this method in the classification and extraction runner seem to be a few parameters. So I would try to define this `setup_parent_run` method directly in the `MultiBenchmark` parent class, and then instead of having a method definition in each child class, only have a couple of variables like `experiment_name` and `input_tag_name` that are defined at the child class level. Do you think this would be possible?

If so, could you also check if there are other method definitions that could be similarly simplified/harmonised elsewhere in this PR? (Probably `finalize_overall_summary` would be another good candidate?)

src/core/mlflow_utils.py
```diff
@@ -0,0 +1,113 @@
...
+
+def setup_mlflow_tracking(
+    *,
+    run_id: str | None,
+    experiment_name: str,
+    runname: str | None = None,
+    nested: bool = False,
+    tags: Mapping[str, Any] | None = None,
+    params: Mapping[str, Any] | None = None,
+    include_git_info: bool = False,
```
stijnvermeeren-swisstopo: I would be in favor of always including git info (i.e. remove this flag from the method signature)

src/core/benchmark_utils.py
```diff
@@ -57,18 +67,277 @@ def configure_logging(level: int = logging.INFO) -> None:
...
+        paths: Child input paths.
+
+    Returns:
+        Group key string.
+    """
+    inputs = " | ".join(sorted(relative_after_common_root(paths)))
+    return f"multi:{inputs}"
+
+
+def finalize_overall_summary(
```
stijnvermeeren-swisstopo: Since this is only used in the MultiBenchmarkRunner classes, might it make more sense to define this method directly on the `MultibenchmarkRunner` parent class?

src/classification/evaluation/benchmark/spec.py
```diff
@@ -10,75 +10,39 @@ class BenchmarkSpec:
 
     Attributes:
         name: Human-readable benchmark name (used for folder/run naming).
-        file_path: JSON file to classify (ground truth JSON or predictions JSON).
-        file_subset_directory: Folder containing subset PDFs; used to filter by filenames.
-        ground_truth_path: Optional ground truth JSON path if file_path is a predictions JSON.
+        file_path: Input path to classify. Can be:
+            - a full ground truth JSON
+            - a subset ground truth JSON
```
stijnvermeeren-swisstopo: I'm still a little confused about how the "subset ground truth JSON" is defined. In this case, does the file_path refer to the directory that defined the subset, rather than to the JSON file itself (which can also be passed to ground_truth_path)? Please clarify in the docstring.

Actually, `load_and_format_input_data` in `data_formatter.py` has a much clearer docstring. Please refer to that documentation here.

Also, the way these options are defined might be something that should be updated / documented in more details in the markdown documentation files?

stijnvermeeren-swisstopo (changes requested):
This is a big improvement in terms of code organisation and readability already!

I tested different ways of running extraction, and all still seems to be working normally. I did not really test classification, but I trust that this has been sufficiently tested already as well.

Here are a couple of additional points that could be addressed. Once these are solved, then I think that this topic of code sharing between extraction and classification can be considered to be solved for the time being.

src/classification/evaluation/benchmark/spec.py
AgathaSchmidt: The "subset ground truth JSON" is an altered/reduced ground truth json file. But actually this does the same thing as using the subset directory as input. 
And in application, we will never change the gt json to include only specific files and use it as input. 
I think this is an artifac from when I just started to get to know the classification pipeline and I missunderstood some things. (I created a GT file containing only one entry and tested the pipeline)
I will remove the comments about subset ground truth JSON".

README.md
```diff
@@ -196,58 +196,68 @@ The script produces output in two different formats:
...
+    Supported modes:
+        1. input_path is a directory:
+            - interpreted as subset directory
+            - ground_truth_path must be provided and must be a ground truth JSON
+        2. input_path is a JSON and ground_truth_path is None:
+            - input_path must be a full ground truth JSON
+            - same file is used as descriptions and ground truth
+        3. input_path is a JSON and ground_truth_path is provided:
+            - if input JSON is ground-truth-like: use it directly as descriptions input and compare to full GT
+            - if input JSON is predictions-like: format it and compare to GT
```
stijnvermeeren-swisstopo: In looks like this is not formatted correctly on GitHub:
https://github.com/swisstopo/swissgeol-boreholes-dataextraction/blob/5d44f9288f794e61b3fb48e05c457afe57410533/README.md

src/core/pipeline_runner.py
```diff
@@ -158,20 +163,14 @@ def _parent_runid_tmp(self) -> Path:
...
-        """Start or resume the MLflow parent run. Must be overridden when MLflow tracking is active.
-
-        Args:
-            runid: Previous parent run id to resume, or None to start a new run.
-
-        Returns:
-            The active MLflow parent run id.
-        """
-        raise NotImplementedError(f"{type(self).__name__} does not implement setup_parent_run")
+        finalize_overall_summary(
```
stijnvermeeren-swisstopo: Since this method `finalize_overall_summary` is only used here, why not remove it from `benchmark_utils.py` and put the contents directly here? Or do you see any value in keeping it in `benchmark_utils.py`?

The same can be said about `setup_mlflow_parent_run` and `finalize_pipeline_run`.

stijnvermeeren-swisstopo (changes requested):

src/core/pipeline_runner.py
AgathaSchmidt: yes you are right, I moved them and made them class methods so they require few input parameters and can use more from self.

stijnvermeeren-swisstopo (approved):
I pushed a small cleanup for the README documentation of the different classification modes.

Now everything looks good to me.

PR #432, opened 2026-03-26
Fix multi page spanning line in visualization

### LINKED ISSUE #416: Fix multi-page spanning line in visualization
The newest changes introduced in #406 , where the correct depth values are now detected for multi page boreholes lead to undesired visualization behavior. 
Since the borehole spans over multiple pages the line that now spans from one page to the other for the elevation is a bit redundant with the side indicator for the borehole span, e.g. 1035_Kreuzlingen_EWS_Wohnueberbauung_Schrofen-Nord (Thurgau dataset).

Image

AgathaSchmidt:
## Before
The previous implementation drew the full background_rect for depth highlights without accounting for layers that span multiple pages. This caused visual highlight-"pillars" stretching the full page height for multi-page boreholes. 

Image

## Changes:

This PR fixes that by 
- clipping the highlight to only cover the depth entries actually visible on the current page. 
- It also removes the connector line between a depth entry and its material description in these cases, since the span marker arrow already makes the connection clear.
- To support this, draw_layer now receives the bounding boxes of the visible depth entries on the page, which are used to figure out how far the highlight should extend.

Screenshot 2026-03-26 at 13 45 45

### DIFF
--- a/src/extraction/annotations/draw.py
+++ b/src/extraction/annotations/draw.py
@@ -294,7 +294,7 @@ class PageDrawer:
                             pymupdf.Point(mat_descr_rect.x0, (mat_descr_rect.y0 + mat_descr_rect.y1) / 2)
                             * self.page.derotation_matrix,
                         )
-                    self.shape.finish(color=pymupdf.utils.getColor(color))
+                        self.shape.finish(color=pymupdf.utils.getColor(color))
 
                 # Get depth highlight rectangle
                 background_rect = layer.depths.get_background_rect(page_number, self.shape.page.rect.height)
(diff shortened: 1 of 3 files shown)


src/extraction/annotations/draw.py
```diff
@@ -265,29 +270,54 @@ def draw_layer(
...
+                if background_rect is not None:
+                    clipped_background_rect = pymupdf.Rect(background_rect)
+
+                    # Avoid full-page "pillars" for spanning layers:
+                    # clip the vertical interval to the actual depth-entry extent visible on this page.
+                    if depth_entry_rects:
+                        # Build one combined rectangle covering all detected depth entries on this page.
+                        page_depth_entries_rect = pymupdf.Rect(depth_entry_rects[0])
+                        for rect in depth_entry_rects[1:]:
+                            page_depth_entries_rect.include_rect(rect)
```
christianabbet: pymupdf support the operator `|` for the union of two rectangles. So you can do something like:
```python
from functools import reduce
page_depth_entries_rect = reduce(lambda a, b: a | b, depth_entry_rects)
```

christianabbet (commented):
Small code comments

stijnvermeeren-swisstopo (changes requested):

stijnvermeeren-swisstopo (approved):
LGTM!

PR #444, opened 2026-04-07
Memory increase after UV

### LINKED ISSUE #443: Memory increase after pip to uv migration
Following the migration from pip to uv, memory usage appears to have doubled from ~2GB to ~4GB. We want to investigate the root cause of this increase, as the uv migration itself should not have had a significant impact on memory consumption. It seems that this issue does not only affect assets and find its root in the Borehole package after UV merge (1.0.157).

| Source | Version  | Image Size (GB) | Memory Idle (GB) |
|--------|----------|-----------------|------------------|
| Local  | 1.0.156 	| 1.48 			  | 0.445   		 |
| GitHub | 1.0.156 	| 1.56			  | 0.617			 |
| Local  | 1.0.157 	| 1.15			  | 1.65			 |
| GitHub | 1.0.157 	| 1.23 			  | 2.78			 |

christianabbet:
Close #443 

# Description

We compare the memory usage from different build and settings

| Source | Version  | Image Size (GB) | Memory Idle (GB) |
|--------|----------|-----------------|------------------|
| Local  | 1.0.156 	| 1.48 			  | 0.445   		 |
| GitHub | 1.0.156 	| 1.56			  | 0.617			 |
| Local  | 1.0.157 	| 1.15			  | 1.65			 |
| GitHub | 1.0.157 	| 1.23 			  | 2.78			 |
| | | | |
| Local  | 1.0.0-dev4  	| 1.38 			  | 0.444			 |
| GitHub  | 1.0.0-dev4   	| 1.52			  | 0.530	 |

Versions prior to the UV merge (< 1.0.157) show significantly lower memory usage. Despite extensive investigation, the root cause remains unclear.

# Investigation

When investigating memory consumption, most of the increase came from `compound_split`, which added ~1.5 GB of memory. Interestingly, in previous versions the same import only added ~300 MB. After further investigation, the problem appears to affect all packages. For example, `PyMuPDF` now takes ~150 MB to load instead of ~100 MB previously.

The root cause lies in how Python handles compiled packages. By default, UV generates "lazy" compiled imports that are not optimized. Adding the `--compile` flag forces UV to run extensive compilation of packages and generate .pyc files, which drastically reduces memory consumption.

On top of that, the Dockerfile was refactored to make it more straightforward and aligned with the assets repo. `pyinstrument` and `moto` were moved to optional dependencies, as they are not needed for the main package run.


AgathaSchmidt (commented):
Do you think it might be worth adding requirements.txt to .gitignore so we do not accidentally commit it again?

Dockerfile
```diff
@@ -42,4 +48,4 @@ EXPOSE 8000
 HEALTHCHECK CMD curl --silent --fail http://localhost:8000/health || exit 1
 
 # Command to run the FastAPI Borehole app with Uvicorn
-CMD ["uvicorn", "app.main:app", "--host", "0.0.0.0", "--port", "8000"]
+CMD ["uv", "run", "uvicorn", "app.main:app", "--host", "0.0.0.0", "--port", "8000"]
```
letao: Why is the "uv", "run" needed here? Would it not work without it? Unless there is a clear reason we could remove this as redundant.

Dockerfile
```diff
@@ -1,39 +1,45 @@
...
+# --no-dev: Exclude development dependencies from the environment
+# --no-install-project: Skip building and installing the project package itself
+# --compile: Force generation of compiled files *.pyc (lowers memory load)
+RUN uv sync --no-dev --no-install-project --compile
 
-## Runtime stage
-FROM python:3.12-slim
+
+## ------ Runtime stage
+FROM ghcr.io/astral-sh/uv:python3.12-bookworm-slim AS runtime
```
letao: If we can remove "uv", "run" in the comment bellow could we also revert back to FROM python:3.12-slim?

letao (commented):
In general looks good, just two minor comments on the dockerfile.

PR #445, opened 2026-04-08
on_file_done callback for per-file csv writing and drawing

### LINKED ISSUE #423: Introduce a callback method for runner.py to cover csv and visualization output
After https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/407 the logic for extraction has been decoupled from mlfow and visualization. However, the logic still remains in the **[src/extraction/runner.py](https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/407/files/74f60431a00627850f1aa5528b732528e2b3969d#diff-8798f1d876a6c854fe650819d5719066864222d118be32d3e43635703ec8e2c4)** script.

Since we anyways want to create csv and visualization outputs as soon as a file is processed, and not only at the very end of the benchmark, it is not possible to move that outside of the for loop in the current method. 

Instead of removing these calculations, we should allow for a "callback method" that can be passed as a parameter to run_predictions, and where the caller can specify what kind of side-effects based on the extraction results are desired.

We can imagine something along these lines for the implementation

```
PostProcess = Callable[[ExtractionResult], None]

def run_predictions(
    input_directory: Path,
    out_directory: Path,
    predictions_path_tmp: Path,
    part: str = "all",
    analytics: MatchingParamsAnalytics | None = None,
    on_file_done: PostProcess | None = None, 
) -> tuple[OverallFilePredictions, int]:
```

AgathaSchmidt:
Closes #423

## Changes 
We separated the prediction loop from the csv writing and drawing, so we can save the csv and draw the png after each file is processed during prediction. 
(I want to notice that since the opening of the issue some restructuring of the code happened, due to #425 )

- We introduce the `on_file_done` callback to `run_extraction_prediction`, where the csv writing and drawing logic was before. 
- csv writing (`write_csv_for_file`) and drawing (`draw_file_predictions`) are now handled by a `PostProcess` callback defined in `ExtractionPipelineRunner.run_predictions`, keeping `run_extraction_predictions` focused solely on PDF processing and prediction
- `run_extraction_predictions`does not return `all_csv_paths` anymore


src/extraction/runner.py
```diff
@@ -159,11 +162,11 @@ class ExtractionOptions:
 
 def run_extraction_predictions(
     input_directory: Path,
-    out_directory: Path,
     predictions_path_tmp: Path,
     options: ExtractionOptions,
     analytics: MatchingParamsAnalytics | None = None,
-) -> tuple[OverallFilePredictions, int, list[Path]]:
+    on_file_done: PostProcess | None = None,
```
christianabbet: I think there is no need for a global type here. We can jsut put it in the text
```suggestion
    on_file_done: Callable[[ExtractionResult], None] | None = None,
```

src/extraction/runner.py
```diff
@@ -274,12 +261,39 @@ def setup_mlflow_run(self, runid: str | None) -> str:
...
+        csv_paths: list[Path] = []
+        any_draw = (
+            not self.options.skip_draw_predictions
+            or self.options.draw_lines
+            or self.options.draw_tables
+            or self.options.draw_strip_logs
+        )
+        input_is_file = self.input_directory.is_file()
+
+        def on_file_done(result: ExtractionResult) -> None:
```
christianabbet: It's a bit odd to have the function in the middle of the code. Maybe we could define a partial function with the predifined arguments


```python
# Define full function 
 def on_file_done(result: ExtractionResult, out_directory: str, input_directory: str, ......) -> None:
    ...

# Create instance of function within the code and fix already all args expect results
callback_fn = partial(on_file_done, out_directory=self.out_directory, input_directory=self.input_directory, ...)

# PAss callback to function
predictions, n_documents = run_extraction_predictions(
    input_directory=self.input_directory,
    predictions_path_tmp=predictions_path_tmp,
    options=self.options,
    analytics=self.analytics,
    on_file_done=callback_fn,
)
```

christianabbet (changes requested):
Most of the comment is how to handle the callback and avoid having it locally with local variables

src/extraction/runner.py
```diff
@@ -157,13 +159,59 @@ class ExtractionOptions:
...
+        skip_draw_predictions (bool): Skip drawing final prediction overlays.
+        draw_lines (bool): Draw detected lines.
+        draw_tables (bool): Draw detected table structures.
+        draw_strip_logs (bool): Draw detected strip log structures.
+    """
+    if write_csv:
+        csv_paths.extend(write_csv_for_file(result.predictions, out_directory))
+    if any_draw:
+        pdf_path = input_directory if input_is_file else input_directory / result.predictions.file_name
+        draw_file_predictions(
```
stijnvermeeren-swisstopo: Is there any value of having `draw_file_predictions` as a separate method? It feels to me as if the code would become more organised and readable if we just include the contents of `draw_file_predictions` into this method here.

stijnvermeeren-swisstopo (changes requested):
It doesn't really feel like as much of an improvement yet as I was hoping for. Is it possible to define the callback method already earlier on in the code flow, for example in main.py before we instantiate the `ExtractionPipelineRunner`? Then we could pass the callback function already to the `ExtractionPipelineRunner` class, and many of the current options of the `ExtractionPipelineRunner` (like `draw_lines`, `draw_tables`, etc.) would become redundant since they are only used in the callback function anyway. Could you check if such a refactoring is posslbe?

src/extraction/runner.py
AgathaSchmidt: Yes, I removed it

Also, I moved on_file_done to `benchmark_utils.py` (I wanted to keep the `main.py` as minimal as possible) and `ExtractionPipelineRunner` now accepts an `on_file_done: Callable[[ExtractionResult], None]` field directly. 
`ExtractionOptions` does not hold the drawing specific parameters anymore. 
instead I added `make_callback_factory` to translate the CLI flags into a callback factory once, which can then be shared with single- and multi-benchmarks. 

Let me know what you think

stijnvermeeren-swisstopo:
@AgathaSchmidt I allowed myself to directly push my proposed changes: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/commit/f43ba415389c48996a4963cfed205c849fc25d20

It uses a class instead of partial method application and removes some redundant wrapper methods (in partiuclar, puting the contents of the `run_extraction_predictions` method directly inside the `ExtractionPipelineRunner`).

Let me know what you think!

AgathaSchmidt:
Thank you @stijnvermeeren-swisstopo 
It looks good to me!

stijnvermeeren-swisstopo (approved):

christianabbet (approved):

PR #451, opened 2026-04-10
api endpoint for classification of lithological attributes from a plain text description

AgathaSchmidt:
# Summary 
- Created a classification endpoint in the API 
- Split models into backbone and head which can be used for the creation of a new Docker image 
- upload the models to the repository using LFS
- Adjusted the code so it can handle the split model approach 
- Updated the API/Docker docs
- add `BERT_ENABLED=false` to `.env` so we can skip `load_models()` of the classification models, run the API - without the classification endpoint and safe memory
- changed the classification pipeline so that it also used backbone and head --> we don't need to have the full model downloaded to run 

More information about the model decoupling process can be found here: 
`stijnvermeeren-boreholes-models/decoupled_classification.md`

# Model decoupling
Both BERT models (`lithology`, `en_main`) were fine-tuned from the same base checkpoint with layers 0–10 frozen. Their backbone weights are byte-for-byte identical, so storing two full models was redundant.
This is why we split the backbone from task-specific heads, reducing baked-in model weights significantly. 
Removed uscs as a supported classification system from the API.

# Changes
We created  `src/scripts/model_decoupling.py`, a  one-time script which downloads both models from S3, verifies backbone identity, splits and re-uploads as`backbone/, lithology_head/, en_main_head/`
`Dockerfile` — COPY backbone once + two heads instead of two full models
`src/classification/models/model.py` — BertModel accepts optional backbone_path; merges backbone + head weights via load_state_dict at init
`src/app/api/v1/endpoints/lithology_classification.py` — resolves `BERT_MODEL_PATH_BACKBONE` env var; removed uscs from model path maps
`src/app/common/schemas.py` — removed "uscs" from classification_system Literal type
Notes

The decoupling script has already been run and the split models are uploaded to S3.

## Current memory numbers
Disk:             1346 MB (full)  →  707 MB (split)  (47% smaller)
Steady-state RAM: 829 MB (full)  →  726 MB (split)  (102 MB less)
Inference peak:   +22.0 MB (full)  →  +6.0 MB (split) 

The approach of how models are loaded is scaling well and becomes very beneficial with increasing number of models, e.g. n= 12 full model RAM   4275  , split model RAM   1316   saving -->     69%

# Test 
Since the command from the `API_and_Docker.md` will only work once this branch is merged and the new offical Docker image is generated do this in the meantime to test the API with the new classification endpoint based on the split models: 
Export your keys: 

```bash 
docker build -t borehole-api .
docker run --env-file .env  -p 8000:8000 borehole-api

```

**BERT_ENABLE is defined in the .env file which is why you have to run it with the --env-file flag** 
Then go to `http://127.0.0.1:8000/docs#/` to test the API


.env.template
```diff
@@ -9,4 +9,15 @@ AWS_DEFAULT_REGION=your_default_region
...
 ANTHROPIC_MODEL_ID="your_model_id"
-ANTHROPIC_MODEL_REGION="anthropic_region"
\ No newline at end of file
+ANTHROPIC_MODEL_REGION="anthropic_region"
+
+# BERT model locations for the lithology classification endpoint.
+# Option A — local directory path (useful after a manual download or local training):
+#   BERT_MODEL_PATH_LITHOLOGY=/path/to/lithology_models/best_model_lithology
+#   BERT_MODEL_PATH_USCS=/path/to/uscs_models/best_model_class_balancing
+#   BERT_MODEL_PATH_EN_MAIN=/path/to/en_models/best_model
```
christianabbet: Maybe better if not commented and have a toggle variable to chose mode like `BERT_LOCAL=true`

.env.template
```diff
@@ -9,4 +9,15 @@ AWS_DEFAULT_REGION=your_default_region
...
+# Option B — S3 key prefix within AWS_S3_BUCKET (files are downloaded to /tmp on first request):
+BERT_MODEL_S3_BUCKET=your_model_bucket_name
+BERT_MODEL_S3_KEY_LITHOLOGY=lithology_models/best_model_lithology
+BERT_MODEL_S3_KEY_USCS=uscs_models/best_model_class_balancing
+BERT_MODEL_S3_KEY_EN_MAIN=en_models/best_model
```
christianabbet: Variables are not up to date with README (same above). I think it should be

```
BERT_MODEL_S3_BUCKET=...
BERT_MODEL_S3_KEY_BACKBONE=backbone
BERT_MODEL_S3_KEY_LITHOLOGY_HEAD=lithology_head
BERT_MODEL_S3_KEY_EN_MAIN_HEAD=en_main_head
```

src/app/api/v1/endpoints/lithology_classification.py
```diff
@@ -0,0 +1,137 @@
...
+    "lithology": "BERT_MODEL_S3_KEY_LITHOLOGY_HEAD",
+    "en_main": "BERT_MODEL_S3_KEY_EN_MAIN_HEAD",
+}
+
+# Where downloaded models are cached on disk between warm Lambda invocations.
+_LOCAL_MODEL_CACHE_DIR = Path("/tmp/bert_models")
+
+# Module-level model cache keyed by (classification_system_name, resolved_model_path).
+# Avoids reloading the BERT model on every request.
+_model_cache: dict[tuple[str, str], BertModel] = {}
```
christianabbet: Instead of this we can use FastAPi asynccontextmanager. 

https://fastapi.tiangolo.com/advanced/events/#lifespan

src/classification/models/model.py
```diff
@@ -71,7 +100,58 @@ def _load_model(self) -> None:
...
+
+        # Load head weights via mmap — keeps them file-backed like the backbone
+        head_key = str(Path(self.model_path) / "model.safetensors")
+        if head_key not in _head_cache:
+            head_handle = safe_open(head_key, framework="pt", device="cpu")
+            _head_cache[head_key] = {
+                "handle": head_handle,
+                "tensors": {key: head_handle.get_tensor(key) for key in head_handle.keys()},  # noqa: SIM118
+            }
+        head_tensors = _head_cache[head_key]["tensors"]
```
christianabbet: It could be nice to have it in a function as `_load_head_model()`

src/classification/models/model.py
```diff
@@ -71,7 +100,58 @@ def _load_model(self) -> None:
...
+            logger.info("Backbone not cached — opening mmap and caching tensor references.")
+            handle = safe_open(str(self.backbone_path), framework="pt", device="cpu")
+            _backbone_cache[backbone_key] = {
+                "handle": handle,  # keep alive so mmap stays valid
+                "tensors": {key: handle.get_tensor(key) for key in handle.keys()},  # noqa: SIM118
+            }
+        else:
+            logger.info("Reusing cached mmap'd backbone tensors.")
+
+        cached_tensors = _backbone_cache[backbone_key]["tensors"]
```
christianabbet: It could be nice to have it in a function as `_load_backbone_model()`

src/classification/models/model.py
```diff
@@ -21,20 +18,45 @@
...
+                When provided, model_path is treated as the head directory and the two weight files are
+                merged before loading. When None, model_path must be a standard full HuggingFace model directory.
         """
         self.classification_system = classification_system
         self._setup_classification_system()
 
-        self.model_path = model_path
+        self.model_path = _resolve_path(model_path)
+        self.backbone_path = _resolve_path(backbone_path) if backbone_path else None
         self._load_model()
```
christianabbet: I would prefer somehting exhausive where we can see how the model is created like:

```
self.model = self._load_model()
```

Ideally, _load_model, _load_split_model, and _load_full_model should all return the model.

src/app/api/v1/endpoints/lithology_classification.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Args:
+        request (ClassifyLithologyRequest): The classification request containing a plain-text
+            material description and the target classification system.
+
+    Returns:
+        ClassifyLithologyResponse: Predicted class name for the input description.
+    """
+    bert_model = _get_bert_model(request.classification_system)
+    predicted_class = bert_model.predict_class(request.description)
```
christianabbet, line 57: you can add the decorator to all predict fuinctions. This way we don't have to use the "with torch.no_grad()" context in the code (line 298)". https://docs.pytorch.org/docs/stable/generated/torch.no_grad.html

```
@torch.no_grad()
```

src/classification/models/model.py
```diff
@@ -46,14 +68,21 @@ def _setup_classification_system(self) -> None:
...
             logger.info(f"Model and tokenizer loaded from local path: {self.model_path}")
         else:
             logger.info(f"Pretrained model and tokenizer loaded from remote source: {self.model_path}")
 
         self.tokenizer: BertTokenizerFast = AutoTokenizer.from_pretrained(self.model_path)
+
+        if self.backbone_path:
+            self._load_split_model()
+        else:
+            self._load_full_model()
```
christianabbet: Once loaded, the model should be put in evaluation mode (to avoid unwanted dropouts)

```
self.model.eval()
```

Dockerfile
```diff
@@ -10,16 +10,47 @@ ENV SETUPTOOLS_SCM_PRETEND_VERSION=$VERSION
...
+ARG AWS_SECRET_ACCESS_KEY
+ARG AWS_DEFAULT_REGION=eu-central-1
+ARG BERT_MODEL_S3_BUCKET
+ARG BERT_MODEL_S3_KEY_BACKBONE=backbone
+ARG BERT_MODEL_S3_KEY_LITHOLOGY_HEAD=lithology_head
+ARG BERT_MODEL_S3_KEY_EN_MAIN_HEAD=en_main_head
+
+ENV AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID
+ENV AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY
+ENV AWS_DEFAULT_REGION=$AWS_DEFAULT_REGION
```
christianabbet: I get a warning from Docker that It seems you can still access the variables after the run

```
docker history swissgeol-boreholes-dataextraction:local --no-trunc | grep -E "BERT_MODEL_S3_BUCKET"
```

It seems you can use something like that instead https://docs.docker.com/build/building/secrets/#using-build-secrets

src/app/api/v1/endpoints/lithology_classification.py
```diff
@@ -0,0 +1,137 @@
...
+# Avoids reloading the BERT model on every request.
+_model_cache: dict[tuple[str, str], BertModel] = {}
+
+
+def _resolve_model_path(classification_system_name: str) -> str:
+    """Resolve the local path to the fine-tuned BERT model.
+
+    Resolution order:
+    1. BERT_MODEL_PATH_<SYSTEM> env var — explicit local path (useful for local dev).
+    2. BERT_MODEL_S3_KEY_<SYSTEM> env var — S3 key prefix; files are downloaded to
```
christianabbet: Documentation var should be `<var>_HEAD`?

christianabbet (changes requested):
Here are the first comments, we can then check how it works afterwards. The running part if fine (except for the ENV variables, see below)

src/app/api/v1/endpoints/lithology_classification.py
```diff
@@ -0,0 +1,160 @@
...
+    env_var = (
+        _MODEL_S3_KEY_HEAD_ENV_VARS[classification_system_name]
+        if bert_split
+        else _MODEL_S3_KEY_ENV_VARS[classification_system_name]
+    )
+    s3_key = os.environ.get(env_var)
+    if s3_key:
+        local_dir = _LOCAL_MODEL_CACHE_DIR / classification_system_name
+        logger.info(f"Downloading {classification_system_name} model from S3 key: {s3_key}")
+        download_model_from_s3(s3_key, local_dir)
```
stijnvermeeren-swisstopo: the README claims that
> Building the image requires AWS credentials to download the BERT classification models at build time. Credentials are passed as [BuildKit secrets](https://docs.docker.com/build/building/secrets/) so they are never stored in any image layer:

...but here, it looks like models are downloaded from S3 at run-time. Can you clarify how this is supposed to work?

More generally, I think that we could make our best trained models public, either in a public S3 bucket, or committed to the repository itself (like we do with the assets classifiers). That would probably simplify the whole credential management?

docs/API_and_Docker.md
```diff
@@ -53,9 +53,62 @@ Additional endpoints and their functionalities can be found in the project's sou
 To stop the FastAPI server, press `Ctrl + C` in the terminal where the server is running. Please refer to the [FastAPI documentation](https://fastapi.tiangolo.com) for more information on how to work with FastAPI and build APIs using this framework.
 
 
+## Classification Model Configuration
+
+The `/api/V1/classify_lithology` endpoint requires a fine-tuned BERT model. Two independent toggles control how the model is loaded:
```
stijnvermeeren-swisstopo: Can we also include an option to not load any BERT models at all, in case the user is only interested in doing data extraction without classification? The `classify_lithology` endpoint would not work, but they could still use the extraction endpoints without "wasting" memory on loading the fine-tuned BERT models.

stijnvermeeren-swisstopo (changes requested):

src/app/main.py
```diff
@@ -20,8 +22,23 @@
 
 load_dotenv()
 
+
+@asynccontextmanager
+async def lifespan(app: FastAPI):
+    """Lifespan function to load the BERT models when the application starts."""
+    bert_enabled = os.environ.get("BERT_ENABLED", "true").lower() != "false"
```
stijnvermeeren-swisstopo: `.env.template` says "Default: False." but here the default seems to be `True`?

models/lithology_head/tokenizer.json
stijnvermeeren-swisstopo, line 1: Is it really necessary to duplicate this very large JSON file (it looks identical between here and `en_main_head/tokenizer.json`)?

stijnvermeeren-swisstopo (dismissed):
When I try to run uvicorn src.app.main:app --reload --host 0.0.0.0 --port 8002 (i.e. running the API without Docker) with BERT_ENABLED=True then I get the following error:

```
  File "/home/stijn3/swissgeol-boreholes-dataextraction/src/app/api/v1/endpoints/lithology_classification.py", line 35, in load_models
    models[system_name] = BertModel(model_path, classification_system, backbone_path=_BACKBONE_PATH)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stijn3/swissgeol-boreholes-dataextraction/src/classification/models/model.py", line 61, in __init__
    self.model = self._load_model()
                 ^^^^^^^^^^^^^^^^^^
  File "/home/stijn3/swissgeol-boreholes-dataextraction/src/classification/models/model.py", line 84, in _load_model
    model = self._load_split_model() if self.backbone_path else self._load_full_model()
            ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stijn3/swissgeol-boreholes-dataextraction/src/classification/models/model.py", line 169, in _load_split_model
    cached_tensors = self._load_backbone_model()
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stijn3/swissgeol-boreholes-dataextraction/src/classification/models/model.py", line 138, in _load_backbone_model
    return self._load_tensors_cached(str(self.backbone_path), _backbone_cache)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stijn3/swissgeol-boreholes-dataextraction/src/classification/models/model.py", line 129, in _load_tensors_cached
    handle = safe_open(file_path, framework="pt", device="cpu")
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
safetensors_rust.SafetensorError: Error while deserializing header: header too large ```

File: models/lithology_head/tokenizer.json
Comment (AgathaSchmidt) [reply models/lithology_head/tokenizer.json:1]: I enabled sharing 
And I also manage the file with LFS now so it takes up less space

**Comment by AgathaSchmidt** [comment]
> When I try to run uvicorn src.app.main:app --reload --host 0.0.0.0 --port 8002 (i.e. running the API without Docker) with BERT_ENABLED=True then I get the following error:
> 
> ```
> safetensors_rust.SafetensorError: Error while deserializing header: header too large ```
> ```

Probably, your backbone.safetensors is a Git LFS pointer, not the real file. Try to run:
```
git lfs install
git lfs pull
````

And then run the API again

The LFS handling was documented in the README.md and now I updated to API_and_Docker.md

stijnvermeeren-swisstopo:
@AgathaSchmidt thanks for the answers.

Unfortunately, I won't have the time to do a proper review round before my holidays.

I find it a little strange that you'd have to install a git extension before you can fully load everything you need for the classification, but if you all think that this is the best solution, then I will not object against it.

You can remove me as a reviewer, and merge as soon as @christianabbet approves.

christianabbet (approved):
LGTM. THe proposed fix works locally as expected!

AgathaSchmidt:
We are merging this for now, but would like to open a discussion on a better way on how to provide models to the user, instead of pushing them to the repository (e.g. public AWS bucket, uploading model to Hunggingface)

PR #452, opened 2026-04-14
Refactoring of evaluation pipeline file by file

christianabbet:
Close #433 

# Description

Refactors the extraction evaluation pipeline from batch-oriented to file-level processing, improving readability and maintainability.

# Changes

Introduces a new `Evaluator` class (`src/extraction/evaluation/evaluator.py`) as the main entry point for evaluation
* `Evaluator.evaluate(file)`: evaluates a single file, compute metrics, and mutates the `is_correct` flags on its entries (layers, depth intervals, material descriptions, groundwater, metadata).
* `Evaluator.aggregate(files)`: Aggregates results for all files.

# Output MLFlow

The logged output files were homogenized across geology and metadata. At the end of the script, we log each metric: F1, precision, and recall. Before that, only F1 was logged, and sometimes recall and precision, depending on the metrics. In addition, the way Dataframes are compute are unified across geology and metadata with `to_df()`

### DIFF
--- a/docs/predictions-json.md
+++ b/docs/predictions-json.md
@@ -32,7 +32,7 @@ predictions.json
     │       ├── page                      # Source page number
     │       └── rect                      # Bounding box of the measurement in the PDF
     │
-    └── metrics                           # Evaluation scores with tp, fp, fn,precision, recall, and f1
+    └── metrics                           # Evaluation scores with tp, fp, fn, precision, recall, and f1
         ├── language                      # Ground truth language for metric computation
         ├── layer_metrics                 # Layer detection
         ├── depth_interval_metrics        # Depth layer detection
@@ -119,6 +119,7 @@ All page numbers are counted starting at 1. All bounding boxes are measured with
               }
             }
           }
+          # ... (more layers)
         ],
         "bounding_boxes": [
           {
@@ -126,7 +127,13 @@ All page numbers are counted starting at 1. All bounding boxes are measured with
             "depth_column_entries": [
               [200.1201171875, 321.8956298828125, 208.59901428222656, 328.6802062988281],
               [200.63790893554688, 331.3035888671875, 207.83108520507812, 338.30450439453125],
-              [201.62551879882812, 374.30560302734375, 210.0361328125, 380.828857421875]
+              [201.62551879882812, 374.30560302734375, 210.0361328125, 380.828857421875],
+              [199.86251831054688, 434.51556396484375, 210.10894775390625, 441.4538879394531],
+              [198.11251831054688, 557.5472412109375, 210.35877990722656, 563.9244995117188],
+              [198.28451538085938, 582.0216674804688, 209.76953125, 588.7603759765625],
+              [198.7814178466797, 616.177001953125, 209.50042724609375, 622.502197265625],
+              [198.6378173828125, 663.2830810546875, 210.75906372070312, 669.5428466796875],
+              [198.26901245117188, 695.974609375, 209.12693786621094, 702.2628173828125]
             ],
             "material_description_rect": [256.777099609375, 345.9997253417969, 392.46051025390625, 728.2700805664062],
             "page": 1
@@ -148,6 +155,7 @@ All page numbers are counted starting at 1. All bounding boxes are measured with
           }
         ]
       }
+      # ... (more boreholes)
     ],
     "metrics": {
       "layer_metrics": {
--- a/docs/predictions-json.md
+++ b/docs/predictions-json.md
@@ -17,7 +17,7 @@ predictions.json
     │   ├── metadata
     │   │   ├── elevation                 # Borehole surface elevation, if found
     │   │   └── coordinates               # Borehole coordinates, if found
-    │   ├── layers[]                      # Detected layers of the borehole profil
+    │   ├── layers[]                      # Detected layers of the borehole profile
     │   │   ├── material_description      # Material text description
     │   │   └── depth_interval            # Measured depth of the layer's upper and lower limits
     │   ├── bounding_boxes[]              # List of bounding boxes, one for each part of a borehole profile, that can be used for visualizations
--- a/docs/predictions-json.md
+++ b/docs/predictions-json.md
@@ -54,7 +54,6 @@ All page numbers are counted starting at 1. All bounding boxes are measured with
 ## Example output
 ```yaml
 {
-# TODO Return metadata as before with page dimmension and language /GT vs predicted
   "B366.pdf": {  # file name
     "language": "de",
     "page_dimensions": [
--- a/docs/predictions-json.md
+++ b/docs/predictions-json.md
@@ -5,7 +5,7 @@ The `predictions.json` file contains the results of a data extraction process in
 
 ## Schema Overview
 
-```yaml
+```text
 predictions.json
 └── "<filename>.pdf"                      # One entry per processed PDF file
     ├── file_metadata
@@ -20,7 +20,7 @@ predictions.json
     │   ├── layers[]                      # Detected layers of the borehole profil
     │   │   ├── material_description      # Material text description
     │   │   └── depth_interval            # Measured depth of the layer's upper and lower limits
-    │   ├── bounding_boxes[]
+    │   ├── bounding_boxes[]              # List of bounding boxes, one for each (part of a) borehole profile, that can be used for visualizations
     │   │   ├── page                      # Page number for this segment
     │   │   ├── sidebar_rect              # Area of the depth sidebar, if found
     │   │   ├── depth_column_entries[]    # Bounding boxes of depth entries
--- a/docs/predictions-json.md
+++ b/docs/predictions-json.md
@@ -30,8 +30,9 @@ All bounding boxes are measured with PDF points as the unit, and with the top-le
 ## Example output
 ```yaml
 {
+# TODO Return metadata as before with page dimmension and language /GT vs predicted
   "B366.pdf": {  # file name
-    "language": "de", 
+    "language": "de",
     "page_dimensions": [
       {
         "width": 591.956787109375,
--- a/docs/predictions-json.md
+++ b/docs/predictions-json.md
@@ -1,31 +1,56 @@
-# `predictions.json` output structure
+# `predictions.json` Output Structure
+
 The `predictions.json` file contains the results of a data extraction process in a machine-readable format. By default, the file is written to `data/output/predictions.json`.
 
-Each key in the JSON object is the name of a PDF file. The extracted data is listed as an object with the following keys:
-- `language`: language that was detected for the document.
-- `page_dimensions`: dimensions of each page in the PDF, measured in PDF points
-- `boreholes`: list of all boreholes identified on the documents. Each borehole object has the following keys.
-  - `borehole_index`: index to differenciate the extracted boreholes, starting at 0.
-  - `metadata`
-    - `elevation`: the detected elevation (if any) and the location in the PDF where they were extraction from.
-    - `coordinates`: the detected coordinates (if any) and the location in the PDF where they were extraction from.
-  - `layers`: a list of objects, where each object represents a layer of the borehole profile, using the following keys:
-    - `material_description`: the text of the material description, both as a single value as well as line-by-line, and the location in the PDF where the text resp. the lines where extracted from.
-    - `depth_interval`: the measured depth of the upper and lower limits of the layer, and the location in the PDF where they were extracted from.
-  - `bounding_boxes`: a list of objects, one for each (part of a) borehole profile in the PDF, that list some bounding boxes that can be used for visualizations. Each object has the following keys:
-    - `sidebar_rect`: the area of the page the contains a "sidebar" (if any), which contains depths or other data displayed to the side of material descriptions.
-    - `depth_column_entries`: list of locations of the entries in the depth column (if any).
-    - `material_description_rect`: the area of the page that contains all material descriptions.
-    - `page`: the number of the page of the PDF.
-  - `groundwater`: a list of objects, one for each groundwater measurement that was extracted from the PDF. Each object has the following keys.
-    - `date`: extracted date for the groundwater measurement (if any) as a string in YYYY-MM-DD format.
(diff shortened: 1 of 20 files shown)


src/extraction/evaluation/evaluator.py
```diff
@@ -0,0 +1,259 @@
...
+        layer_metrics, depth_interval_metrics, material_description_metrics = Evaluator.evaluate_layers(
+            file_predictions
+        )
+        gw_metrics = Evaluator.evaluate_gw(file_predictions)
+        metadata_metrics = Evaluator.evaluate_metadata(file_predictions)
+
+        return layer_metrics, depth_interval_metrics, material_description_metrics, gw_metrics, metadata_metrics
+
+    @staticmethod
+    def evaluate_layers(file_predictions: FilePredictionsWithGroundTruth) -> tuple[Metrics, Metrics, Metrics]:
```
AgathaSchmidt: This method is only used in this file. 
Maybe add `_` prefix to signal that it is internal?

src/extraction/evaluation/evaluator.py
```diff
@@ -0,0 +1,259 @@
...
+                        borehole.predictions.layers_in_borehole if borehole.predictions else None,
+                        borehole.ground_truth.get("layers", []),
+                    )
+                    for borehole in file_predictions.boreholes
+                ],
+            )
+        )
+
+    @staticmethod
+    def evaluate_metadata(
```
AgathaSchmidt: This method is only used in this file. 
Maybe add `_` prefix to signal that it is internal?

src/extraction/evaluation/evaluator.py
```diff
@@ -0,0 +1,259 @@
...
+                        predictions.predictions.metadata if predictions.predictions else None,
+                        predictions.ground_truth.get("metadata", {}),
+                    )
+                    for predictions in file_predictions.boreholes
+                ],
+            )
+        )
+
+    @staticmethod
+    def evaluate_gw(file_predictions: FilePredictionsWithGroundTruth) -> GroundwaterMetrics:
```
AgathaSchmidt: This method is only used in this file. 
Maybe add `_` prefix to signal that it is internal?

src/extraction/evaluation/evaluator.py
```diff
@@ -0,0 +1,259 @@
...
+                - OverallBoreholeMetadataMetrics: aggregated metadata metrics across all files.
+        """
+        fp_languages = {fp.filename: fp.language for fp in overall_predictions.predictions_list}
+        languages = set(fp_languages.values())
+
+        overall_groundwater_metrics = OverallGroundwaterMetrics()
+        overall_metadata_metrics = OverallBoreholeMetadataMetrics()
+
+        overall_geology_metrics = OverallMetricsCatalog(languages=languages)
+        overall_layer_metrics = OverallMetrics()
```
AgathaSchmidt: `OverallMetricsCatalog.__init__` already initialises layer_metrics, depth_interval_metrics, and material_description_metrics, so you can write them directly through `overall_geology_metrics`

```python 
# before
overall_layer_metrics = OverallMetrics()
...
overall_layer_metrics.metrics[file_predictions.filename] = layer_metrics
...
overall_geology_metrics.layer_metrics = overall_layer_metrics

# after
overall_geology_metrics.layer_metrics.metrics[file_predictions.filename] = layer_metrics

```

src/extraction/evaluation/benchmark/score.py
```diff
@@ -86,48 +86,66 @@ def evaluate_all_predictions(
...
+    logger.info("Macro avg:")
+    logger.info(
+        "layer f1: %.1f%%, depth interval f1: %.1f%%, material description f1: %.1f%%",
+        geology_metrics.layer_metrics.macro_f1() * 100,
+        geology_metrics.depth_interval_metrics.macro_f1() * 100,
+        geology_metrics.material_description_metrics.macro_f1() * 100,
+    )
+
     #############################
     # Evaluate the borehole extraction
```
AgathaSchmidt: `# Log geology metrics `

(evaluation is not happening here anymore)

src/extraction/evaluation/benchmark/score.py
```diff
@@ -86,48 +86,66 @@ def evaluate_all_predictions(
...
+
+    geology_metrics_dict = geology_metrics.metrics_dict()
 
     # Format the metrics dictionary to limit to three digits
-    formatted_metrics = {k: f"{v:.3f}" for k, v in metrics_dict.items()}
+    formatted_metrics = {k: f"{v:.3f}" for k, v in geology_metrics_dict.items()}
     logger.info("Performance metrics: %s", formatted_metrics)
 
     #############################
     # Evaluate the borehole extraction metadata
```
AgathaSchmidt: `# Log metadata metrics`

AgathaSchmidt: maybe also just add a comment instead of the large banner since nothing major is happening anymore

AgathaSchmidt (commented):
Looks good, thank you 
Just some minor comments

christianabbet:
@stijnvermeeren-swisstopo I investigated the uv change with the 1000 lines. Two things to note:

* The changes stem from the 7-day rolling update we configured. When running `uv sync`, packages now update to their latest versions (within that 7-day delay). The most critical packages are pinned, so core functionality stays consistent. I think it is a nice thing to keep in terms of security.
* There's a minor issue with how versioning works locally. Running `uv sync` also installs the package itself, which conflicts with the GitHub versions. Two ways to fix this: either run `uv sync --all-extras --no-install-project` to skip the local install, or bump the version locally as a workaround. I've gone with the latter for now (see pyproject.toml).

tests/test_groundwater.py
stijnvermeeren-swisstopo: We're not really testing any functionality of the `OverallGroundwaterMetrics` class here, but rather we're only making assertions about the results of the individual `GroundwaterEvaluator.evaluate` calls. So for improved simplicity and clarity, I would assign each results of `GroundwaterEvaluator.evaluate` to its own variable, and then make the assertions directly on those variables.

src/extraction/runner.py
```diff
@@ -149,6 +151,9 @@ def run_predictions(self, predictions_path_tmp: Path) -> PipelineRunResult[Overa
                 predictions.add_file_predictions(result.predictions)
 
                 if self.on_file_done is not None:
+                    # Evaluate results
+                    result.predictions = evaluate_prediction(result.predictions, ground_truth)
```
stijnvermeeren-swisstopo: Would it make sense to always evaluate predictions for individual inputs immediately, even when `self.on_file_done is None`?

And if we indeed do that, wouldn't that remove the need to again call `Evaluator.match_with_ground_truth` and  `Evaluator.evaluate` for all files from `evaluate_all_predictions` after all files have been processed?

stijnvermeeren-swisstopo (changes requested):

src/extraction/runner.py
christianabbet: I considered that approach, but it requires a few more changes. To avoid recomputing metrics, we'd need to persist them during inference (to support pipeline resume). This means introducing a new object that encapsulates both the predictions and their metrics. We'd also need to track the ground truth language in order to correctly aggregate results by language downstream. The pipeline would look something like:

* Load previous predictions with metrics and ground truth language (if resuming)
* Run inference on files and save results as a new object (e.g. FilePredictionsWithMetrics)
* Dump predictions with metrics and language
* Extract metrics from precomputed predictions and aggregate metrics

Does that align with what you had in mind?

stijnvermeeren-swisstopo: @christianabbet Thanks for highlighting the implications of my proposed change. While I did not fully consider that when I initially wrote my comment, I still believe that this would be a cleaner solution compared to the current implementation. So I'd be happy if you spend a bit of additional time to implement it like that, including the additional functionality required for the pipeline resume feature.

docs/predictions-json.md
```diff
@@ -1,47 +1,73 @@
...
-  - `groundwater`: a list of objects, one for each groundwater measurement that was extracted from the PDF. Each object has the following keys.
-    - `date`: extracted date for the groundwater measurement (if any) as a string in YYYY-MM-DD format.
-    - `depth`: the measured depth (in m) of the groundwater measurement.
-    - `elevation`: the elevation (in m above sea level) of the groundwater measurement.
-    - `page` and `rect`: the location in the PDF where the groundwater measurement was extracted from.
 
-All page numbers are counted starting at 1.
+## Schema Overview
+
+```bash
```
stijnvermeeren-swisstopo: I like this improved representation, but it's definitely not bash syntax :).

docs/predictions-json.md
```diff
@@ -1,47 +1,73 @@
...
+    │   └── page_dimensions[]             # Dimensions of each page, in PDF points
+    │
+    ├── boreholes[]                       # All boreholes identified in the document
+    │   ├── borehole_index                # Zero-based index to distinguish boreholes
+    │   ├── metadata
+    │   │   ├── elevation                 # Borehole surface elevation, if found
+    │   │   └── coordinates               # Borehole coordinates, if found
+    │   ├── layers[]                      # Ordered geological layers of the borehole profile
+    │   │   ├── material_description      # Soil/rock description for this layer
+    │   │   └── depth_interval            # Vertical extent of the layer
```
stijnvermeeren-swisstopo: AI (I assume) did a pretty good job overall with the comments, but this one ("Vertical extent of the layer") is not very clear, I prefer the old one.

docs/predictions-json.md
```diff
@@ -1,47 +1,73 @@
...
+    │   ├── bounding_boxes[]
```
stijnvermeeren-swisstopo: Keep a comment that emphasizes that these bounding boxes can be useful to produce visualizations of the extracted data.

stijnvermeeren-swisstopo (changes requested):
While I did not read through every individual line of code, in general, everything seems to make sense. The computed metrics are identical to the develop branch, except for the naming that has been harmonized as described, which I also approve of.

What remains is just a few small details regarding the markdown docs.

src/extraction/evaluation/evaluation_dataclasses.py
```diff
@@ -1,89 +1,57 @@
...
+        return BoreholeMetadataMetrics(
+            elevation_metrics=Metrics.from_json(json["elevation"]),
+            coordinates_metrics=Metrics.from_json(json["coordinates"]),
+            name_metrics=Metrics.from_json(json["name"]),
         )
 
 
 @dataclass
-class OverallBoreholeMetadataMetrics(metaclass=abc.ABCMeta):
+class OverallBoreholeMetadataMetrics:
```
AgathaSchmidt: I think this is dead code 
`add_metadata_metrics`is never used

src/extraction/evaluation/evaluation_dataclasses.py
```diff
@@ -1,89 +1,57 @@
...
-            dict[str, float]: The object as a dictionary.
+            dict[str, dict]: The object as a dictionary.
         """
         return {
-            **self.elevation_metrics.to_json("elevation"),
-            **self.coordinates_metrics.to_json("coordinate"),
-            **self.name_metrics.to_json("borehole_name"),
+            "elevation": self.elevation_metrics.to_json(),
+            "coordinates": self.coordinates_metrics.to_json(),
+            "name": self.name_metrics.to_json(),
```
AgathaSchmidt, line 25: is there a reason you renamed `borehole_name` to `name`?  
I was a bit confused when checking mlflow columns

docs/predictions-json.md
christianabbet: I tried with `text` instead but then you loose the color. I changed now to `yaml` (enven though it is still not yaml). If you prefer to put `text` instead i can do it no problem.

I also corrected the second bloc that was labelled as `yaml` instead of `jsonc`

docs/predictions-json.md
stijnvermeeren-swisstopo: @christianabbet this is not addressed yet, as far as I can tell

docs/predictions-json.md
stijnvermeeren-swisstopo: @christianabbet I prefer declaring it correctly as general plain text, and forgoing the automatic coloring of the comments. I don't think the colors are essential; I think it's more important to avoid incorrect language declarations.

src/extraction/evaluation/evaluator.py
```diff
@@ -0,0 +1,225 @@
...
+            file_predictions (FilePredictionsWithGroundTruth): Per-file borehole predictions
+                paired with their ground truth data.
+
+        Returns:
+            tuple[Metrics, Metrics, Metrics, GroundwaterMetrics, BoreholeMetadataMetrics]:
+                A 5-tuple of:
+                - layer_metrics (Metrics): Metrics for layer detection.
+                - depth_interval_metrics (Metrics): Metrics for depth intervals.
+                - material_description_metrics (Metrics): Metrics for material descriptions.
+                - groundwater_metrics (GroundwaterMetrics): Metrics for Groundwater detection.
```
AgathaSchmidt: gw_metrics 
or return groundwater_metrics

src/extraction/evaluation/evaluation_dataclasses.py
stijnvermeeren-swisstopo: Good point. I'm now wondering the same about `OverallGroundwaterMetrics`. It is instantiated as  `overall_groundwater_metrics = OverallGroundwaterMetrics()` in `evaluator.py`, but on first sight it looks like this `overall_groundwater_metrics` is never really used and could be removed from the code.

stijnvermeeren-swisstopo (changes requested):
See comment on Agatha's comment regarding `OverallBoreholeMetadataMetrics` and `OverallGroundwaterMetrics`.

src/extraction/evaluation/evaluation_dataclasses.py
christianabbet: It was to be consistent with the variable names. Is there a reason to keep it as `borehole_name` ?

stijnvermeeren-swisstopo (approved):
LGTM!

AgathaSchmidt (approved):

PR #456, opened 2026-04-16
Feat/issue 442/fix mismatching of material descriptions and depth entries for specific borehole type sondierprofil

### LINKED ISSUE #442: Fix mismatching of material descriptions and depth entries for specific borehole type (Sondierprofil)
When analyzing the material description quality of the Thurgau dataset in #418 we found that for some specific boreholes of the type "Sondierprofil" a mismatch between material descriptions and depth intervals happens. 
E.g. 3184_KB_41_37_Kreuzlingen_Weinberg and 3266_RKB_41_131_Kreuzlingen_Weinberg 
In those examples, two line breaks are not correctly detected. One material description was split into two and later two material descriptions which do not belong together where combines to one. 

Both of those boreholes have a **AabvobeB** sidebar so I would propose to try to find the source of the problem in the matching algorithm in the AaboveB sidebar type.

Image
Image

AgathaSchmidt:
Closes #442 
# Problem

Sondierprofil boreholes were classified as a_above_b sidebars. The matching algorithm for this sidebar type was leadinf to incorrect matching. When classifying it as Protocl sidebar the matching would become correct. 
So we were working towards a solution which recognized this boreholes as protocol sidebar without distroying performance on other files. 

# Changes

The sidebar extraction pipeline now branches on whether a strip log is detected:

- Strip log present — original blocking logic: spulprobe → a_to_b → a_above_b (with entry exclusion); protocol only runs if neither a_above_b nor a_to_b has a valid description match.
- No strip log — a_above_b and protocol both run unconditionally with the full entry pool; a_to_b still blocks protocol if it has a valid match. score_match decides the winner between a_above_b and protocol.

Two mechanisms ensure structured sidebar types are preferred over protocol when they compete for the same description column:

- Tiebreaker: when a_above_b and another type score equally, the non-a_above_b type wins (a truly correct a_above_b should score clearly higher).
- Validation fallback: if a non-a_above_b type wins the tiebreaker but then fails borehole validation (e.g. protocol requires every entry to have a matched description), the pipeline retries with the a_above_b sidebar as a fallback.

The extractor now starts from keyword headers (Tiefe, bis, Depth, etc.) rather than clustering all numeric entries and checking for a header afterward. Only entries below a detected header and x-aligned with it (within an expanded window) are considered. This eliminates false positives from unrelated numeric columns.

# side effects 

**geoquat/val**
<google-sheets-html-origin>
file name | comment | effect 
-- | -- | --
1088 | wrong sidebar (atoB not detected) |  negative
 B147 | newly detected sidebar entries --> more md  (page2) |  positive
A8384 |  different md matching (both incorrect) | neutral
V69  |  some sidebar entries newly detected --> some md lost |  ?

**zurich**

<google-sheets-html-origin>

file name | comment | effect 
-- | -- | --
691251009 | slightly better md matching | positive
267124198 | atob not detected but all as protocolsidebar --> more layers and more md | ? 
268124112 | FP added (?)
268124125 | md lost (page3) | negative

<google-sheets-html-origin>
develop | 0.842 | 0.893 | 0.88 | 0.373 | 0.814 | 0.694 | 0.73
442 | 0.838 | 0.894 | 0.879 | 0.373 | 0.812 | 0.694 | 0.728

<google-sheets-html-origin>
develop | 0.479 | 0.6296 | 0.628 | 0.05 | 0.615 | 0.475 | 0.604
442 | 0.477 | 0.632 | 0.624 | 0.05 | 0.615 | 0.475 | 0.604

### DIFF
--- a/src/extraction/features/extract.py
+++ b/src/extraction/features/extract.py
@@ -912,9 +912,6 @@ def get_pairs_based_on_line_affinity(
     weights = matching_params["affinity_params"]["no_sidebar"]["weights"]
     threshold = -0.3
 
-    for line, aff in zip(description_lines, affinities, strict=False):
-        print(aff.weighted_affinity(**weights), aff.vertical_spacing_affinity, line.text)
-
     for line_idx, affinity in enumerate(affinities):
         # note: the affinity of the first line is always 0.0
         if affinity.weighted_affinity(**weights) <= threshold:
--- a/src/extraction/features/extract.py
+++ b/src/extraction/features/extract.py
@@ -122,11 +122,6 @@ class MaterialDescriptionRectWithSidebarExtractor:
             if borehole is None and id(pair) in fallback_dict:
                 fallback = fallback_dict[id(pair)]
                 if not any(self._pairs_intersect(fallback, accepted) for accepted in pairs_from_valid_boreholes):
-                    logger.debug(
-                        "Page %s: primary pair (kind=%s) failed validation, retrying with a_above_b fallback",
-                        self.page_number,
-                        pair.sidebar.kind if pair.sidebar else "none",
-                    )
                     borehole = self._create_borehole_from_pair(fallback)
                     if borehole is not None:
                         accepted_pair = fallback
@@ -342,27 +337,14 @@ class MaterialDescriptionRectWithSidebarExtractor:
         words = sorted([word for line in self.lines for word in line.words], key=lambda word: word.rect.y0)
 
         has_strip_log = bool(self.strip_logs)
-        logger.debug("Page %s: strip_log detected=%s", self.page_number, has_strip_log)
 
         spulprobe_sidebars = SpulprobeSidebarExtractor.find_in_lines(self.lines)
-        logger.debug(
-            "Page %s: found %d spulprobe sidebar(s): %s",
-            self.page_number,
-            len(spulprobe_sidebars),
-            [f"entries={[str(e.rect) for e in s.entries]}" for s in spulprobe_sidebars],
-        )
         sidebars_noise: list[SidebarNoise] = [
             SidebarNoise(sidebar=sidebar, noise_count=noise_count(sidebar, line_rtree))
             for sidebar in spulprobe_sidebars
         ]
 
         a_to_b_sidebars = AToBSidebarExtractor.find_in_words(words)
-        logger.debug(
-            "Page %s: found %d a_to_b sidebar(s): %s",
-            self.page_number,
-            len(a_to_b_sidebars),
-            [f"entries={[str(e.start.rect) + '->' + str(e.end.rect) for e in s.entries]}" for s in a_to_b_sidebars],
-        )
         a_to_b_sidebars_noise = [
             SidebarNoise(sidebar=sidebar, noise_count=noise_count(sidebar, line_rtree)) for sidebar in a_to_b_sidebars
         ]
@@ -384,15 +366,6 @@ class MaterialDescriptionRectWithSidebarExtractor:
                 list(used_entry_rects),
                 sidebar_params=self.matching_params["depth_column_params"],
             )
-            logger.debug(
-                "Page %s: found %d a_above_b sidebar(s): %s",
-                self.page_number,
-                len(a_above_b_sidebars_noise),
-                [
-                    f"entries={[str(e.rect) for e in sn.sidebar.entries]}, noise={sn.noise_count}"
-                    for sn in a_above_b_sidebars_noise
-                ],
-            )
             sidebars_noise.extend(a_above_b_sidebars_noise)
 
             for sidebar_noise in a_above_b_sidebars_noise:
@@ -400,7 +373,6 @@ class MaterialDescriptionRectWithSidebarExtractor:
                     used_entry_rects.add(entry.rect)
 
             block_protocol = self._should_block_protocol(a_above_b_sidebars_noise + a_to_b_sidebars_noise)
-            logger.debug("Page %s: block_protocol=%s (strip log present)", self.page_number, block_protocol)
 
             if not block_protocol:
                 protocol_sidebars_noise = ProtocolSidebarExtractor.find_in_words(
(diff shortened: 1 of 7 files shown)


src/extraction/features/extract.py
```diff
@@ -264,7 +265,13 @@ def _get_interval_block_pairs(self, pair: MaterialDescriptionRectWithSidebar) ->
             left_line_length_threshold=self.matching_params["left_line_length_threshold"],
         )
         return (
-            match_lines_to_interval(pair.sidebar, description_lines, line_affinities, diagonals)
+            match_lines_to_interval(
+                pair.sidebar,
+                description_lines,
+                line_affinities,
+                diagonals,
+                group_by_spacing=not bool(self.strip_logs),
```
christianabbet: I get that this leads to better results, but that sound odd. It's taken out of nowhere. Maybe discuss with @stijnvermeeren-swisstopo if okay to add condition.

src/extraction/features/extract.py
```diff
@@ -882,12 +932,33 @@ def match_lines_to_interval(
...
+            _, mapping = IntervalToLinesDP(depth_interval_zones, representatives, affinity_scores).solve(
+                sidebar.dp_scoring_fn
+            )
+            mapping = [(zone, [line for rep in reps for line in group_by_rep_id[id(rep)]]) for zone, reps in mapping]
+            result = sidebar.post_processing(mapping)
+            if sum(1 for ibp in result if ibp.block.lines) == n_intervals:
+                return result
+            # Grouped DP left some interval(s) empty — fall through to the original DP.
+
+    # Original (ungrouped) DP
```
christianabbet: I would argue to do the group by default (whitout group_by_spacing) with a function like: `result = hard_group_match()`, then if that function fails, use fallback `result = soft_match()` with previous approach. 

Could you tell what was the issue with group matching and files with detected strip logs ?

christianabbet (changes requested):

src/extraction/features/extract.py
AgathaSchmidt: Okay, I understand. 
I tried to explain the reasoning behind this step in the PR: material description lines are scored line by line and this behavior is good and actually for the cases covered in this issue it makes sense that they are mismatched with the wrong depth interval. Therefore: instead of remocing this behavior, apply it on the whole material description entry instead of every single line. 
Maybe we should discuss it in general in the next StandUp

src/extraction/features/extract.py
AgathaSchmidt: Yes we can do that 
It has more effect that my previous solution did: adding some more positive effects as well as negatives ones. 

The issues which come with it are ones I had before. (See PR)
I was trying to avoid any negative influence on the files which were matched correctly before for at all costts but I agree that maybe in this case it might be worth it.

stijnvermeeren-swisstopo (commented):
@AgathaSchmidt could you elaborate on the statement "So the logic as it is actually does make sense and I did not want to break it."

Why exactly do you think that the logic makes sense for the given borehole profile layout?

The increase in accuracy on our benchmarking datasets looks great. However, I'm not yet 100% convinced that we really need to split between a "hard" and "soft" match method. I have some hope that we might be able to achieve the same accuracy by fine-tuning the dynamic programming approach, and if that is the case, that would be preferrable over splitting the logic into two different approaches.

stijnvermeeren-swisstopo:
I can confirm that when I treat the sidebar as a ProtocolSidebar instead of as a AAboveBSidebar in `match_lines_to_interval`, then the extraction for the two Thurgau examples mentioned in the ticket become correct. This indeed would be a conceptually sound solution, since indeed these layouts have more in common with the ProtocolSidebar than with the AAboveBSidebar.

The big question is: how do we decide when a ProtocolSidebar would be a better match than an AAboveBSidebar? I tried comparing the scores that are returned by the `solve` method of the dynamic programming implementation, but unfortunately that does not seem to work well.

stijnvermeeren-swisstopo:
Just to explain my preferences a little better:

**Why am I skeptical about the proposed "hard" vs "soft" match implementation**

While it works well for the benchmarking data, I don't like that a very small change in the input (e.g. a single depth value or material description that is missed) can flip the condition that controls whether we follow the "hard match" or the "soft match". Consequently, a small change in the input document could lead to drastic changes in the extracted data, which is not great for the explainability/understandability of the pipeline. Methods that are more "continuous" in their behaviour (small input change => small output change) are also more likely to generalize better to new dataset.

**Why not increase the weight of the vertical spacing affinity in the DP approach then?**

The fact that that the "hard match" approach works well, hints that it might be beneficial to increase the weight of the vertical spacing between description lines in the dynamic programming matching method. While it is worthwhile to investigate this, and it might even potentially fix the extraction results for the two examples from Thurgau, I still believe that it would not be the most robust fix.

Indeed, imagine borehole profiles such as the ones from Thurgau, but with more and more layers. With the AAboveBSidebar DP matching, the misalignments between the depth intervals and their corresponding descriptions will add up proportionally to the number of layers. No matter how high we put the weight of the vertical spacing affinity, at some point the DP will still prefer "violating" the vertical spacing affinity in two places, because the summed distances from the misalignment between depths and descriptions will be more significant compared to twice taking the negative vertical spacing affinity.

This is why I would prefer treating these borehole profiles as having a Protocol layout, because with the corresponding matching logic for the ProtocolSidebar, we will naturally get the correct interpretation, regardless of how many layers the borehole has.

stijnvermeeren-swisstopo:
@AgathaSchmidt can you explain in detail why the behaviour for 267124198 changed in the current implementation, i.e. why the AToBSidebar is no longer detected/selected?

AgathaSchmidt:
@stijnvermeeren-swisstopo  
Before, when the a_to_b sidebar was detected, those detected values were added to the exclusion list which made the detected values invisible for other sidebars --> this way, no solid a_above_b sidebar could be build
Now, we don't add the used valued to `used_entry_rects`. I did this to enable multiple sidebar types to be extracted (e.g. for Sodierprofil:  to still check if a protocol sidebar may be better than the a_above_b sidebar --> we need all values available ) 

And because the a_above_b sidebar detects all 8 entries it gets a higher score and therefore wins: 

  [a_to_b] idx=1, n_entries=6, noise=0, best_score=341.48, all_scores=[341.48]
    entries: [(0.2, 0.6), (0.6, 0.9), (0.9, 1.3), (1.3, 1.8), (1.8, 2.3), (2.3, 3.1)]
  [a_above_b] idx=2, n_entries=8, noise=0, best_score=378.48, all_scores=[378.48, -304.00]
    entries: [0.1, 0.2, 0.6, 0.9, 1.3, 1.8, 2.3, 3.1]

stijnvermeeren-swisstopo:
@AgathaSchmidt do you think that we should just close this PR in favour of https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/463 now, or do you think that there is some value in these changes regardless of #463 ?

AgathaSchmidt:
> @AgathaSchmidt do you think that we should just close this PR in favour of #463 now, or do you think that there is some value in these changes regardless of #463 ?

Yes, I think we should close this PR without merging and go with #463 instead

AgathaSchmidt:
We close this PR because #463 is closing #442

PR #457, opened 2026-04-20
Incorrect overlap between successive boreholes

### LINKED ISSUE #330: Improve layer pairing logic between consecutive pages
The pairing of depth and description remains suboptimal in some cases even after matching logic update PR https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/319 and continuous borehole overlap matching PR https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/317, Issue https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/132

The issue arises when the layers detected on the upper page do not fully match those on the lower page. Since the current model does not handle partial overlaps or mismatched layer sets, it may either skip certain layers or create duplicate entries.

In the example below (`GQ:525`), two issues are visible:
* Extra layer: Ruler ID `12692` is being incorrectly detected as a layer.
Occluded text: The ruler overlaps with the page content, causing the text for interval 62→64 to be cut off.

To address these, we could remove layers that contain only numerical values, and allow partially matched text entries. That said, we need to be careful that these changes don't introduce errors in files that are currently working correctly.

Image
Image

### LINKED ISSUE #454: Incorrect overlap detected between two borehole profiles on consecutive pages
Thurgau (train dataset) file `1780_Romanshorn_933_BS_1_5_14.pdf` page 5-6:

Image

Image

The borehole profile on page 6 is incorrectly identified as having an overlap with the borehole form page 5. Indeed, the material descriptions are very similar. Still, there are several clear indicators that these are really separate boreholes:
- smaller differences in the material descriptions
- different depth values
- different borehole names on the page
- an overlap that spans more than one layer is not really plausible when the second page is a full page scan (with header information / metadata visible above the depths and material descriptions), as opposed to a partial scan of a larger page (where the overlapping data will start immediately at the top of the second page).

We should try to make the requirements for detecting an overlap stricter, while still preserving the current accuracy on any examples with overlap from the other benchmarking datasets (Zurich, GeoQuat).

christianabbet:
Closes #454 and closes #330

# Description

Boreholes from different pages were being incorrectly merged when their material descriptions were similar. This PR addresses the root causes across three areas of the pipeline and simplifies the implementation.

# Changes

* **Partial matching**: The pipeline was designed to allow text cut off at page boundaries, supporting overlap between pages. However, the previous implementation permitted partial overlap for any layer, including those in the middle of a page. This caused a description like "Description A, Description B" to incorrectly match "Description B" even when it wasn't the last layer on the page. Partial matching is now restricted to boundary layers only.

* **Sliding window** (convolutional strategy): The previous approach iterated layer-by-layer through a while loop, checking compliance only for the last matched layer. If multiple layers overlapped, only the final one was evaluated. The new approach uses a convolutional strategy: it iterates over all candidate overlaps simultaneously and scores each one, then selects the best. Benefits: Shorter, more readable code, all overlap candidates are evaluated (not just the last one).

* **Elevation matching**: Added a depth-matching requirement: when depth information is available (start or end), it must match before layers can be merged.

# Performance

| Dataset | Layer F1 (before) | Layer F1 (after) | Notes |
|---------|-----------------|------------------|------|
| Geoquat | 47.7 | 47.6 | Negligible, see details below |
| Zurich  | 84.3 | 84.3 | — |
| Thurgau | — | Fixed |  TH:1780 now correctly separated |

* `GQ:A234`: Elevation bar mismatch. The previous approach accepted a non-monotonic depth sequence (23.35 → 0.6); the new logic requires a consistent span.
* `GQ:A525`: One layer is split into multiple entries due to inconsistent text grouping.

# Test

* Existing tests updated to reflect the convolutional overlap strategy
* New tests added for one-to-one layer similarity checks
* `test_find_last_duplicate_wrong_line_grouping` removed. It tested behaviour that is explicitly no longer coherent under the new sliding window logic

### DIFF
--- a/src/extraction/features/stratigraphy/layer/overlap_detection.py
+++ b/src/extraction/features/stratigraphy/layer/overlap_detection.py
@@ -121,7 +121,7 @@ def find_split_by_convolution(layers_prev: list[Layer], layers_curr: list[Layer]
 
     for i in list(range(1, min(len(layers_prev), len(layers_curr)) + 1)):
         # All layers should match to enforce overlap
-        convoluted_layers_match = [
+        if all(
             _are_layers_similar(
                 layer_prev=layer_prev,
                 layer_curr=layer_curr,
@@ -130,9 +130,7 @@ def find_split_by_convolution(layers_prev: list[Layer], layers_curr: list[Layer]
                 is_extremity=(j == 0 or j == i - 1),  # Indicate to function that one layer might be cut (extremities)
             )
             for j, (layer_prev, layer_curr) in enumerate(zip(layers_prev[-i:], layers_curr[:i], strict=True))
-        ]
-        print(convoluted_layers_match)
-        if all(convoluted_layers_match):
+        ):
             idx_next = i
 
     return idx_next
--- a/src/extraction/features/stratigraphy/layer/overlap_detection.py
+++ b/src/extraction/features/stratigraphy/layer/overlap_detection.py
@@ -83,10 +83,18 @@ def _are_layers_similar(
     # Rule 3: If depth exists, should match
     if force_depth_matching and layer_curr.depths and layer_prev.depths:
         # Rule 3.1 Strating depth should match
-        if layer_curr.depths.start and layer_prev.depths.start and layer_curr.depths.start != layer_prev.depths.start:
+        if (
+            layer_curr.depths.start
+            and layer_prev.depths.start
+            and layer_curr.depths.start.value != layer_prev.depths.start.value
+        ):
             return False
         # Rule 3.2 Ending depth should match
-        if layer_curr.depths.end and layer_prev.depths.end and layer_curr.depths.end != layer_prev.depths.end:
+        if (
+            layer_curr.depths.end
+            and layer_prev.depths.end
+            and layer_curr.depths.end.value != layer_prev.depths.end.value
+        ):
             return False
 
     # All rules validated
@@ -113,7 +121,7 @@ def find_split_by_convolution(layers_prev: list[Layer], layers_curr: list[Layer]
 
     for i in list(range(1, min(len(layers_prev), len(layers_curr)) + 1)):
         # All layers should match to enforce overlap
-        if all(
+        convoluted_layers_match = [
             _are_layers_similar(
                 layer_prev=layer_prev,
                 layer_curr=layer_curr,
@@ -122,7 +130,9 @@ def find_split_by_convolution(layers_prev: list[Layer], layers_curr: list[Layer]
                 is_extremity=(j == 0 or j == i - 1),  # Indicate to function that one layer might be cut (extremities)
             )
             for j, (layer_prev, layer_curr) in enumerate(zip(layers_prev[-i:], layers_curr[:i], strict=True))
-        ):
+        ]
+        print(convoluted_layers_match)
+        if all(convoluted_layers_match):
             idx_next = i
 
     return idx_next
--- a/src/extraction/features/stratigraphy/layer/overlap_detection.py
+++ b/src/extraction/features/stratigraphy/layer/overlap_detection.py
@@ -2,6 +2,7 @@
 
 import logging
 import math
+from dataclasses import dataclass
 
 import Levenshtein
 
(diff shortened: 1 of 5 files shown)


src/extraction/features/stratigraphy/layer/overlap_detection.py
```diff
@@ -26,156 +27,151 @@ def select_boreholes_with_overlap(
...
-    return bottom_duplicate_idx
-
-
-def _is_duplicate(cur_text: str, prev_text: str, t: float):
+            idx_next = i
+
+    return idx_next
+
+
+def _is_duplicate(cur_text: str, prev_text: str, t: float, is_extremity: bool) -> bool:
```
AgathaSchmidt: maybe we should use "threshold" instead of t, to be more precise

src/extraction/features/stratigraphy/layer/overlap_detection.py
```diff
@@ -26,156 +27,151 @@ def select_boreholes_with_overlap(
...
-        # 1. check of the current layer with the compare_against_idx'th layer of previous page.
-        if _is_duplicate(
-            current_layer.material_description.text,
-            previous_page_layer.material_description.text,
-            duplicate_layer_threshold,
+    material_threshold = matching_params["duplicate_layer_threshold"]
+    force_depth = matching_params["duplicate_layer_force_depth"]
+    idx_next = None
+
+    for i in list(range(1, min(len(layers_prev), len(layers_curr)) + 1)):
```
AgathaSchmidt: `range` is iterable, you dont need to wrap it in `list()`

```
for i in range(1, min(len(layers_prev), len(layers_curr)) + 1):
```

stijnvermeeren-swisstopo:
@christianabbet Thanks for your analysis and code changes. I agree that the new behaviour generally makes sense and I like the simplifications. Only for A525, I wanted to keep the matching a little bit more flexible. For that reason, I implemented an exception that says: if there are two consecutive layers that match (including defined depth values), then we accept it as a valid overlap. This improved the results for A525 (even better than the current `develop` branch) while not causing any other side-effects on GeoQuat, Zurich nor the TG 1780 profile.

Other changes:
- Force `duplicate_layer_force_depth` instead of making it configurable. If anything, we could make the number of consecutive layers in my code configurable, but I don't think that's really necessary either.
- Revert the looping order for the convolution check, so that the longest overlaps (which we want to prioritize) are checked first.

I allowed myself to commit these changes directly to your branch.

The code for the nested for loop can probably be cleaned up a little bit, which I will try to do in a minute, but maybe you can already check if you generally agree with my changes.

stijnvermeeren-swisstopo:
@christianabbet I added another commit that further improves the results for A525, which might be sufficient to consider https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/330 to be solved by this PR as well.

Let me know what you think!

PS: for A234, the depth column of the second page is already extracted incorrectly, so the overlap detection doesn't really have any change to work correctly for this file. Therefore, I agree we can ignore that file as far as this PR is concerned.

christianabbet:
@stijnvermeeren-swisstopo LGTM. The code works well on my side. It is indeed cleaner with the object that returns the Overlap. Moreover, 2 layer matching is probably a safe assumption with elevation and descrption alignment.
I added #330 as well in the text even though it was already linked

stijnvermeeren-swisstopo (approved):

src/extraction/features/stratigraphy/layer/overlap_detection.py
```diff
@@ -9,173 +11,184 @@
...
 
-    When a borehole continues on the next page, some top layers of the new page
-    may duplicate the bottom layers of the previous page. This function walks the
-    overlap (from the bottom up) and shifts the split upward if the upper
-    layer lacks depth information while the corresponding lower layer has it.
-    This preserves the most complete layer data across the boundary.
+    Validates layers through three rules:
+    1. Both layers must have material descriptions
+    2. Material descriptions must exceed similarity threshold
+    3. If force_depth_matching is enabled, depths must match
```
AgathaSchmidt: `force_depth_matching ` parameter does not exist

src/extraction/features/stratigraphy/layer/overlap_detection.py
```diff
@@ -9,173 +11,184 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclass
+class OverlapResult:
+    """Data class that contains all information regaring overlapping layers on consecutive pages.
```
AgathaSchmidt: regarding

AgathaSchmidt (changes requested):

AgathaSchmidt (approved):

PR #459, opened 2026-04-21
Sort groundwater entries before JSON dumping

### LINKED ISSUE #458: Non-deterministic ordering of groundwater predictions
# Description

 When dumping predictions for groundwater, the order of entries in the output array is not guaranteed to be consistent across runs. This produces diffs when the output is tracked with version control tools. 

**Run 1**
```json
"groundwater": [
  { "date": "1962-04-27", ... },
  { "date": "1962-01-08", ... }
]
```

**Run 2**
```json
"groundwater": [
  { "date": "1962-01-08", ... },
  { "date": "1962-04-27", ... }
]
```

christianabbet:
Closes #458 

# Description

Sort groundwater entries before JSON dumping based on depth, date and elevation.


AgathaSchmidt (approved):
Thank you for this fix! I helps a lot when checking differences! 
Looks good

PR #460, opened 2026-04-22
Remove gap_with_previous_affinity for DP matching

stijnvermeeren-swisstopo:
While investigating #442 / #456, I made the following observations:
- `gap_with_previous_affinity` and `vertical_spacing_affinity` serve very similar purposes: making a split between layers more likely when the vertical distance between the descriptions is bigger.
- `gap_with_previous_affinity` was only used by default for a small number of sidebars (no sidebar and ProtocolSidebar).
- `gap_with_previous_affinity` did not behave in a reasonable way for many other inputs. In particular, there were issues with negative distance values. Also, because the computation only compared the current gap with the previous gap, it gave inconsistent results when the previous layer's description consisted of a single line vs. multiple lines.
- For #442, @AgathaSchmidt discovered that it could be beneficial to give more weight to the vertical gap between description lines. However, giving more weight to `gap_with_previous_affinity` does not work well in many cases because of the behaviour described in the previous point.
- For this reason, I wanted to check if we could completely remove `gap_with_previous_affinity` and instead only use `vertical_spacing_affinity` as a measure for the vertical gaps. This seems to be possible and gives good results on the Zurich and GeoQuat datasets.

(Note: although these changes are related to / inspired by #422, they don't provide a solution for the issue itself.)

**Zurich**:
- Layer F1 goes up by 0.1%
- Material description F1 goes up by 0.1%
- Improved: 267123120-bp

**Geoquat**
- Layer F1 goes up by 0.01%
- Material description F1 goes down by 0.1%
- Improved: A135
- Worse: A437
- Other changes: 10088, A11406, A1220, A185, B140, B366, V69

I don't think that we will ever get a perfect solution for splitting material descriptions for "no sidebar" borehole profiles by looking at vertical gaps between description lines alone. A more robust solution for splitting material descriptions in the "no sidebar" case might be to take into account the dividing lines in the corresponding strip log (assuming a strip log was detected). But that would be something for a future ticket.

### DIFF
--- a/src/extraction/features/extract.py
+++ b/src/extraction/features/extract.py
@@ -912,9 +912,6 @@ def get_pairs_based_on_line_affinity(
     weights = matching_params["affinity_params"]["no_sidebar"]["weights"]
     threshold = -0.3
 
-    for line, aff in zip(description_lines, affinities, strict=False):
-        print(aff.weighted_affinity(**weights), aff.vertical_spacing_affinity, line.text)
-
     for line_idx, affinity in enumerate(affinities):
         # note: the affinity of the first line is always 0.0
         if affinity.weighted_affinity(**weights) <= threshold:
--- a/src/extraction/features/extract.py
+++ b/src/extraction/features/extract.py
@@ -910,7 +910,10 @@ def get_pairs_based_on_line_affinity(
     pairs = []
     prev_block_idx = 0
     weights = matching_params["affinity_params"]["no_sidebar"]["weights"]
-    threshold = -1
+    threshold = -0.3
+
+    for line, aff in zip(description_lines, affinities, strict=False):
+        print(aff.weighted_affinity(**weights), aff.vertical_spacing_affinity, line.text)
 
     for line_idx, affinity in enumerate(affinities):
         # note: the affinity of the first line is always 0.0
--- a/src/extraction/features/extract.py
+++ b/src/extraction/features/extract.py
@@ -910,11 +910,11 @@ def get_pairs_based_on_line_affinity(
     pairs = []
     prev_block_idx = 0
     weights = matching_params["affinity_params"]["no_sidebar"]["weights"]
-    # The presence of >=3 horiz. lines should tighten the vertical spacing constrain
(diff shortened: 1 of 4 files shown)


src/extraction/features/extract.py
```diff
@@ -910,11 +910,20 @@ def get_pairs_based_on_line_affinity(
     pairs = []
     prev_block_idx = 0
     weights = matching_params["affinity_params"]["no_sidebar"]["weights"]
-    # The presence of >=3 horiz. lines should tighten the vertical spacing constrain
-    threshold = -0.99 if sum(affinity.long_lines_affinity for affinity in affinities) <= -3.0 else 0.0
+
+    horizontal_lines_significance = sum(-affinity.long_lines_affinity for affinity in affinities)
+    vertical_spacing_significance = sum(max(0.0, -affinity.vertical_spacing_affinity) for affinity in affinities)
+    if horizontal_lines_significance > vertical_spacing_significance:
+        # if the precense of horizontal lines seems to be a stronger differentiator than the vertical spacing between
```
christianabbet:
```suggestion
        # if the presence of horizontal lines seems to be a stronger differentiator than the vertical spacing between
```

src/swissgeol_doc_processing/text/textline_affinity.py
christianabbet:
```suggestion
    """Class holding the 6 types of affinities.

    Each affinity is a float between -1.0 and 1.0 and measures how likely two lines belong together.
    
    long_lines_affinity: affinity based on the presence of long horizontal lines inbetween the two text lines.
    lines_on_the_left_affinity: affinity based on the presence of lines cutting the left side of the text lines.
    vertical_spacing_affinity: affinity based on the vertical spacing between the two text lines.
    right_end_affinity: affinity based on the alignment of the right end of the two text lines.
    diagonal_line_affinity: affinity based on the presence of diagonal lines inbetween the two text lines.
    indentation_affinity: affinity based on the indentation difference between the two text lines.
    """
```

christianabbet (commented):
LGTM (minor comments for documentation)

PR #461, opened 2026-04-28
New material description detection

### LINKED ISSUE #437: First lines of material descriptions are often missed
Here are some samples of borehole profiles from a new dataset from the canton of SO.

[635469_244974.pdf](https://github.com/user-attachments/files/26397991/635469_244974.pdf)
[560.pdf](https://github.com/user-attachments/files/26397992/560.pdf)
[637975_245193.pdf](https://github.com/user-attachments/files/26397993/637975_245193.pdf)
[637674_245174.pdf](https://github.com/user-attachments/files/26397994/637674_245174.pdf)

They have something in common, namely that the extracted material descriptions are missing one or a couple of lines at the top of the log.

Image
Image

A quick fix might be to add some additional material description keywords (like "Probe" or "Oberboden") to the config. However, this would make us more dependent on manually configured keywords, while actually we would rather like to become less dependent on them. Indeed, it is quite likely that new possible keywords will keep appearing, and we would never be done fine-tuning the list of material description keywords.

Therefore, I would rather look into the method `_find_all_material_description_candidates` of the `MaterialDescriptionWithSidebarExtractor`. This method has not changed for a long time, and there a quite a few assumptions in the implementation that probably don't generalize very well.

 At first sight, I notice the following:
- There is a `check_y0_condition` that depends on the available sidebar (if any). It ensures that material description lines are "too far away (vertically)" from the detected sidebar. While this does not make a difference for the examples in this ticket, it feels like a relatively flaky condition.
- What does make a difference for these examples, is that while we check for additional lines that continue a material description at the bottom (even when they don't have a matched keywords), we don't do that for additional lines at the top. In other words, we don't check if there are text lines immediately above the first line with a matched material description keywords, that don't have a matched key themselves, but based on the positioning/layout look like they should be part of the material descriptions as well.
- The method does not make use of any detected stip logs or table structure, since those were only added to the repo at a later point in time. I think that there is a lot of potential to make use of these to get a quick estimate of where the first line of the first material description might be.

So the goals for this ticket are:
- If possible, simplify the logic in `_find_all_material_description_candidates`. I would not be opposed to completely throwing out the existing code and coming up with a new logic (maybe more strongly based on strip logs and/or table structures).
- Ensure that the material description of the first layer in the four examples above is correctly included in the extracted data.
- Ensure that there is no degradation on the Zurich and Geoquat benchmarks.

AgathaSchmidt:
Closes #437 

 **Before**, the first lines of material descriptions were missed sometimes because the detection relied almost entirely on keyword inclusion lists. 
We want to become less dependent on the keyword list. 

The current ideas which could help to detect the material description --> focus on geometrical features and quality of content regardless of keywords
- prefer text which is close to a sidebar/striplog  or which is within a table structure
- check if line has one real world —> exclude numbers and pure abbreviations 
- md column has to have at least one keyword matching 
- reduced keyword list 
- check if two pairs are horizontally adjacent with vertical overlap -> for cases where we detected a trash column tright next to the real md column —> ignore this one
- created header keyword list and check if the column has a header matching the keyword _has_column_header
check for horizontal line above md —> exclude header from being added to md cluster 
- also use strip log top edge as a boundary, because it is unlikely to find md above that —> same with table top edge 
prefer wide columns over narrow columns
- `_get_geometric_description_seeds` —> check for md on the sides of the striplog
-  Upward expansion of the md cluster (is_above loop)
- Multi-line column header expansion: after detecting a header like "Beschreibung", the code iteratively extends header_y1 downward to absorb continuation lines (e.g. "des aufgeschlossenen Bohrgutes") before using it as the cluster's upper boundary.
- Column header excluded from description output text
- 
Currently we are fixing the missing first lines issue described in the issue #437 (but adding DP as well) but overall we do not reach the original performance with the keyword based md detection 

<google-sheets-html-origin>
develop | 0.842 | 0.893 | 0.88 | 0.373 | 0.814 | 0.694 | 0.73
437 |0.81 | 0.872 | 0.852 | 0.389 | 0.8395 | 0.718 | 0.755

<google-sheets-html-origin>
develop | 0.479 | 0.6296 | 0.628 | 0.05 | 0.615 | 0.475 | 0.604
437 | 0.461 | 0.607 | 0.596 | 0.056 | 0.635 | 0.489 | 0.607

There are some examples where the first lines are now detected as well but the main remaining issues are 
- additional lines are included 
- false positives besides the actual material description
- the _line_has_real_word is very strict and is extracting 

I tried to fix some of these, e.g. by not checking `_line_has_real_word` in `_get_interval_block_pairs` or `_get_geometric_description_seeds` which solved the issue with missed non-word lines within a material description block but overall performance got worse, because of FP 
But I cannot beat the current performance despite the efforts.

By implementing only the `is_above` loop we solve the issue and reach _okay_ metrics scores (improving mainly depth_intervall and losing a few percent on the rest) but the keyword logic stays the same. 

<google-sheets-html-origin>
geoquat | layer_f1 | depth_intervall_f1 | Material_description_f1 | groundwaterf1 | elevation_f1 | coordinates_f1 | Borehole_name_f1
develop | 0.479 | 0.6296 | 0.628 | 0.05 | 0.615 | 0.475 | 0.604
check above | 0.474 | 0.657 | 0.637 | 0.039 | 0.558 | 0.376 | 0.643

(EDIT cant reproduce those numbers anymore --> see #470 for current implementation of check above)

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,19 +95,6 @@ We use [uv](https://docs.astral.sh/uv/) to manage package dependencies. Install 
 curl -LsSf https://astral.sh/uv/install.sh | sh
 ```
 
-This repository uses [Git LFS](https://git-lfs.github.com) to store the BERT model files. Install and enable it before cloning so the model files are downloaded automatically:
-
-```bash
-brew install git-lfs   # macOS — see https://git-lfs.github.com for other platforms
-git lfs install        # one-time setup per machine
-```
-
-If you already cloned without LFS installed, fetch the model files afterwards:
-
-```bash
-git lfs pull
-```
-
 The below commands will install the package for you (assuming you have successfully cloned the repository):
 ```bash
 uv sync --all-extras
@@ -216,11 +203,9 @@ To execute the layer description classification, follow these steps:
 ### 1. Setup
 
 Repeat steps 1 and 2 of the [data extraction pipeline](#run-data-extraction) to set up the environment and download the data.
-
-The pre-trained BERT models (`backbone`, `lithology_head`, `en_main_head`) are stored in the repository via Git LFS and are fetched automatically when cloning with LFS installed. If the `models/` directory contains only small pointer files instead of the actual weights, run:
-
+Pre-trained models can be downloaded with
 ```bash
-git lfs pull
+aws s3 sync s3://stijnvermeeren-boreholes-models models/uscs/your_model_folder
 ```
 
 ### 2. Run the Classification Pipeline
--- a/README.md
+++ b/README.md
@@ -95,6 +95,19 @@ We use [uv](https://docs.astral.sh/uv/) to manage package dependencies. Install 
 curl -LsSf https://astral.sh/uv/install.sh | sh
 ```
 
+This repository uses [Git LFS](https://git-lfs.github.com) to store the BERT model files. Install and enable it before cloning so the model files are downloaded automatically:
+
+```bash
+brew install git-lfs   # macOS — see https://git-lfs.github.com for other platforms
+git lfs install        # one-time setup per machine
+```
+
+If you already cloned without LFS installed, fetch the model files afterwards:
+
+```bash
+git lfs pull
+```
+
 The below commands will install the package for you (assuming you have successfully cloned the repository):
 ```bash
 uv sync --all-extras
@@ -203,9 +216,11 @@ To execute the layer description classification, follow these steps:
 ### 1. Setup
 
 Repeat steps 1 and 2 of the [data extraction pipeline](#run-data-extraction) to set up the environment and download the data.
-Pre-trained models can be downloaded with
+
+The pre-trained BERT models (`backbone`, `lithology_head`, `en_main_head`) are stored in the repository via Git LFS and are fetched automatically when cloning with LFS installed. If the `models/` directory contains only small pointer files instead of the actual weights, run:
+
 ```bash
-aws s3 sync s3://stijnvermeeren-boreholes-models models/uscs/your_model_folder
+git lfs pull
 ```
 
 ### 2. Run the Classification Pipeline
--- a/docs/API_and_Docker.md
+++ b/docs/API_and_Docker.md
@@ -53,42 +53,6 @@ Additional endpoints and their functionalities can be found in the project's sou
 To stop the FastAPI server, press `Ctrl + C` in the terminal where the server is running. Please refer to the [FastAPI documentation](https://fastapi.tiangolo.com) for more information on how to work with FastAPI and build APIs using this framework.
 
 
-## Classification Model Configuration
-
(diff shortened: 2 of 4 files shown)


stijnvermeeren-swisstopo:
I think that there are several good ideas here. It's also very promising that there is a significant number of files for which the accuracy improves.

A few thoughts based on a look at the results for the Zurich dataset...
- `_line_has_real_word` seems too strict indeed. It filters out correct lines such as "2.4- 2.6m do." (`268124354-bp.pdf`), and it completely rejects all-uppercase descriptions (`268124232-bp.pdf`, "HUMUS" in `267125361-bp.pdf` and `267125358-bp.pdf`). Similarly, the strict exclusion of header keywords is sometimes counterproductive, such as in `267123107-bp.pdf`, where a correct description line containing the word "Lithologie" is incorrectly removed.
- Even when allowing uppercase descriptions for `268124232-bp.pdf`, the code line `candidate_rects = [max(candidate_rects, key=lambda r: (self._has_keyword_match(r), r.width))]` removes the correct match for the second (right-hand-side) borehole profile on that page. I don't fully understand yet why this code is needed. Would it be possible to achieve the same thing by fine-tuning the scoring function for the sidebar-material-description-pair?
- We could probably be a bit stricter in filtering out some false positives, especially for material descriptions without sidebar. Maybe we could require that (if there is no sidebar) then most description lines should be relatively nicely aligned with each other? That could avoid problematic cases such as the FP in `268124320-bp.pdf` and `267123080-bp.pdf`.
- We should probably be a little bit more flexible with the `y0_min` for upwards expansion of material descriptions. For example in `268124571-bp.pdf`, the first description line is a few points above the top of the detected strip log.
- In the case of an AAboveBSidebar, we could also try putting an `upward_limit` for the upward expansion at the (extrapolated) y-coordinate of the 0.0 depth. This could avoid some incorrect upward expansions such as in `267124190-bp.pdf`, `267124097-bp.pdf` and `268125729-bp.pdf`.
- The detection of a separating line to stop the upwards expansion fails when the line is already intersecting the upper part of the bounding of box of the first already included line (e.g. `679249004-bp.pdf`, `268125729-bp.pdf`, `680248008-bp.pdf`)

Given the complexity of the implementation, I think that it would be beneficial to go step-by-step. Your idea of first only implementing the `is_above` loop and only later changing the keyword logic sounds good to me.

By doing smaller changes, it will also become easier to assess the impact of the individual changes, and spot when some change is not really necessary, because I feel like it should be possible to simplify the current implementation somewhat, but it's not immediately clear where/how.

Finally, it would be really beneficial to make the code a little bit more modular, i.e. avoid very long methods, and move certain parts of code away from the huge `extract.py` file.

AgathaSchmidt:
As proposed we split this issue 
--> check above, solving the original issue #437 was merged with #470 
--> opened a new issue for restructuring the material description detection #471

PR #463, opened 2026-04-28
#442: fallback from AAboveBSidebar to ProtocolSidebar if it's a perfect fit

### LINKED ISSUE #442: Fix mismatching of material descriptions and depth entries for specific borehole type (Sondierprofil)

stijnvermeeren-swisstopo:
Resolves #442

Builds on top of #460, and is inspired by the discussions with @AgathaSchmidt in the context of #456.

Fixes the two examples from the issue, as well as significantly improves A11358, leading to a 0.3% improvement in Layer F1 score for GeoQuat validation. There are no other changes on Zurich or GeoQuat validation (compared to #460).

@AgathaSchmidt what do you think of this approach? Before merging, I would still improve a couple of things, but maybe you can first let me know what you think of this method in general.

To be improved:
- Parameter descriptions in the docstrings of the new methods.
- The threshold value `0.1` should probably be configurable.
- The interval zones for the ProtocolSidebar are defined in a rather counterintuitive way. I had to use `start` whereas I intuitively thought that I would have to use `end`. Maybe this could be simplified as well?

3184_KB_41_37_Kreuzlingen_Weinberg:
image

3266_RKB_41_131_Kreuzlingen_Weinberg:
image

A11358 new:
image

A11358 old:
image

### DIFF
--- a/src/app/api/v1/endpoints/extract_stratigraphy.py
+++ b/src/app/api/v1/endpoints/extract_stratigraphy.py
@@ -75,7 +75,8 @@ def extract_stratigraphy(filename: str, include_groundwater: bool = False) -> Ex
     boreholes_per_page = []
     for page_index, page in enumerate(document):
         # 2. load the png image to infer the scaling, MUST have been generated before
-        png_page = load_png(filename, page_index + 1)  # page number is 1-indexed
+        page_number = page_index + 1  # page number is 1-indexed
+        png_page = load_png(filename, page_number)
         pdf_img_scalings.append((png_page.shape[1] / page.rect.width, png_page.shape[0] / page.rect.height))
 
         # 3. extract layers
@@ -95,7 +96,7 @@ def extract_stratigraphy(filename: str, include_groundwater: bool = False) -> Ex
             table_structures,
             strip_logs,
             language,
-            page_index,
+            page_number,
             page.rect.width,
             page.rect.height,
             line_detection_params,
@@ -107,7 +108,7 @@ def extract_stratigraphy(filename: str, include_groundwater: bool = False) -> Ex
         # Extract groundwater if requested (uses the extracted boreholes as context)
         if include_groundwater and groundwater_extractor and groundwater_in_doc:
             groundwater_entries = groundwater_extractor.extract_groundwater(
-                page_number=page_index + 1,
+                page_number=page_number,
                 text_lines=text_lines,
                 geometric_lines=long_or_horizontal_lines,
                 extracted_boreholes=extracted_boreholes,
--- a/src/app/api/v1/endpoints/extract_stratigraphy.py
+++ b/src/app/api/v1/endpoints/extract_stratigraphy.py
@@ -8,7 +8,7 @@ from app.common.schemas import (
     ExtractStratigraphyResponse,
     GroundwaterSchema,
 )
-from extraction.features.extract import extract_page
+from extraction.features.extract import MaterialDescriptionRectWithSidebarExtractor
 from extraction.features.groundwater.groundwater_extraction import (
     GroundwaterInDocument,
     GroundwaterLevelExtractor,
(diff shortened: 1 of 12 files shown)


AgathaSchmidt:
I really like this approach! 
It is very elegant and short and leaves AtoB_sidebars as they are 

The only question I have: 
We transform the AaboveB sidebar entries to a protocol sidebar by `protocol_sidebar = ProtocolSidebar(sidebar.entries)` correct?  We just wrap the already extracted AaboveB entries into a ProtocolSidebar object?
This means that when the AaboveB sidebar does not extract all sidebar entries, we end up with an incomplete protocol sidebar as well, even though the protocol sidebar extractor _might_ be able to extract all sidebar entries? 
Currently, I do not have a specific case for this in mind (I was just inspired by the discussion with the AtoB sidebar before) but I just wanted to make sure that I understand the concept

stijnvermeeren-swisstopo:
@AgathaSchmidt yes that's a good point. E.g. if you increase `corr_coef_threshold` to `0.99` in `is_valid` in `a_above_b_sidebar_validator`, then for the Thurgau examples from this ticket, the borehole profile is still interpreted as a ProtocolSidebar, but some entries are missing, because they were already removed by the `AAboveBSidebar.remove_entry_by_correlation_gradient` method.

On the other hand, I don't want to rerun the protocol sidebar extraction from scratch for the entire page. I just want to check if the current `AAboveBSidebar` works better when interpreted as a `ProtocolSidebar`.

For this reason, I now keep track of the unfiltered entries (i.e. before any `remove_entry_by_correlation_gradient` application) for the `AAboveBSidebar`, and use those for constructing the `ProtocolSidebar`.

While this does not have an impact on any of our benchmarking files, it seems to be a more robust and conceptually correct solution in generall.

stijnvermeeren-swisstopo:
I also extended the docstrings, make the threshold configurable, and applied a little code cleanup/restructuring. @letao this includes the two wrapper methods that we talked about in the daily this morning.

AgathaSchmidt:
we also have a positive effect on zurich 267123120, where the material descriptions are now matched better

AgathaSchmidt:
> For this reason, I now keep track of the unfiltered entries (i.e. before any `remove_entry_by_correlation_gradient` application) for the `AAboveBSidebar`, and use those for constructing the `ProtocolSidebar`.
 
I think this is a good idea, since the dropping of values i order to fit into the correlation threshold is one of the main reasons why AAboveB sidebar is missing values

stijnvermeeren-swisstopo:
@AgathaSchmidt the positive effect on 267123120 comes from #460

AgathaSchmidt:
It looks good to me

AgathaSchmidt (approved):

letao (approved):
Looks good to me, from what I have seen no breaking changes in the minimal pipeline. Currently assets imports only:

`from extraction.minimal_pipeline import ExtractionContext, extract_page_features`
 
Both of them are unchanged + all the fields that are accessed there are present and unchanged. 

One thing I noticed is that `MaterialDescriptionRectWithSidebarExtractor` is constructed twice when `extract_boreholes=True ` but this was already present before and current approach just makes it more visible. This could be changed but I do not think we need to solve it as part of this issue and I can create a quick /fix or /bug issue for this.

PR #470, opened 2026-05-05
Feat/issue 437/check for md above

### LINKED ISSUE #437: First lines of material descriptions are often missed

AgathaSchmidt:
Closes #437 
- 
## Changes

We introduce 'is_above' which checks if there are other material description lines above the currently identified lines available, similar to 'is_below' checks if there are more lines below the current lines. 
We differentiate between two cases: 
**a)** a sidebar is available --> check new lines iteratively --> when one new line is detected, check if we can have a perfect match with the depth intervals before searching for new lines 
**b)** no sidebar available --> same search behaviour as we have in `is_below` --> since this led to FP lines (mostly header lines (GQ 'A1287' , 'V116')) we check wether the new line might be the header column 

Since we learned in #461 that using keywords or separation lines for column header detection is not robust we check if the candidate line has any other lines on the page which shares approx. the same y-level --> column header 
So far this approach is robust, mitigating the negative effects of the mentioned missbehaviour while not introducing further issues. 

## Main effect
**Now, we detect the first lines of all 4 files mentioned in #437** 
Screenshot 2026-05-05 at 14 58 23

Screenshot 2026-05-05 at 15 10 18

Screenshot 2026-05-05 at 14 58 42

Screenshot 2026-05-05 at 15 12 09

## Side effects 
<google-sheets-html-origin>
develop | 0.843 | 0.893 | 0.881 | 0.373 | 0.814 | 0.694 | 0.73
checkabove | 0.842 | 0.893 | 0.881 | 0.373 | 0.814 | 0.694 | 0.73

<google-sheets-html-origin>
develop | 0.484 | 0.631 | 0.629 | 0.05 | 0.615 | 0.475 | 0.604
checkabove | 0.483 | 0.633 | 0.629 | 0.05 | 0.611 | 0.472 | 0.601

**zurich** 

267125047,  first md line detected (but incorrect due to OCR mistake)

268124094 first line found (matched with correct interval, but which was not 100% correctly extracted)

268124307 , 690251013 FP line added in FP block 

684252058 new md found but matching is still chaotic

**geoquat**

_positive:_ 
- 1506 better  

first line(s) added
- A430
- A437 (page2)
- A472
- A481
- B366
- A7012
- A11395

_negative:_ 
- 12305 , 12306 first line incorrectly includes sidebar entries now 
- A11406 new FP lines added to FP block 
- 2537 one FP line added 

---
--> if we are able to eliminate the FP blocks in the future they will not be extended via `is_above` (A11406, 268124307 , 690251013 )


src/extraction/features/extract.py
```diff
@@ -499,6 +500,44 @@ def is_below(best_x0, best_y1, line: TextLine):
...
 
+            # Expand upward one line at a time.
+            # With sidebar: stop at the topmost entry's y-level (avoids column headers above first depth entry).
+            # Without sidebar: stop when candidate has sibling lines outside the column (header row signal).
+            min_y0_limit = (
+                min(e.rect.y0 for e in sidebar.entries) + 5
+                if sidebar is not None and sidebar.entries
+                else -float("inf")
+            )
+            sorted_above = sorted(candidate_description, key=lambda c: c.rect.y0, reverse=True)
```
christianabbet: sort can be done before for loop once

src/extraction/features/extract.py
```diff
@@ -499,6 +500,44 @@ def is_below(best_x0, best_y1, line: TextLine):
...
+            while best_y0 > min_y0_limit:
+                next_line = next(
+                    (
+                        desc_line
+                        for desc_line in sorted_above
+                        if desc_line.rect.x0 > best_x0 - 5
+                        and desc_line.rect.x0 < (best_x0 + best_x1) / 2
+                        and desc_line.rect.y1 > best_y0 - 10
```
christianabbet: do you think we could have it as a vairable like `margin=5` to avoid the raw values in the code like that ?

christianabbet (changes requested):
Two small comments. Test worked good on my side

src/extraction/features/extract.py
AgathaSchmidt: yes, and I will change that for is_below as well 
(and will create a proper is_above function)

christianabbet (approved):
LGTM

PR #472, opened 2026-05-07
Generate GT for Thurgau dataset

### LINKED ISSUE #462: Create Thurgau ground truth file with relevant boreholes and material description for lithology classification
Thurgau ground truth currently does not have data for the material descriptions. However, this is necessary for training and evaluating lithology classification models. Thurgau very valuable labelled data for some of the attributes that we want to classify (e.g. color). Therefore, we should enrich the Thurgau GT with extracted descriptions where they are reliable enough.
 
- Take only those boreholes where the extracted layers match the GT depths exactly (ca. 35%, see https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/418)
- Add the extracted material descriptions to the GT.
- Store those in a new file that can be used in the lithology classification tasks.

christianabbet:
Closes #462

# Description

* `is_correct` is now included in the file output for layer-level predictions, making it possible to determine whether depth is correctly matched for each layer.
* Pydantic model validators have been added for ground truth data. 
* The forward pass has been refactored to eliminate local dictionaries, ensuring evaluated data is always used instead.
* A new script has been created to update ground truth for the Thurgau dataset based on predictions. The model treats a material description as correct when its depth interval was properly matched in the prediction.

The updated ground truth is available in the S3 bucket as `thurgau_ground_truth.462.json`. The previous ground truth has been kept in the meantime.
To validate the new ground truth, predictions were run on the Thurgau small dataset to verify that metrics are consistent between the original (`thurgau_ground_truth.json`) and the updated version (`thurgau_ground_truth.462.json`).


src/core/ground_truth.py
```diff
@@ -0,0 +1,152 @@
...
+    unconsolidated: GroundTruthUnconsolidated | None = None
+
+
+class GroundTruthGroundwater(BaseModel):
+    """A recorded groundwater measurement."""
+
+    model_config = {"extra": "forbid"}
+    date: str | None = None
+    depth: float
+    elevation: float
```
AgathaSchmidt, line 73: should we make this optional as well ? 
elevation: float | None = None

src/extraction/evaluation/layer_evaluator.py
AgathaSchmidt: its not only about groundwater levels 

change to something like 
"""Classes for evaluating the layer and depth predictions of a borehole."""

src/extraction/evaluation/metadata_evaluator.py
AgathaSchmidt: elevation

AgathaSchmidt (changes requested):
Works well 
Script runs fine, tests work as well

src/core/ground_truth.py
christianabbet: I don't think so. I could not find any example where groundwater was present but elevation was not

AgathaSchmidt (approved):

PR #482, opened 2026-05-27
Support multiple datasets for classification

### LINKED ISSUE #480: Support multiple datasets for training and test sets
# Summary
The pipeline currently only accepts a single dataset for training and testing. We need to support multiple input datasets, concatenating them into a single training set and allowing multiple independent test sets.

# Desired Behavior
Training accepts multiple datasets, concatenated into one unified training set (eg: conso + unconso)
Testing accepts multiple datasets, each evaluated independently (eg: conso, unconso)

christianabbet:
Resolves #480

## Summary
- Extends the classification training pipeline to support multiple training/test datasets via a typed configuration system
- Replaces single-dataset config with a list-based typed config allowing flexible dataset combinations
- Remove `src/classification/models/config.json` that seems to have no utilization


src/classification/models/train.py
```diff
@@ -147,42 +168,43 @@ def train_model(config_file_path: Path, out_directory: Path, model_checkpoint: P
...
+    bert_model = BertModel(model_config.model_path, classification_system)
     bert_model.freeze_all_layers()
-    bert_model.unfreeze_list(model_config.get("unfreeze_layers", []))
+    bert_model.unfreeze_list(model_config.unfreeze_layers)
     bert_model.model.train()
 
     # Load datasets
     logger.info("Loading datasets (transformers library).")
-    train_dataset, eval_dataset, test_dataset = setup_data(bert_model, model_config)
+    train_dataset, eval_dataset, test_datasets = setup_data(bert_model, model_config)
```
AgathaSchmidt, line 191: Due to our special splitting strategy we sometimes end up with 0 samples in the test dataset, especially when the dataset is small, as we discussed already once 

Maybe we need to rethink this approach in the future, when we have to deal with smaller datasets for other tasks. 

Do you think we should implement a Warning when we have 0 samples in the test dataset? 
Maybe printing "Test: 0 samples" might not be enough to inform users

src/classification/utils/datasets/classification.py
AgathaSchmidt: not really belonging to this issue but `get_class_from_entry` is not used anymore

src/classification/models/train.py
```diff
@@ -133,9 +154,9 @@ def common_options(f):
 @common_options
 def train_model(config_file_path: Path, out_directory: Path, model_checkpoint: Path):
```
AgathaSchmidt: `model_checkpoint: Path | None` , see click.option definition

AgathaSchmidt (changes requested):
Looks good and works on my side 
Maybe we should implement a warning when we have 0 samples in the test dataset, as its a common issue for small datasets, as discussed

src/classification/models/train.py
christianabbet: I agree we can maybe change the logic in the future. However I think we should keep the per document split, meaning that with 70 / 15 / 15 we should (statistically) at least have 10-20 different documents to train the model. Otherwise we should increase the sets with Generated examples. If it's fine for you, let's keep it like this for now. I'll add the warning indeed

AgathaSchmidt (approved):

PR #483, opened 2026-05-27
Feat/issue 481/log per class results and confusion matrix for validation and test

### LINKED ISSUE #481: Log per-class results and confusion matrix for validation and test
# Summary

Add detailed per-class metrics and confusion matrix logging across all output channels: terminal, local files, and MLflow, for both validation and test phases.

# Desired Behavior

Per-class metrics (precision, recall, F1, support) and confusion matrix are logged:
* Terminal: printed at the end of validation and test
* Files: saved locally (e.g. CSV/JSON for per-class metrics, PNG/CSV for confusion matrix)
* MLflow: logged as artifacts and/or metrics for both validation and test runs

AgathaSchmidt:
Closes #481 

Changes:

- Log per-class precision/recall/F1 and confusion matrices via a new training callback to MLFlow and also save them as csv and print it our it the terminal 
- create a confusion matrix plot ad save it as an artifact to MLFlow


src/classification/models/train.py
```diff
@@ -35,6 +45,75 @@
 mlflow_tracking = os.getenv("MLFLOW_TRACKING") == "True"
 
 
+class PerClassMetricsCallback(TrainerCallback):
+    """Logs per-class precision/recall/F1 to terminal, CSV, and MLflow after each evaluation."""
+
+    def __init__(self, bert_model: BertModel, eval_dataset: datasets.Dataset, out_directory: Path):
+        self._bert_model = bert_model
+        self._eval_dataset = eval_dataset
+        self._out_directory = out_directory
```
christianabbet: Not needed (see below)

src/classification/models/train.py
```diff
@@ -188,7 +267,33 @@ def train_model(config_file_path: Path, out_directory: Path, model_checkpoint: P
...
+    # Per-class metrics on test set
+    preds = [bert_model.id2classEnum[i] for i in test_results.predictions.argmax(axis=-1)]
+    labels = [bert_model.id2classEnum[i] for i in test_results.label_ids]
+    test_metrics = per_class_metric(preds, labels)
+
+    header = f"{'Class':<30} {'Precision':>10} {'Recall':>10} {'F1':>10} {'TP':>6} {'FP':>6} {'FN':>6}"
+    logger.info("\n=== Per-class metrics — Test ===")
+    logger.info(header)
+    for cls, m in sorted(test_metrics.items(), key=lambda x: x[0].value):
+        logger.info(
```
christianabbet: Instead of code duplication, the test prediction should call `on_predict`. Then we can use the same logic as in `on_evaluate`

src/classification/models/train.py
```diff
@@ -35,6 +45,75 @@
...
 
+class PerClassMetricsCallback(TrainerCallback):
+    """Logs per-class precision/recall/F1 to terminal, CSV, and MLflow after each evaluation."""
+
+    def __init__(self, bert_model: BertModel, eval_dataset: datasets.Dataset, out_directory: Path):
+        self._bert_model = bert_model
+        self._eval_dataset = eval_dataset
+        self._out_directory = out_directory
+
+    def on_evaluate(
```
christianabbet: The signature in TrainerCallback shows args (output_dir), model and metrics are available with callback. Instead of recomputing the metrics in the `on_evaluate` callback, compute it directly in `compute_metrics`, then log and save in callback

src/classification/utils/plots.py
```diff
@@ -0,0 +1,143 @@
...
+
+    Args:
+        predictions (list[ClassificationSystem.EnumMember]): Predicted classes.
+        labels (list[ClassificationSystem.EnumMember]): Ground truth classes.
+
+    Returns:
+        tuple[np.ndarray, list[str]]: Raw confusion matrix and class names.
+    """
+    all_classes = sorted(set(predictions) | set(labels), key=lambda c: c.value)
+    class_names = _get_class_names(all_classes)
```
christianabbet: Already computed above in model ? (id2classEnum, id2label, label2id)

src/classification/models/train.py
```diff
@@ -35,6 +45,75 @@
...
+        self,
+        args: TrainingArguments,
+        state: TrainerState,
+        control: TrainerControl,
+        model=None,
+        **kwargs,
+    ):
+        if model is None:
```
christianabbet: Documentation

src/classification/models/train.py
```diff
@@ -35,6 +45,75 @@
...
+        self._bert_model = bert_model
+        self._eval_dataset = eval_dataset
+        self._out_directory = out_directory
+
+    def on_evaluate(
+        self,
+        args: TrainingArguments,
+        state: TrainerState,
+        control: TrainerControl,
+        model=None,
```
christianabbet: Typing

src/classification/models/train.py
```diff
@@ -35,6 +45,75 @@
...
+
+        # --- CSV ---
+        rows = [{"class": cls.name, **m.to_json()} for cls, m in metrics.items()]
+        df = pd.DataFrame(rows).sort_values("class")
+        csv_path = self._out_directory / f"per_class_metrics_eval_epoch{epoch}.csv"
+        self._out_directory.mkdir(parents=True, exist_ok=True)
+        df.to_csv(csv_path, index=False)
+
+        # --- MLflow ---
+        if mlflow_tracking:
```
christianabbet: I think you can use the `arg[report_to]` to check to MLFLow logging

christianabbet (changes requested):
Let's already update the base logic and then go for a second round. You can use the compute metric function to compute metric at the same time then feed to the callback. The trainer callback already gives you access to all variables you should need

src/classification/models/train.py
```diff
@@ -351,8 +434,13 @@ def compute_metrics(eval_pred: EvalPrediction) -> dict[str, float]:
...
         logits, labels = eval_pred
         predictions = logits.argmax(axis=-1)
-        metrics = per_class_metric(predictions, labels)
-        return AllClassificationMetrics.compute_micro_average(metrics.values())
+        pred_classes = [bert_model.id2classEnum[p] for p in predictions]
+        label_classes = [bert_model.id2classEnum[lbl] for lbl in labels]
+        per_class = per_class_metric(pred_classes, label_classes)
+        per_class_callback.last_per_class_metrics = per_class
+        per_class_callback.last_preds = pred_classes
+        per_class_callback.last_labels = label_classes
```
christianabbet: I would say not good practice to modify arguments in another function like that. I would just compute all the thing needed for confusion matrix and return for logging. "class1_tp", " class2_tp", .... That way they appear in the logging on mlflow. I had something like that in mind (you can change the keying).

```python

# Define a custom compute_metrics function
def make_compute_metrics(id2enum: dict) -> dict[str, float]:

    def compute_metrics(eval_pred: EvalPrediction) -> dict[str, float]:
        """Function used for evaluating prediction, and logging during the training.

        Note: The metrics are not used to optimize the model during the training, just to evaluate it.
            The model is trained by trying to lower the cross-entropy loss.

        Args:
            eval_pred (EvalPrediction): Object of type EvalPrediction that will be passed to this function.

        Returns:
            dict[str, float]: Dictionary containing all the metrics produced to evaluate the predictions.
        """
        logits, labels = eval_pred
        predictions = logits.argmax(axis=-1)
        pred_classes = [id2enum[p] for p in predictions]
        label_classes = [id2enum[lbl] for lbl in labels]
        per_class = per_class_metric(pred_classes, label_classes)
        per_class_metrics = {
            f"{key}_{key.name}_{subkey}": value
            for key, metric in per_class.items()
            for subkey, value in metric.to_json().items()
        }
        overall = AllClassificationMetrics.compute_micro_average(per_class.values())
        overall_metrics = {f"all_{key}": value for key, value in overall.items()}
        return overall_metrics | per_class_metrics

    return compute_metrics

...
    trainer = HeadOnlyTrainer(
        model=bert_model.model,
        args=training_args,
        train_dataset=train_dataset,
        eval_dataset=eval_dataset,
        processing_class=bert_model.tokenizer,
        data_collator=DataCollatorWithPadding(tokenizer=bert_model.tokenizer),
        compute_metrics=make_compute_metrics(bert_model.id2classEnum),
        compute_loss_func=compute_loss_func,
        callbacks=[per_class_callback],
    )
```

Then we can have the function outside setup_trainer Another issue is that we don0t compute the metrics for classes that are not present in the evaluation set. We need to make sure that the number of classes matched the size of the logits. Then we can transmit the metrics from all data to the later stage to just display it nicely.

src/classification/models/train.py
```diff
@@ -336,6 +417,8 @@ def setup_trainer(
     # load the training arguments from the config file
     training_args = setup_training_args(model_config, out_directory)
 
+    per_class_callback = PerClassMetricsCallback(out_directory, list(bert_model.id2classEnum.values()))
```
christianabbet: I think if we follow the logic below, the function should not need any argument in init. out_directory should already be available in args (in `on_evaluate` and `on_predict`) ?

src/classification/models/train.py
```diff
@@ -36,6 +46,80 @@
...
+        state: TrainerState,
+        control: TrainerControl,
+        **kwargs,
+    ):
+        """Huggingface Trainer callback that runs after predict to log per-class metrics and confusion matrices."""
+        if self.last_per_class_metrics is None:
+            return
+        self._log_and_save("test", args)
+
+        raw_path, pct_path = plot_confusion_matrices(
```
christianabbet: the logic would be to have


```python
    def on_predict(
        self,
        args: TrainingArguments,
        state: TrainerState,
        control: TrainerControl,
        metrics: dict[str, float],
        **kwargs,
    ):
        raw_path, pct_path = plot_confusion_matrices( metrics)
```

I think we do not even need _log_and_save afterward

christianabbet (changes requested):

src/classification/models/train.py
```diff
@@ -291,6 +297,11 @@ def compute_trainset_weights(
 class HeadOnlyTrainer(Trainer):
     """Trainer that saves only fine-tuned parameters at every checkpoint."""
 
+    def __init__(self, *args, id2class_enum: dict, out_directory: Path, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._id2class_enum = id2class_enum
+        self._out_directory = out_directory
+
```
christianabbet: We don't need this anymore I think

src/classification/models/train.py
```diff
@@ -314,6 +325,55 @@ def clean_checkpoints(self) -> None:
...
 
 
+class ConfusionMatrixCallback(TrainerCallback):
+    """Trainer callback to compute and save confusion matrix after evaluation."""
+
+    def __init__(self, id2class_enum: dict, out_directory: Path):
+        self._id2class_enum = id2class_enum
+        self._out_directory = out_directory
+        self._preds = None
+        self._labels = None
```
christianabbet: can't we just store ?


```suggestion
        self._cm: np.ndarray | None = None
```

src/classification/models/train.py
```diff
@@ -314,6 +325,55 @@ def clean_checkpoints(self) -> None:
...
+        self._labels = labels
+        pred_classes = [self._id2class_enum[p] for p in predictions]
+        label_classes = [self._id2class_enum[lbl] for lbl in labels]
+        per_class_results = per_class_metric(pred_classes, label_classes)
+        class_metrics = {cls: per_class_results.get(cls, Metrics()) for cls in self._id2class_enum.values()}
+        per_class_metrics = {
+            f"{cls.name}_{subkey}": value
+            for cls, metric in class_metrics.items()
+            for subkey, value in metric.to_json().items()
+        }
```
christianabbet: If we passe the CM then I think you where right, we can just keep the F1 score for each class

```suggestion
        per_class_metrics = {f"{cls.name}_f1": metric.f1 for cls, metric in class_metrics.items()}
```

src/classification/models/train.py
```diff
@@ -314,6 +325,55 @@ def clean_checkpoints(self) -> None:
...
+        overall = Metrics.micro_average(class_metrics.values())
+        overall_metrics = {
+            "all_micro_precision": round(overall.precision, 4),
+            "all_micro_recall": round(overall.recall, 4),
+            "all_micro_f1": round(overall.f1, 4),
+        }
```
christianabbet: What about this ?

```suggestion
        overall_metrics = Metrics.micro_average(class_metrics.values()).to_dict("all_micro")
```

src/classification/models/train.py
```diff
@@ -314,6 +325,55 @@ def clean_checkpoints(self) -> None:
...
+    def compute_metrics(self, eval_pred: EvalPrediction) -> dict[str, float]:
+        """Compute per-class and overall metrics, and store predictions for confusion matrix plotting."""
+        logits, labels = eval_pred
+        predictions = np.argmax(logits, axis=-1)
+        self._preds = predictions
+        self._labels = labels
+        pred_classes = [self._id2class_enum[p] for p in predictions]
+        label_classes = [self._id2class_enum[lbl] for lbl in labels]
+        per_class_results = per_class_metric(pred_classes, label_classes)
+        class_metrics = {cls: per_class_results.get(cls, Metrics()) for cls in self._id2class_enum.values()}
```
christianabbet: I think you can even simplify this with
```suggestion
        per_class_results = per_class_metric(predictions, labels)
        class_metrics = {cls: per_class_results.get(id_cls, Metrics()) for id_cls, cls in self._id2class_enum.items()}
```

src/classification/models/train.py
```diff
@@ -314,6 +325,55 @@ def clean_checkpoints(self) -> None:
...
+        self._out_directory = out_directory
+        self._preds = None
+        self._labels = None
+
+    def compute_metrics(self, eval_pred: EvalPrediction) -> dict[str, float]:
+        """Compute per-class and overall metrics, and store predictions for confusion matrix plotting."""
+        logits, labels = eval_pred
+        predictions = np.argmax(logits, axis=-1)
+        self._preds = predictions
+        self._labels = labels
```
christianabbet: As mentioned, I think we can just save the computed cm
```suggestion
        self._cm = confusion_matrix(labels, predictions, labels=sorted_ids)
```

src/classification/utils/plots.py
```diff
@@ -0,0 +1,81 @@
...
+                str(cm_plot[i, j]),
+                ha="center",
+                va="center",
+                color="white" if cm_plot[i, j] > thresh else "black",
+                fontsize=max(6, 10 - n_plot // 5),
+            )
+
+    fig.tight_layout()
+    fig.savefig(png_path, dpi=150, bbox_inches="tight")
+    plt.close(fig)
```
christianabbet, line 53: I was wondering whether we could use the plotting from Sklearn as we already have the dependency (see [doc](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.ConfusionMatrixDisplay.html#sklearn.metrics.ConfusionMatrixDisplay.from_predictions)) ? Maybe it's easier for the future to maintain.

```
disp = ConfusionMatrixDisplay(confusion_matrix=cm, display_labels=classes_)
```

christianabbet (changes requested):

src/classification/models/train.py
AgathaSchmidt: @christianabbet  
After the discussion in the Sprint planning, do we still think to keep only f1? 

Or should we log
- log f1
- precision
- recall 
- one overall micro score/ accuracy?

src/classification/utils/plots.py
AgathaSchmidt: yes, great idea

src/classification/models/train.py
```diff
@@ -314,6 +320,40 @@ def clean_checkpoints(self) -> None:
...
+        return overall_metrics | per_class_metrics
+
+    def on_predict(self, args, state, control, metrics, **kwargs):
+        """After predictions are made, compute and save the confusion matrix."""
+        csv_path, png_path = plot_confusion_matrix(
+            self._cm,
+            self._out_directory,
+            split="test",
+            all_classes=list(self._id2class_enum.values()),
+        )
```
christianabbet, line 350:
```suggestion
        csv_path, png_path = plot_confusion_matrix(
            self._cm,
            args.output_dir,
            split="test",
            all_classes=list(self._id2class_enum.values()),
        )
```

src/classification/models/train.py
```diff
@@ -314,6 +320,40 @@ def clean_checkpoints(self) -> None:
                 shutil.rmtree(folder)
 
 
+class ConfusionMatrixCallback(TrainerCallback):
+    """Trainer callback to compute and save confusion matrix after evaluation."""
+
+    def __init__(self, id2class_enum: dict, out_directory: Path):
+        self._id2class_enum = id2class_enum
+        self._out_directory = out_directory
```
christianabbet:
```suggestion

```

src/classification/models/train.py
```diff
@@ -314,6 +320,40 @@ def clean_checkpoints(self) -> None:
                 shutil.rmtree(folder)
 
 
+class ConfusionMatrixCallback(TrainerCallback):
+    """Trainer callback to compute and save confusion matrix after evaluation."""
+
+    def __init__(self, id2class_enum: dict, out_directory: Path):
```
christianabbet:
```suggestion
    def __init__(self, id2class_enum: dict):
```

src/classification/models/train.py
```diff
@@ -336,45 +376,29 @@ def setup_trainer(
...
-    use_class_balacing = model_config.get("use_class_balancing", "false").lower() == "true"
+    use_class_balancing = model_config.get("use_class_balancing", "false").lower() == "true"
     compute_loss_func = None
-    if use_class_balacing:
+    if use_class_balancing:
         class_weights = compute_trainset_weights(train_dataset)
         # create the object that will be called to compute the loss function (standard in transformers lib).
         compute_loss_func = WeightedLabelSmoother(class_weights=class_weights)
 
+    cm_callback = ConfusionMatrixCallback(id2class_enum=bert_model.id2classEnum, out_directory=out_directory)
```
christianabbet:
```suggestion
    cm_callback = ConfusionMatrixCallback(id2class_enum=bert_model.id2classEnum)
```

src/classification/utils/plots.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # Filter to classes with at least one true or predicted sample for a readable plot.
+    active = (cm.sum(axis=1) > 0) | (cm.sum(axis=0) > 0)
+    active_idx = np.where(active)[0]
+    cm_plot = cm[np.ix_(active_idx, active_idx)]
+    names_plot = [class_names[i] for i in active_idx]
+
+    png_path = out_directory / f"confusion_matrix_{split}.png"
+
+    disp = ConfusionMatrixDisplay(confusion_matrix=cm_plot, display_labels=names_plot)
```
christianabbet: Small, comment would that work ? Is it maybe "easier to follow" ?
```suggestion
    active = (cm.sum(axis=1) > 0) | (cm.sum(axis=0) > 0)

    png_path = out_directory / f"confusion_matrix_{split}.png"
    disp = ConfusionMatrixDisplay(confusion_matrix=cm[:, active][active], display_labels=np.array(name_plots)[active])
```

christianabbet (approved):
LGTM. I already approve PR. Small comments before closing issue: 
 * `out_directory` that can be removed
 * potential simplification in plot_confusion_matrix.

src/classification/models/train.py
christianabbet: Good point, let's keep micro for now and if needed we will add f1 macro

PR #484, opened 2026-05-27
Fix add checkpoint resume support to BERT training

christianabbet:
## Summary
- Enables resuming training from an existing checkpoint by reusing the checkpoint's parent directory as `work_directory` and passing it to `trainer.train(resume_from_checkpoint=...)`
- Saves the model head before training starts and cleans checkpoint folders after training
- Removes the `--backbone-path` CLI option and split-model loading logic
- Fixes the dataset filename in `bert_config_color.yml`

### DIFF
--- a/src/classification/models/train.py
+++ b/src/classification/models/train.py
@@ -117,7 +117,7 @@ def common_options(f):
         "--model-checkpoint",
         type=click.Path(exists=True, path_type=Path),
         default=None,
-        help="Path to a local folder containing an existing bert model (e.g. models/your_model_folder).",
+        help="Path to a local folder containing an existing bert model (e.g. models/../checkpoint-xx).",
     )(f)
     f = click.option(
         "-o",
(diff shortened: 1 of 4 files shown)


src/classification/models/train.py
```diff
@@ -295,34 +291,27 @@ def compute_trainset_weights(
...
+        out_dir = Path(self.args.output_dir)
         out_dir.mkdir(parents=True, exist_ok=True)
 
         # Gather only trained layers (gradient is available)
         fine_tuned_names = {n for n, p in self.model.named_parameters() if p.requires_grad}
         head_state = {k: v.cpu() for k, v in self.model.state_dict().items() if k in fine_tuned_names}
 
         # Save trained layer and model config
-        save_file(head_state, out_dir / "model.safetensors")
+        save_file(head_state, out_dir / "head.safetensors")
```
AgathaSchmidt: I think we might have to adjust some code on inference side now that we changes the name  
I will check

AgathaSchmidt (commented):
We need to check if this works fine at inference level as well 
Otherwise code looks good and works as expected

src/classification/models/train.py
christianabbet: We can keep the old name then my bad

AgathaSchmidt: I think this is easier 
Otherwise there were some problems because we would have a different name when loading the full odel etc

AgathaSchmidt (approved):
Inference works now 
(We just have to not forget to specify tokenizer flag "-f" when using the split model)

Issue #411, opened 2026-03-10
Classification of organic components (unconsolidated rock)


stijnvermeeren-swisstopo:
For example, from _"Siltiger Feinsand mit **wenig Pflanzenresten (Wurzeln)**, schwach tonig, braun, bindig"_ we want the value `"roots"`. Multiple values for one layer are possible.

There are 11 values for organic component that are currently available in the Boreholes database schema: [organic_components_codelist.csv](https://github.com/user-attachments/files/25872092/organic_components_codelist.csv)

This classification is Prio 1 for unconsolidated rock.

There is a good amount of data available in the Boreholes database: [organic_components_stats.csv](https://github.com/user-attachments/files/25872098/organic_components_stats.csv)

Most of the data comes from the "swisstopo" workgroup (primarily from GeoQuat boreholes). Within this workgroup, we have...
- 61840x `"not specified"`
- 3366x `"humus"`
- 2347x `"undifferentiated organic material"`
- 871x `"earth"`
- 685x `"remains of woods"`
- 517x `"roots"`
- 428x `"remains of plants"`
- etc...

```sql
WITH data AS (
SELECT
  id_bho,
  name_wgp,
  lithological_description.depth_from,
  (SELECT text_cli_en from bdms.codelist WHERE codelist.id_cli = lithology_description_component_uncon_organic_codelist.component_uncon_organic_id) as property,
  (SELECT STRING_AGG(identifier_value, ',') FROM bdms.borehole_identifiers_codelist WHERE identifier_id = 100000006 AND borehole_id = id_bho) as geoquat_id
FROM bdms.borehole
JOIN bdms.stratigraphy ON stratigraphy.borehole_id = borehole.id_bho
LEFT JOIN bdms.lithological_description ON
  stratigraphy.id = lithological_description.stratigraphy_id
LEFT JOIN bdms.lithology ON
  lithological_description.depth_from = lithology.depth_from AND 
  lithological_description.depth_to = lithology.depth_to AND
  lithology.stratigraphy_id = stratigraphy.id 
LEFT JOIN bdms.lithology_description ON
  lithology_description.lithology_id = lithology.id
JOIN bdms.lithology_description_component_uncon_organic_codelist ON lithology_description_id = lithology_description.id
JOIN bdms.workgroups ON id_wgp_fk = id_wgp
)

[...]

PR #487, opened 2026-06-02
Feat/issue 485/support multilabel ground truth

### LINKED ISSUE #485: Support multilabel ground truth
Currently, each sample is assigned a single ground truth label. 

# Proposed change

Extend the data pipeline, training, and evaluation to support multilabel ground truth, where a single sample can be associated with more than one class simultaneously.

# Scope

* Update reduce_label (and callers) to return a list of labels instead of a single string
* Adapt the dataset/collation logic to produce multilabel targets (e.g. multi-hot vectors)
* Switch the loss function to one compatible with multilabel classification (e.g. BCEWithLogitsLoss)

# Example data 
#411  Organic components (unconsolidated rock) can habe multiple values, e.g. 

```
"organic_components": [
                            "humus",
                            "roots"
                        ],
```

AgathaSchmidt:
Closes #485 (support multi labels)

Using the classification of organic components of unconsolidated rock as well as the unclassifies coarse components (debris) as an example dataset multi-label classification support was implemented. 

#411 classification of organic components (unconsolidated rock)
#412 classification of unclassifiable coarse components (debris) unconsolidated rock)

## Changes 
- introduce `OrganicComponentsUnconsolidatedSystem` 
- adapt tokenization by preprocessing the labels to create a one hot encoded list of labels 
- adapt `compute_metrics` to handle multi-label per class metrics 
- allow `ground_truth_class` to be a list and adapt `reduce_label` to support multi-label
- create a special confusion matrix for multi class 
--> common approaches like `multilabel_confusion_matrix` are not representing what I was looking for **(feel free to discuss with me if you disagree!)**
--> I thought that we are mostly interested in wether we are able to predict the correct class, independantly from the sample 
--> the count of predictions represented in the cm does not have to equal the number of samples 
--> if a sample has two labels we count every one of them as single entry for the confusion matrix 
  ```
      cm = np.zeros((n_labels, n_labels), dtype=int)
      for true_row, pred_row in zip(labels, predictions, strict=False):
          true_labels = np.where(true_row == 1)[0]
          pred_labels = np.where(pred_row == 1)[0]
          for t in true_labels:
              for p in pred_labels:
                  cm[t, p] += 1


--> this way we focus on the detailled confusions of each label 

## Sources 
https://huggingface.co/blog/Valerii-Knowledgator/multi-label-classification


src/classification/utils/datasets/__init__.py
```diff
@@ -22,11 +23,14 @@ class ExistingClassificationSystems(Enum):
...
     color_consolidated = ColorConsolidatedSystem
     color_unconsolidated = ColorUnconsolidatedSystem
+    organic_components = OrganicComponentsUnconsolidatedSystem
 
     @classmethod
     def get_classification_system_type(
         cls,
-        class_system: Literal["uscs", "lithology", "en_main", "color", "color_consolidated", "color_unconsolidated"],
+        class_system: Literal[
+            "uscs", "lithology", "en_main", "color", "organic_components", "color_consolidated", "color_unconsolidated"
```
christianabbet: Now that we have multiple tasks, let's order them by name

src/classification/utils/datasets/classification.py
christianabbet: We should update the typing for this -> list[ClassificationSystem.EnumMember] | None

src/classification/models/model.py
```diff
@@ -264,20 +264,35 @@ def unfreeze_all_layers(self):
...
 
-        Args:
-            layers (list[LayerInformation]): A list of layers.
+        def encode_label(gt):
+            """Encodes the ground truth class into its corresponding integer value."""
+            if gt is None:
+                return None
+            if isinstance(gt, list):
+                return [cls.value for cls in gt]
+            return gt.value
```
christianabbet: If we only keep the list then we don't need this function. We can do the processing in  `preprocess_entry`

src/classification/models/model.py
```diff
@@ -264,20 +264,35 @@ def unfreeze_all_layers(self):
...
+        result = self.tokenizer(entry["layer"], truncation=True, padding="max_length", max_length=512)
+        label = entry["label"]
+        if isinstance(label, list):  # if its a list then its a multi-label classification problem
+            labels = [0.0] * self.num_class
+            for class_id in label:
+                labels[class_id] = 1.0  # multi-label binary encoding (1.0 for present classes, 0.0 for absent classes)
+            result["labels"] = labels
+        elif label is not None:
+            result["labels"] = label
+        return result
```
christianabbet, line 283: Same logic here, only consider case where we ave a list as input (isinstance(label, list)).

src/classification/utils/datasets/classification.py
```diff
@@ -177,9 +177,8 @@ def reduce_label(
         if label_str is None:
             return None
 
-        # Multilabel is not supported yet, reduce to first class
         if isinstance(label_str, list):
-            label_str = label_str[0]
+            return [cls.map_most_similar_class(s) for s in label_str]
 
         return cls.map_most_similar_class(label_str)
```
christianabbet: I would consider it as a list as well to make processing later easier 

```suggestion

        return [cls.map_most_similar_class(label_str)]
```

src/classification/utils/datasets/classification.py
```diff
@@ -123,7 +123,7 @@ class LayerInformation:
     language: str
     material_description: str
     class_system: type[ClassificationSystem]
-    ground_truth_class: None | ClassificationSystem.EnumMember
+    ground_truth_class: None | ClassificationSystem.EnumMember | list[ClassificationSystem.EnumMember]
```
christianabbet:
```suggestion
    ground_truth_class: None | list[ClassificationSystem.EnumMember]
```

src/classification/utils/datasets/organic_components.py
```diff
@@ -0,0 +1,61 @@
...
+
+    @classmethod
+    def normalize_class_string(cls, class_str: str) -> str:
+        """Normalize an organic components class string.
+
+        Args:
+            class_str (str): The class string to be normalized (e.g. "remains of wood").
+
+        Returns:
+            str: The normalized color class string (e.g., "remains_of_wood").
```
christianabbet: copy paste

src/classification/models/train.py
```diff
@@ -379,37 +390,118 @@ def clean_checkpoints(self) -> None:
...
+            # compute per-class metrics and micro-average, since sklearn doesn't handle multi-label well
+            metric_list = [
+                Metrics(
+                    tp=int(((predictions[:, i] == 1) & (labels[:, i] == 1)).sum()),
+                    fp=int(((predictions[:, i] == 1) & (labels[:, i] == 0)).sum()),
+                    fn=int(((predictions[:, i] == 0) & (labels[:, i] == 1)).sum()),
+                )
+                for i in range(predictions.shape[1])
+            ]
+            self._cm = multilabel_confusion_matrix_nxn(labels, predictions, n_labels)
```
christianabbet: Can't we use `multilabel_confusion_matrix` from scikit-learn ?

src/classification/models/train.py
```diff
@@ -379,37 +390,118 @@ def clean_checkpoints(self) -> None:
...
-
-        self._cm = confusion_matrix(labels, predictions, labels=self._sorted_ids)
-        per_class_results = per_class_metric(predictions, labels)
-        class_metrics = {cls: per_class_results.get(id_cls, Metrics()) for id_cls, cls in self._id2class_enum.items()}
-        per_class_metrics = {f"{cls.name}_f1": metric.f1 for cls, metric in class_metrics.items()}
-        overall_metrics = Metrics.micro_average(class_metrics.values()).to_dict("all_micro")
-        return overall_metrics | per_class_metrics
+        n_labels = len(self._id2class_enum)
+        if labels.ndim == 2:  # multi-label
+            predictions = (1 / (1 + np.exp(-logits)) > 0.5).astype(int)
```
christianabbet: Why not only have "predictions = logits > 0" as this is basically the same boundary (1 / (1 + exp(0))) ? 

We still need to keep track of the task classification vs multi-label to know if we just argmax or check per-label values

For the first version it makes make sense like that. But in the future I think it could be nice to keep in mind this logits aspects. Like having a per class logits threshold

src/classification/models/train.py
```diff
@@ -379,37 +390,118 @@ def clean_checkpoints(self) -> None:
...
+            labels = labels.astype(int)
+            # compute per-class metrics and micro-average, since sklearn doesn't handle multi-label well
+            metric_list = [
+                Metrics(
+                    tp=int(((predictions[:, i] == 1) & (labels[:, i] == 1)).sum()),
+                    fp=int(((predictions[:, i] == 1) & (labels[:, i] == 0)).sum()),
+                    fn=int(((predictions[:, i] == 0) & (labels[:, i] == 1)).sum()),
+                )
+                for i in range(predictions.shape[1])
```
christianabbet: Can0t we just use _cm in both cases to compute `self._per_class_metrics` so we have a single function

src/classification/models/train.py
```diff
@@ -379,37 +390,118 @@ def clean_checkpoints(self) -> None:
...
+        ] + [
+            row("macro_avg", macro["macro_precision"], macro["macro_recall"], macro["macro_f1"]),
+            row(
+                "micro_avg",
+                round(micro.precision, 4),
+                round(micro.recall, 4),
+                round(micro.f1, 4),
+                micro.tp,
+                micro.fp,
+                micro.fn,
```
christianabbet: I think you were right. There is no need to store all tp, fn , ..... etc. As it is already in the CM. The f1 is enought for each class. We can just dumpy it as a json. No need for a CSV writter ? I'm thinking that maybe it's just easier to log at every epoch the f1 for each class even though for color we will have 40 plots. At least it's cleaner no ?

christianabbet (commented):
According to the logic of the multilabel approach, I would transform all classes to multilabel. Color (and the others) are also multilabel task just where we have a single label. 

Also:
* Update docs/train_BERT.md: with new task

Check comments below

src/classification/models/train.py
AgathaSchmidt: I tried but I was not really happy with what it represents. I tried to explain my motivation in the PR description
Lets discuss soon!

src/classification/models/train.py
AgathaSchmidt: I am actually testing the code on a second dataset currently 
And there we have the problem (with my approach as well as with your approach) that no single logit is passing the threshold and we end up with all 0 predictions 
We indeed have to think about an other solution there

AgathaSchmidt: I tried your approach and added an argmax fallback so at least one label gets selected always 
Let me know what you think

src/classification/models/train.py
```diff
@@ -379,37 +389,85 @@ def clean_checkpoints(self) -> None:
...
+            no_prediction = predictions.sum(axis=-1) == 0
+            predictions[no_prediction, logits[no_prediction].argmax(axis=-1)] = 1
+            self._cm = multilabel_confusion_matrix_nxn(labels.astype(int), predictions, n_labels)
+        else:  # single-label
+            self._cm = confusion_matrix(labels, logits.argmax(axis=-1), labels=self._sorted_ids)
+        self._per_class_metrics = self._metrics_from_cm(self._cm)
+        metric_list = list(self._per_class_metrics.values())
+        return AllClassificationMetrics.compute_macro_average(
+            metric_list
+        ) | AllClassificationMetrics.compute_micro_average(metric_list)
```
christianabbet: Can't we just return the f1 here and then we don't even need to use  the `on_evaluate` method at all

```suggestion
        return (
            {f"{k}_f1": v.f1 for k, v in self._per_class_metrics.items()}
            | AllClassificationMetrics.compute_macro_average(metric_list)
            | AllClassificationMetrics.compute_micro_average(metric_list)
        )
```

src/classification/models/train.py
```diff
@@ -379,37 +389,85 @@ def clean_checkpoints(self) -> None:
...
     """Trainer callback to compute and save confusion matrix after evaluation."""
 
     def __init__(self, id2class_enum: dict):
         self._id2class_enum = id2class_enum
         self._sorted_ids = sorted(id2class_enum.keys(), key=lambda i: id2class_enum[i].value)
         self._cm: np.ndarray | None = None
+        self.current_test_name: str = "test"
+
+    def _metrics_from_cm(self, cm: np.ndarray) -> dict:
+        metric_list = [
```
christianabbet, line 422: Documentation

src/classification/models/train.py
```diff
@@ -379,37 +389,85 @@ def clean_checkpoints(self) -> None:
...
+def multilabel_confusion_matrix_nxn(labels, predictions, n_labels):
+    """Create nxn confusion matrix for multi-label classification.
+
+    We calculate (true, pred) pairs for each sample and increment the corresponding cell in the confusion matrix.
+    One sample can contribute to multiple cells in the confusion matrix.
+
+    Args:
+        labels: binary indicator matrices for labels of shape (n_samples, n_labels)
+        predictions: binary indicator matrices for predictions  of shape (n_samples, n_labels)
+        n_labels: total number of labels
```
christianabbet: Typing missing in doc and function

src/classification/models/train.py
```diff
@@ -333,7 +335,15 @@ def compute_trainset_weights(
...
     """
-    label_counts = Counter(trainset["label"])
+    labels = trainset["labels"]
+    if labels and isinstance(labels[0], list):
+        label_counts: Counter = Counter()
+        for row in labels:
+            for idx, val in enumerate(row):
+                if val > 0:
+                    label_counts[idx] += 1
+    else:
```
christianabbet, line 341:
```
        label_counts = Counter(idx for row in labels for idx, val in enumerate(row) if val > 0)
```

src/classification/models/train.py
```diff
@@ -379,37 +389,85 @@ def clean_checkpoints(self) -> None:
...
+    """
+    cm = np.zeros((n_labels, n_labels), dtype=int)
+    for true_row, pred_row in zip(labels, predictions, strict=False):
+        true_labels = np.where(true_row == 1)[0]
+        pred_labels = np.where(pred_row == 1)[0]
+
+        for t in true_labels:
+            for p in pred_labels:
+                cm[t, p] += 1
+    return cm
```
christianabbet: Can we just do

```python
return (labels.T @ predictions).astype(int)
```

christianabbet (changes requested):

christianabbet (approved):
I already accept the PR

Regarding the logits,  I would say that for now let's accept that we have a full 0 prediction if no classes are detected. So if that works, i would just remove for now:

```
no_prediction = predictions.sum(axis=-1) == 0
predictions[no_prediction, logits[no_prediction].argmax(axis=-1)] = 1
```

Then we can open a proper issue for the training that fails due to that

src/classification/models/train.py
```diff
@@ -379,37 +385,73 @@ def clean_checkpoints(self) -> None:
...
-        per_class_results = per_class_metric(predictions, labels)
-        class_metrics = {cls: per_class_results.get(id_cls, Metrics()) for id_cls, cls in self._id2class_enum.items()}
-        per_class_metrics = {f"{cls.name}_f1": metric.f1 for cls, metric in class_metrics.items()}
-        overall_metrics = Metrics.micro_average(class_metrics.values()).to_dict("all_micro")
-        return overall_metrics | per_class_metrics
+        if labels.ndim == 2:  # multi-label
+            predictions = (logits > 0).astype(int)
+            no_prediction = predictions.sum(axis=-1) == 0
+            predictions[no_prediction, logits[no_prediction].argmax(axis=-1)] = 1
+            self._cm = multilabel_confusion_matrix_nxn(labels.astype(int), predictions)
```
christianabbet: If we consider all datasets as lists, then we will always have "labels.ndim". One solution would be to add an additional property to the dataset like "single-label" or "multi-label" and pass it to the compute metrics. This would actually be useful for cases where we have an order in the component (like primary, secondary, ...)

christianabbet (changes requested):
Just realized this issue now

AgathaSchmidt:
> I already accept the PR
> 
> Regarding the logits, I would say that for now let's accept that we have a full 0 prediction if no classes are detected. So if that works, i would just remove for now:
> 
> ```
> no_prediction = predictions.sum(axis=-1) == 0
> predictions[no_prediction, logits[no_prediction].argmax(axis=-1)] = 1
> ```
> 
> Then we can open a proper issue for the training that fails due to that

If we have all 0 predictions we have to deal with errors during cm plot generation 
I think we can handle this, its no big deal 

Despite that, when looking at the tests I have done, those two lines lead to selecting class `not_specified` in almost all cases 
So the argmax is not selecting random classes
Instead, we get a clear signal that not much learning has been going on, without having NaN metrics / broken plots 

What do you think @christianabbet ?

christianabbet (approved):
LGTM

Issue #415, opened 2026-03-10
Classification of cementation (consolidated rock)


stijnvermeeren-swisstopo:
For example, from _"Siltstein, **mässig zementiert**, grau"_ we want the value `"moderately cemented"`.

The values for cementation that are currently available in the Boreholes database schema are `"uncemented"`, `"weakly cemented"`, `"moderately cemented"`, `"well cemented"`, `"strongly cemented"`, `"other"` and `"not specified"`. [cementation_codelist.csv](https://github.com/user-attachments/files/25875116/cementation_codelist.csv)

This classification is Prio 1 for consolidated rock.

Unfortunately, there is rather little data available in the boreholes database: [cementation_stats.csv](https://github.com/user-attachments/files/25875113/cementation_stats.csv)

There are just a few boreholes in the workgroups "swisstopo" and "Tiefe Bohrungen" where we have labelled layers.

We have...
- 91x `"well cemented"`
- 15x `"weakly cemented"`
- 9x `"uncemented"`
- 5x `"strongly cemented"`
- 1x `"moderately cemented"`

```sql
WITH data AS (
SELECT
  id_bho,
  name_wgp,
  lithological_description.depth_from,
  (SELECT text_cli_en from bdms.codelist WHERE codelist.id_cli = cementation_id) as property,
  (SELECT STRING_AGG(identifier_value, ',') FROM bdms.borehole_identifiers_codelist WHERE identifier_id = 100000006 AND borehole_id = id_bho) as geoquat_id
FROM bdms.borehole 
JOIN bdms.stratigraphy ON stratigraphy.borehole_id = borehole.id_bho
LEFT JOIN bdms.lithological_description ON
  stratigraphy.id = lithological_description.stratigraphy_id
LEFT JOIN bdms.lithology ON
  lithological_description.depth_from = lithology.depth_from AND 
  lithological_description.depth_to = lithology.depth_to AND
  lithology.stratigraphy_id = stratigraphy.id 
LEFT JOIN bdms.lithology_description ON
  lithology_description.lithology_id = lithology.id
JOIN bdms.workgroups ON id_wgp_fk = id_wgp
WHERE NOT unconsolidated
)

SELECT name_wgp, property, COUNT(*) as layers, COUNT(DISTINCT id_bho) as boreholes, COUNT(geoquat_id) as geoquat_layers FROM data
WHERE property IS NOT NULL
GROUP BY name_wgp, property
[...]

Issue #414, opened 2026-03-10
Classification of alteration degree from lithological descriptions


stijnvermeeren-swisstopo:
For example, from _"blaugrauer Ton (**verwitterter** Mergel)"_ we want the value `"moderately weathered"`.

The alteration degrees that are currently available in the Boreholes database schema are `"fresh"`, `"weathered"`, `"slightly weathered"`, `"moderately weathered"`, `"highly weathered"`, `"completely weathered"`, `"other"` and `"not specified"`.

**Unconsolidated rock**

This classification is Prio 1 for unconsolidated rock.

Unfortunately, there is very little data available in the boreholes database: [alteration_degree_uncon_stats.csv](https://github.com/user-attachments/files/25873596/alteration_degree_uncon_stats.csv)

We have...

- 65x `"moderately weathered"`
- 11x `"highly weathered"`
- 1x `"weathered"`
- 1x `"fresh"`
- 1x `"slightly weathered"`

**Consolidated rock**

This classification is Prio 2 for consolidated rock.

There is much more labelled data available in the boreholes database compared to unconsolidated rock: [alteration_degree_con_stats.csv](https://github.com/user-attachments/files/25873646/alteration_degree_con_stats.csv)

However, the data is pretty skewed, with most values being either `"moderately weathered"` or `"fresh"`. Additionally, it seems like the value in the DB often cannot be directly derived from the description text.

Most of the data comes from the "swisstopo" workgroup (all from GeoQuat boreholes). Within this workgroup, we have...

- 1643x `"moderately weathered"`
- 1112x `"fresh"`
- 62x `"highly weathered"`
- 57x `"slightly weathered"`
- 2x `"weathered"`

```sql
WITH data AS (
SELECT
  id_bho,
  name_wgp,
  lithological_description.depth_from,
  (SELECT text_cli_en from bdms.codelist WHERE codelist.id_cli = alteration_degree_id) as property,
  (SELECT STRING_AGG(identifier_value, ',') FROM bdms.borehole_identifiers_codelist WHERE identifier_id = 100000006 AND borehole_id = id_bho) as geoquat_id
FROM bdms.borehole 
JOIN bdms.stratigraphy ON stratigraphy.borehole_id = borehole.id_bho
LEFT JOIN bdms.lithological_description ON
  stratigraphy.id = lithological_description.stratigraphy_id
LEFT JOIN bdms.lithology ON
[...]

Issue #413, opened 2026-03-10
Classification of grain shape (unconsolidated rock)


stijnvermeeren-swisstopo:
For example, from _"Bergsturzmaterial, Gehängeschutt, grau mit grossen und kleinen Blöcken, Steine **plattig** und kantig, Lagerung mässig hart"_ we want the value `"platy"`. Multiple values for one layer are possible.

The grain shapes that are currently available in the Boreholes database schema are `"cubic"`, `"platy"`, `"elongated"`, `"other"` and `"not specified"`: [grain_shape_codelist.csv](https://github.com/user-attachments/files/25872988/grain_shape_codelist.csv)

This classification is Prio 1 for unconsolidated rock.

There is a good amount of data available in the Boreholes database, albeit very dominated by the value `"not specified"`: [grain_shape_stats.csv](https://github.com/user-attachments/files/25872993/grain_shape_stats.csv)

Most of the data comes from the "swisstopo" workgroup (all from GeoQuat boreholes). Within this workgroup, we have...

- 69442x `"not specified"`
- 231x `"platy"`
- 137x `"cubic"`
- 37x `"elongated"`

```sql
WITH data AS (
SELECT
  id_bho,
  name_wgp,
  lithological_description.depth_from,
  (SELECT text_cli_en from bdms.codelist WHERE codelist.id_cli = lithology_description_grain_shape_codelist.grain_shape_id) as property,
  (SELECT STRING_AGG(identifier_value, ',') FROM bdms.borehole_identifiers_codelist WHERE identifier_id = 100000006 AND borehole_id = id_bho) as geoquat_id
FROM bdms.borehole 
JOIN bdms.stratigraphy ON stratigraphy.borehole_id = borehole.id_bho
LEFT JOIN bdms.lithological_description ON
  stratigraphy.id = lithological_description.stratigraphy_id
LEFT JOIN bdms.lithology ON
  lithological_description.depth_from = lithology.depth_from AND 
  lithological_description.depth_to = lithology.depth_to AND
  lithology.stratigraphy_id = stratigraphy.id 
LEFT JOIN bdms.lithology_description ON
  lithology_description.lithology_id = lithology.id
JOIN bdms.lithology_description_grain_shape_codelist ON lithology_description_id = lithology_description.id
JOIN bdms.workgroups ON id_wgp_fk = id_wgp
)

[...]

Issue #409, opened 2026-03-10
Classification of mineral components (consolidated rock)


stijnvermeeren-swisstopo:
For example, from _"Calcaires bioclastiques péloidal, fins à moyens, légèrement oolithique, jaune à orangeâtre, finement gréseux, avec **pyrite** et **glauconie**"_ we want the values `"glauconite"` and `"pyrite"`.

There are 111 values for mineral components that are currently available in the Boreholes database schema: [mineral_components_codelist.csv](https://github.com/user-attachments/files/25870675/mineral_components_codelist.csv)

This classification is Prio 1 for consolidated rock.

Unfortunately, there is almost no labelled data available in the Boreholes database. [mineral_components_stats.csv](https://github.com/user-attachments/files/25870988/mineral_components_stats.csv)

Available data is limited to the boreholes "GEo-01" (boreholes.swissgeol.ch ID 36243) and "Niederlenz 1-1" (boreholes.swissgeol.ch ID 45035) and a few isolated values elsewhere.

```sql
WITH data AS (
SELECT
  id_bho,
  name_wgp,
  (SELECT text_cli_en from bdms.codelist WHERE codelist.id_cli = lithology_description_component_con_mineral_codelist.component_con_mineral_id) as property
FROM bdms.borehole 
JOIN bdms.stratigraphy ON stratigraphy.borehole_id = borehole.id_bho
LEFT JOIN bdms.lithological_description ON
  stratigraphy.id = lithological_description.stratigraphy_id
LEFT JOIN bdms.lithology ON
  lithological_description.depth_from = lithology.depth_from AND 
  lithological_description.depth_to = lithology.depth_to AND
  lithology.stratigraphy_id = stratigraphy.id 
LEFT JOIN bdms.lithology_description ON
  lithology_description.lithology_id = lithology.id
JOIN bdms.lithology_description_component_con_mineral_codelist ON lithology_description_id = lithology_description.id
JOIN bdms.workgroups ON id_wgp_fk = id_wgp
)

SELECT name_wgp, property, COUNT(*), COUNT(DISTINCT id_bho) FROM data
WHERE property IS NOT NULL
GROUP BY name_wgp, property
ORDER BY COUNT(*) DESC
```

PR #492, opened 2026-06-09
Dataset prio 1 (consolidated + unconsolidated)

### LINKED ISSUE #490: Prio 1 datasets (consolidated and unconsolidated)
Follow up of #427

Include all Prio 1 labels for both consolidated and unconsolidated rock (see https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/390, https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/391, https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/409, https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/410, https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/411, https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/412, https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/413, https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/414, https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/415) as datasets

christianabbet:
Closes #490 

# Data Overview

| Dataset               | Support (num classes) | Target | Issue | BERT* |
|-----------------------|-----------------------|--------|-------|-------|
| Accessory components  | 75 (66)	            | Multi  | #410  |   NO  |
| Alteration degree (c) | 2,716 (8)             | Single | #414  |  YES+ |
| Alteration degree (u) | 41 (8)                | Single | #414  |   NO+ |
| Cementation           | 78 (7)                | Single | #415  |   NO  |
| Color (c)             | 16,143 (91)           | Single | #390  |  YES+ |
| Color	(u)             | 20,121 (91)           | Single | #390  |  YES+ |
| Debris                | 70,084, (7)           | Multi  | #412  |  YES  |
| EN Main               | 89,842 (34)           | Single | -     |  YES  |
| Grain angularity      | 70,377 (8)            | Multi  | #391  |  YES  |
| Grain shape           | 70,087 (5)            | Multi  | #413  |  YES  |
| Lithology             | 45,323 (61)           | Single | -     |  YES  |
| Mineral components    | 63 (111)	            | Multi  | #409  |   NO  |
| Organic components    | 70,102 (11)           | Multi  | #411  |  YES  |
| USCS                  | 9,917 (38)            | Single | -     |  YES  |

> (*) Indicate if BERT training is possible considering amount of data.
(+) Training possible with BERT including consolidated and unconsolidated

Class ordering is now consistent: "not specified" is always first, followed by all dataset classes in alphabetical order.

## Lithology data leakage fix

Updated lithology class names to match the current ground truth: "tuffit" → "tuff", "ka" → "not specified", and added the "other" label. Also corrected "cataclastite" → "cataclasite" and "unconsolidated deposits" → "unconsolidated" in ground truth. 

The legacy ground truth file in data/lithology_splits has been refactored to match the new structure. Test/val/train splits were audited and fixed to remove entries sourced from the same document (data leakage).


docs/train_BERT.md
```diff
@@ -11,11 +11,21 @@ Each file contains boreholes, and each borehole contains layers. For each layer:
 
 | Class | Config string | JSON layer tag |
 |-------|--------------|----------------|
+| `AccessoryComponentsSystem` | `accessory_components` | `consolidated.accessory_components` |
```
AgathaSchmidt, line 14: when does the ClassificatioSystem name include "(un)consolidated" and when not? 
For sure it  has to be there when we have both like in Color and AlterationDegree but for the other ones sometimes its there, e.g. "DebrisUnconsolidatedSystem" , "OrganicComponentsUnconsolidatedSystem" . but not fall all unconsolidated e.g. "GrainAngularitySystem" and for no with consolidated

src/classification/config/bert/bert_config_alteration_degree_a2a.yml
```diff
@@ -0,0 +1,63 @@
...
+  global_consolidated:
+    classification_system: alteration_degree_consolidated
+    ground_truths:
+      - geoquat_ground_truth.json
+      - nagra_ground_truth.json
+  geoquat_consolidated:
+    classification_system: alteration_degree_consolidated
+    ground_truths:
+      - geoquat_ground_truth.json
+  nagra_consolidated:
```
AgathaSchmidt, line 31: do you think it will be a problem/error, that nagra has no data for consolidated ? --> empty test set in this case
Only unconsolidated

src/classification/models/train.py
```diff
@@ -391,30 +414,57 @@ def __init__(self, id2class_enum: dict):
...
         logits, labels = eval_pred
-        predictions = np.argmax(logits, axis=-1)
-
-        self._cm = confusion_matrix(labels, predictions, labels=self._sorted_ids)
-        per_class_results = per_class_metric(predictions, labels)
-        class_metrics = {cls: per_class_results.get(id_cls, Metrics()) for id_cls, cls in self._id2class_enum.items()}
-        per_class_metrics = {f"{cls.name}_f1": metric.f1 for cls, metric in class_metrics.items()}
-        overall_metrics = Metrics.micro_average(class_metrics.values()).to_dict("all_micro")
-        return overall_metrics | per_class_metrics
+        if labels.ndim == 2:  # multi-label
```
AgathaSchmidt: maybe we should push the multi class PR to develop first and then you pull the changes an resolve potential conflicts

src/classification/utils/datasets/grain_shape.py
```diff
@@ -0,0 +1,55 @@
...
+
+class GrainShapeSystem(ClassificationSystem):
+    """Classification system for grain shape of unconsolidated geological layers."""
+
+    @classmethod
+    def normalize_class_string(cls, class_str: str) -> str:
+        """Normalize a grain shape class string.
+
+        Args:
+            class_str (str): The class string to be normalized (e.g. "platy").
```
AgathaSchmidt: use an example which is actually normalized , e.g. "not specified"

src/classification/utils/datasets/grain_shape.py
```diff
@@ -0,0 +1,55 @@
...
+
+        Returns:
+            str: The normalized grain shape class string.
```
AgathaSchmidt: add normalized example "not_specified"

src/classification/utils/datasets/lithology.py
```diff
@@ -17,23 +17,27 @@ class LithologySystem(ClassificationSystem):
...
     def normalize_class_string(cls, class_str: str) -> str:
         """Normalize a lithology class string.
 
-        Examples include test that are comma separated (e.g. "limestone, bioclasts")
-            or semicolon ("limestone: micritic")
+        Examples include texts that are semicolon/comma separated (e.g. "limestone: micritic" -> "limestone",
+        "sandstone, marly" -> "sandstone") or white space (e.g. "not specified" -> "not_specified")
 
         Args:
-            class_str (str): The class string to be normalized (e.g. " "limestone, bioclasts",").
```
AgathaSchmidt, line 28: why did you specifically remove the example? 
We have them for every other task
Also, this one is more complicated than the other ones, so an example could be especially useful in my opinion

src/classification/utils/datasets/mineral_components.py
```diff
@@ -0,0 +1,161 @@
...
+
+class MineralComponentsSystem(ClassificationSystem):
+    """Classification system for mineral components of consolidated geological layers."""
+
+    @classmethod
+    def normalize_class_string(cls, class_str: str) -> str:
+        """Normalize a mineral class string.
+
+        Args:
+            class_str (str): The class string to be normalized (e.g. "pyrite").
```
AgathaSchmidt: Maybe we should use an example where the normalization is actually doing something, e.g. "not specified" --> "not_specified"

src/classification/utils/datasets/uscs.py
```diff
@@ -59,42 +59,41 @@ class USCSClasses(IntEnum):
             at 0, so using 0-based indexing avoids the need to address the issue later and prevents potential bugs.
         """
 
-        kunst = 0
+        kA = 0
```
AgathaSchmidt: In lithology we changed "kA" to "not specified" 
Should we do the same thing here?

src/classification/utils/datasets/classification.py
AgathaSchmidt: due to the change in 182 where `ground_truth_class` was changed to `None | list[ClassificationSystem.EnumMember] ` 

we need to change this line as well 

 ```
           "ground_truth_class": [c.name for c in self.ground_truth_class]
            if self.ground_truth_class is not None
            else None,
```

src/classification/utils/data_utils.py
AgathaSchmidt, line 134: Also, because `ground_truth_class` is a list now
`get_first_key_class=lambda layer: layer.ground_truth_class[0] if layer.ground_truth_class else None,`

AgathaSchmidt, line 149: `get_second_key_class=lambda layer: layer.ground_truth_class[0] if layer.ground_truth_class else None,`

AgathaSchmidt: `all_classes = set(predictions) | {cls for gt in ground_truths if gt for cls in gt}`

AgathaSchmidt: `num_ground_truth = sum(class_ in gt for gt in ground_truths if gt)`

AgathaSchmidt: `"ground_truth_class": [c.name for c in layer.ground_truth_class] if layer.ground_truth_class else None,`

AgathaSchmidt (changes requested):
I tested the bert training pipeline on `organic_components` (multi) and `cementation` (single) and it works fine. 

Mostly few consistency remarks.  

When running the classification pipeline I encountered several errors which came due to changing the ground_truth_class to a list 

This one has to be corrected as well: 


in `src/classififcation/evaluation/evaluate.py `

 change line 226 

`   labels = [layer.ground_truth_class for layer in layers]`

to 

` labels = [layer.ground_truth_class[0] if layer.ground_truth_class else None for layer in layers]
`

AgathaSchmidt:
Also, we need to change line 39 `in src/classification/classifiers/bert_classifier.py`

from 
`model_path = self.config["model_path"]`
to 
` model_path = Path(self.config["model_path"])`

docs/train_BERT.md
christianabbet: Good point I missed the two. I'll put conso/unconso only if both exists (color and alteration)

src/classification/config/bert/bert_config_alteration_degree_a2a.yml
christianabbet: I was not sure. I kept the full schema even if it was empty such that we have the same logic everytime. We just have the explicit message that it's empty. In the future if we augment the datasets then we don't need to modify it

src/classification/models/train.py
christianabbet: Agree! I'll wait for the other PR

src/classification/utils/datasets/lithology.py
christianabbet: I put the examples above. Do you think we should add more ?

AgathaSchmidt: Ah, sorry, no its fine like that! 
I was looking at the wrong place

AgathaSchmidt (approved):
Looks good
I tested classification, extraction and Bert training pipeline (single and multi label) , all worked fine

PR #494, opened 2026-06-15
Feat/issue 479/research run classification on remote cluster

### LINKED ISSUE #479: Research: Run classification on remote cluster
# Summary
Investigate how to offload classification tasks from the local machine to a remote cluster or distributed compute environment.

# Motivation
Running classification locally is resource-intensive and slow. Moving execution to a cluster would improve throughput and free up local resources.

AgathaSchmidt:
Closes #479 

Adapts the MLFlow logging structure so it aligns with Azure. Provides and example job.py for training a Bert model. 

## Changes 

- add `"azure-ai-ml"` and` "azure-identity"` to pyproject.toml  --> therefore `uv.lock` was updated as well 
- create `src/scripts/azure_jobs/job_train_bert.py` --> this is the job script to train a bert model 
- adjust MLFlow set up on  `train.py ` so it is compatible with Azure and keeps the current pipelines intact as well


src/scripts/azure_jobs/job_train_bert.py
```diff
@@ -0,0 +1,68 @@
...
+# Connect to your workspace
+ml_client = MLClient(
+    credential=DefaultAzureCredential(),
+    subscription_id="f12e214d-46c6-49bf-a083-f89cf9c3179d",
+    resource_group_name="rg-swisstopo-compute",
+    workspace_name="aml-swisstopo-sn",
+    # compute="swisstopo-boreholes",
+    compute="swisstopo-dev",
+)
+
```
christianabbet: Remove comment and set all load env variables from outside to avoid committing variable to repo

src/scripts/azure_jobs/job_train_bert.py
```diff
@@ -0,0 +1,68 @@
...
+        dockerfile_path="Dockerfile.azureml",
+    ),
+    description="Pre-built environment for BERT training (deep-learning + experiment-tracking extras)",
+)
+
+registered_env = ml_client.environments.create_or_update(env)
+
+job = command(
+    code=".",
+    command=job_command,
```
christianabbet: Passing the whole code is a bit too much. It would be nice to limit this to src folder using `.amlignore`.

src/classification/models/train.py
```diff
@@ -232,7 +240,7 @@ def setup_training_args(model_config: ExperimentHyperparameters, out_directory:
     Returns:
         TrainingArgument: the training arguments.
     """
-    report_to = "mlflow" if mlflow_tracking else "none"
+    report_to = "none"  # metrics logged via MetricsMLflowCallback to avoid Azure ML's 200-param limit
```
christianabbet: Put directly as argument in function below, no need for local var

Also what happens to `"mlflow" in args.report_to` ? The code is never reached ?

src/scripts/azure_jobs/job_train_bert.py
```diff
@@ -0,0 +1,68 @@
...
+    compute="swisstopo-dev",
+    environment=f"{registered_env.name}:{registered_env.version}",
+    experiment_name="test-bert-training-cementation",
+    inputs=job_inputs,
+    environment_variables=job_env_vars,
+)
+
+returned_job = ml_client.jobs.create_or_update(job)
+print(f"Job submitted: {returned_job.name}")
+print(f"Monitor at: {returned_job.studio_url}")
```
christianabbet: Can we log the model output to Azure Model list so we keep there the model version instead of within the run?

christianabbet (changes requested):
Let's make directly a proper script for training that we can use locally. Calling the script in a similar / same way as the main training. An idea would be to load by default all datasets and only pass the "-cf" argument for the training script

It would be nice documenting this new script in the BERT training file for the future

christianabbet (approved):

PR #496, opened 2026-06-16
Feat/issue 430/classification   ground truth data creation for description labels

AgathaSchmidt:
Closes #430 

- [x] accessory components 
- [x] mineral components 
- [x] cementation
- [x] alteration degree unconsolidated

### Changes
- crate full example lists with all available **classes**  (in contrast to the other example list, where only those clases are present, which actually occur in the datasset)
`src/classification/config/bedrock/full_examples`
- for those 4 respective config files change the path to the full examples and change max_tokens: 8192
- adjust BERT configs to use all enhanced datasets for training
- implement the `--predict-all` option in `src/classification/main.py` to classify every layer description regardless of existing labels and update the `src/classification/runner.py` respectively 
- Introduce a dedicated multi-label tool schema for the Bedrock classifier,

FIX TYPO FOR ACCESORY COMPONENTS RESULTS

### Main deliverable
Uploaded the four enhanced datasets to the AWS bucket (did not run for Thurgau since we had enough data already) 
- stijnvermeeren-boreholes-data/deepwells_ground_truth_enhanced.json
- stijnvermeeren-boreholes-data/zurich_ground_truth_enhanced.json
- stijnvermeeren-boreholes-data/nagra_ground_truth_enhanced.json
- stijnvermeeren-boreholes-data/geoquat_ground_truth_enhanced.json


src/classification/classifiers/aws_bedrock_classifier.py
```diff
@@ -102,11 +103,19 @@ def __init__(
...
 
-        self.tool = read_params(self.config["tool_file"])["reasoning" if self.reasoning_mode else "classification"]
+        if self.multilabel_mode and self.reasoning_mode:
+            tool_key = "multilabel_reasoning"
+        elif self.multilabel_mode:
+            tool_key = "multilabel"
+        elif self.reasoning_mode:
+            tool_key = "reasoning"
+        else:
+            tool_key = "classification"
```
christianabbet: It's more aesthetics but what do you think of this notation to make it "easier to read". Up to you to include it or not

```suggestion
        tool_key = {
            (True, True): "multilabel_reasoning",
            (True, False): "multilabel",
            (False, True): "reasoning",
            (False, False): "classification",
        }[(self.multilabel_mode, self.reasoning_mode)]
```

src/classification/classifiers/aws_bedrock_classifier.py
```diff
@@ -93,6 +93,7 @@ def __init__(
         self.pattern_version = self.config["pattern_version"]
         self.prompt_version = self.config["prompt_version"]
         self.reasoning_mode = self.config["reasoning_mode"]
+        self.multilabel_mode = self.config.get("multilabel_mode", False)
```
christianabbet: I cannot find any file where `multilabel_mode` is defined ? Do we need it ? If not then we can also remove the corresponding tool

src/classification/runner.py
```diff
@@ -86,12 +87,23 @@ def run_classification_predictions(
...
+            is_prediction = True
+        else:
+            layer_descriptions_gt = classification_system_cls.process(ground_truth=gt_boreholes)
+            _, _, layer_descriptions = split_samples(layer_descriptions_gt)
+
+            # No labeled data for this classification system — classify all descriptions without evaluation
+            if not layer_descriptions:
+                logger.info("No labeled data found for this classification system. Classifying all descriptions.")
+                layer_descriptions = classification_system_cls.process(ground_truth=gt_boreholes, allow_none=True)
+                is_prediction = True
```
christianabbet: Could we maybe encapsulate this in a functions ? Also more about readability. Something like that ?

```python
    layer_descriptions, is_prediction = _load_layer_descriptions(file_path, classification_system_cls, options)
```

```python
def _load_layer_descriptions(
    file_path: Path | None,
    classification_system_cls: type[ClassificationSystem],
    options: ClassificationOptions,
) -> tuple[list[LayerInformation], bool]:
    """Load layer descriptions either as predictions or as ground truth (test set).

    Returns:
        tuple[list[LayerInformation], bool]: The layer descriptions to classify, and whether
            they came from prediction data (True) or ground truth data (False).
    """
    ...
```

christianabbet (changes requested):
Two minor comments on code structure. One main on use of `self.multilabel_mode`

christianabbet (approved):
LGTM

PR #497, opened 2026-06-18
Distribute models through HugginFace

### LINKED ISSUE #473: Improve model storage and distribution
When we decided to push the classification models to the repo we were aiming to keep the same approach as we did in the assets repository. 
The main difference though: the model size. 
Since the assets model is a simple XGBoost model and the model in this repo is a fine tuned BERT model, simply pushing it to the repo was not possible. 
The best way to upload files that large to the repo is via LFS. 
While Git LFS technically supports files of this size, it is generally not recommended for model artifacts in a code repository, as it inflates repo size and complicates the development workflow. 
We therefore propose an alternative solution such as uploading the models to Hugging Face Hub. 

To be discussed.

AgathaSchmidt:
Closes #473 

## Changes

- Organizational HuggingFace account was opened (https://huggingface.co/swissgeol)
- En Main (https://huggingface.co/swissgeol/en_main) and Lithology (https://huggingface.co/swissgeol/lithology) model were uploaded (full models --> recoupled our fine tuned head with backbone)
--> lithology had to be retrained due to the new class structure introduced in #492 
- adjusted `model_path` typing and multiple descriptions to fit the new huggingface option
- HuggingFace's` from_pretrained() `natively accepts either a local directory path or a Hub repo ID string (e.g. "swissgeol/en_main"). That's built into the transformers library --> no extra code was needed.
- remove LSF artifacts and adjust Docs 
- adapt API instructions so it is not dependent on LFS anymore --> downloa dmodels from Huggingface --> split them --> upload head to Docker
- Docker building had to be changed --> new script `src/app/prepare_models.py` downloads modes from Huggingface splits them and provides heads for Docker


README.md
```diff
@@ -196,13 +196,9 @@ Repeat steps 1 and 2 of the [data extraction pipeline](#run-data-extraction) to
...
 
-- **From the repository (recommended):** Models are committed under `models/` via Git LFS. Pull them with:
-  ```bash
-  git lfs pull
-  ```
-  This provides `models/backbone/backbone.safetensors`, `models/en_main_head/`, and `models/lithology_head/`.
+- **From HuggingFace (recommended):** Published models are hosted on [HuggingFace](https://huggingface.co/swissgeol). No manual download is needed — pass the HuggingFace model ID (e.g. `swissgeol/lithology`) directly via the `-p` flag at runtime and the model is loaded automatically.
 
-- **From S3 (swisstopo internal):** If you have access to the swisstopo S3 bucket, you can download models directly:
+- **From S3 (swisstopo internal):** If you have access to the swisstopo S3 bucket, you can download models directly (e.g.For models not yet published on HuggingFace), and pass the local path via `-p`:
```
christianabbet, line 201: Shouldn't we also put in the documentation that we need to run the script locally to download the models (if we don't use the Dockerfile and classification then model/backbone does not exist...). Or just drop an error message when running API that user need to download model using ...

```python
python src/app/prepare_models.py
```

src/app/prepare_models.py
```diff
@@ -0,0 +1,73 @@
...
+
+It downloads each full fine-tuned model, extracts the frozen backbone (shared across all systems) and the task-specific
+head weights, and writes them to the directory layout expected by lithology_classification.py:
+
+    models/backbone/backbone.safetensors  — shared frozen BERT encoder weights
+    models/backbone/                      — tokenizer files
+    models/lithology_head/model.safetensors
+    models/lithology_head/config.json
+    models/en_main_head/model.safetensors
+    models/en_main_head/config.json
```
christianabbet: maybe we can already say something like that to avoid maintenance on the long term

```suggestion
    models/model_head_1/model.safetensors
    models/model_head_1/config.json
    ...
    models/model_head_n/model.safetensors
    models/model_head_n/config.json
```

src/app/prepare_models.py
```diff
@@ -0,0 +1,73 @@
...
+]
+
+BACKBONE_DIR = Path("models/backbone")
+
+# Parameter prefixes belonging to the task-specific head.
+# Mirrors unfreeze_layers: ["classifier", "pooler", "layer_11"] in the BERT config YAMLs.
+_HEAD_PREFIXES = ("classifier.", "bert.pooler.", "bert.encoder.layer.11.")
+
+
+def _is_head_param(name: str) -> bool:
```
christianabbet, line 40: oneliner comment

christianabbet (changes requested):
Great! Works like a charm! Minor comments

* How do we upload a model om Hugging face? Do we do it manually? If so, is it worth documenting something internally ?

AgathaSchmidt:
So far, I uploaded the models manually. 
Do you think we should create a script to upload? 
I would prefer to not have a completely automatic uploading after ever training. At least there should be a kind of upload_to_huggingface flag which can be set to False 
What do you think? Manually is okay, or enhance code for uploading flag?

README.md
AgathaSchmidt: I included this in the Dockerfile 
```
# Download models from HuggingFace, split into backbone + heads, then wipe the download cache
RUN /app/.venv/bin/python src/app/prepare_models.py && \
    rm -rf /root/.cache/huggingface
```

So that the users does not have to execute the script themselves, instead it is executed whenever the Docker Image is build

christianabbet:
As discussed during internal meeting. We decided to keep upload manual as developing solution would take more time than performing the action.

christianabbet (approved):

PR #517, opened 2026-07-09
Feat/issue 508/feature reduce docker image size

### LINKED ISSUE #508: Feature: Reduce Docker image size
## Description
After the merge of the branch introducing deep learning classification models, the Docker image size increased dramatically from a manageable size to ~8 GB pushed on GitHub Container Registry. The root cause is that `torch==2.8.0` resolves to its full CUDA build by default, pulling in ~4–5 GB of NVIDIA runtime libraries (`nvidia-cudnn-cu12`, `nvidia-cublas-cu12`, `nvidia-nccl-cu12`, `nvidia-cusparse-cu12`, `nvidia-cufft-cu12`, `triton`, etc.) that are entirely unused at inference time.

## Motivation
The production API container only runs BERT inference (sequence classification). It does not require GPU support. Carrying 4–5 GB of CUDA libraries into every image push and pull is wasteful and slows down CI/CD, cold starts, and deployments. GPU training is handled separately in Azure.

## Proposed Solution
- Add a `deep-learning-cpu` optional dependency group in `pyproject.toml` that mirrors `deep-learning` but routes `torch` through the PyTorch CPU-only wheel index (`https://download.pytorch.org/whl/cpu`) via `[tool.uv.sources]`.
- Update the `Dockerfile` to install `--extra deep-learning-cpu` instead of `--extra deep-learning`
- Keep the existing `deep-learning` extra unchanged for GPU training workloads in Azure
- Check if inference is not taking too much time.

Expected image size reduction: **~4–5 GB** (all `nvidia-*` packages and `triton` disappear from the production image).

AgathaSchmidt:
Closes #508 

### Requested changes
- deep-learning-cpu extra was added to pyproject.toml with the pytorch-cpu source
- Dockerfile was updated to --extra deep-learning-cpu

**What we found during investigation**: The nvidia packages were still present in the amd64 image ( amd64 (x86_64 Linux, the architecture used in CI and production)  Two problems:

Stale uv.lock — The lock file was never regenerated after adding the pytorch-cpu source. It still pinned torch from https://pypi.org/simple, which on x86_64 Linux pulls nvidia-cublas-cu12, nvidia-cudnn-cu12, triton, etc. as mandatory dependencies. The source configuration in pyproject.toml has no effect if the lock file overrides it.

Dockerfile didn't copy uv.lock — uv sync inside Docker resolved packages from scratch on every build (ignoring any lock file), so even a correct lock wouldn't have helped.

### Addtional changes
- Ran uv lock to regenerate the lock file — torch now resolves from https://download.pytorch.org/whl/cpu with zero nvidia dependencies
- Added uv.lock to the COPY line in the builder stage so Docker uses the pinned resolution deterministically

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ COPY pyproject.toml README.md /app/
 # --no-dev: exclude development dependencies
 # --no-install-project: skip building and installing the project package itself
 # --compile: generate *.pyc files to lower memory load at startup
-RUN uv sync --no-dev --no-install-project --compile --extra deep-learning
+RUN uv sync --no-dev --no-install-project --compile --extra deep-learning-cpu
 
 
 ## ------ Runtime stage
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ ENV SETUPTOOLS_SCM_PRETEND_VERSION=$VERSION
 
 # Setup working directory and copy pyproject files
 WORKDIR /app
-COPY pyproject.toml README.md /app/
+COPY pyproject.toml README.md uv.lock /app/
 
 # Install all dependencies including deep-learning extras into a virtualenv.
 # --no-dev: exclude development dependencies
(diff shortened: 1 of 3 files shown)


christianabbet (approved):
Looks good on my side. Tested locally and then remote (https://github.com/swisstopo/swissgeol-boreholes-dataextraction/actions/runs/29264612277/job/86866546103). Only 2GB (much less than before)

PR #518, opened 2026-07-09
Classify All API endpoint

### LINKED ISSUE #509: Feature: Unified classification endpoint(s) for all BERT tasks
## Description
Create a single endpoint that, given one input description, predicts **all** relevant classification tasks at once, replacing the current `/classify_lithology` endpoint.

The model computes a shared backbone embedding `Z` from the description once, then feeds `Z` independently into each task-specific classification head (lithology, en_main, en_secondary, color, uscs, etc.) to produce all outputs in a single pass. Based on whether the material is consolidated or unconsolidated  (inferred by the model itself from the description) only the subset of tasks relevant to that rock type is returned.

## Motivation
The primary use case is: *"here is a description, give me all the properties you can derive/classify from it."* Calling a per-task endpoint once per classification system is inefficient, since it recomputes the backbone embedding redundantly for every task and doesn't match how the endpoint is actually used. Computing the embedding once and reusing it across all task heads is significantly more efficient, and filtering to only relevant tasks avoids returning meaningless predictions.

## Proposed Solution
1. Design a single endpoint (e.g. `/classify`) that accepts a description and returns a mapping of `task_name → predicted_class` for all relevant tasks.
2. Model architecture:
   - Load the shared backbone once and sub models.
   - Compute embedding `Z` from the input description.
   - Feed `Z` independently into each task-specific classification head to produce all outputs in one forward pass.
3. Use the model's own prediction (e.g. a dedicated consolidated/unconsolidated head or existing task) to determine rock type, then filter the returned tasks to only those relevant to that category.
4. Update schemas in `src/app/common/schemas.py`: request needs only a description; response is a dict of task, containing only relevant tasks.
5. Update `src/app/api/v1/router.py` to replace the per-task endpoint with this unified multi-task endpoint.
6. Update documentation to describe the new unified, single-pass, multi-task endpoint.

### LINKED ISSUE #516: Classify consolidated/unconsolidated based on other classification outputs (lithology / EN norm)
## Description
Add a new classification task that predicts whether a given description is **consolidated** or **unconsolidated**. Training labels for this task are derived from the existing `lithology` and `en_main` ground truths, by defining a mapping from their labels to consolidated/unconsolidated. In addition, audit all other classification tasks/attributes to determine which ones are only valid for consolidated material, which are only valid for unconsolidated material, and which apply to both.

## Motivation
Issue #509 proposes a unified multi-task classification endpoint that, given a description, should return only the task outputs relevant to the material's category — e.g. it should not return consolidated-only attributes for a layer classified as unconsolidated, or vice versa. That requires (a) a way to predict consolidated/unconsolidated for a given description, and (b) a definition of which attributes are applicable to each category. This issue provides both, laying the groundwork for #509's filtering logic.

## Proposed Solution
1. Define a mapping from `lithology` and `en_main` classes to consolidated/unconsolidated labels.
2. Apply this mapping to existing data to generate consolidated/unconsolidated labels for training.
3. Train a new classification head/task for consolidated vs. unconsolidated, following the same backbone + head architecture as the other classification tasks.
4. Audit all existing classification tasks/attributes and determine, for each one, whether it applies to consolidated material, unconsolidated material, or both.

AgathaSchmidt:
Closes #509 
Closes #516

### Changes 
- `src/app/prepare_models.py` added all model paths 
- `src/app/api/v1/endpoints/classify_all.py` 
     - contains the endpoint for classifying all lithological attributes
     - differentiates `_UNCONSOLIDATED_TASKS` and `_CONSOLIDATED_TASKS`
     - loads all models 
     - classifies in two steps: lithology first and based on this all relevant tasks with respect to the fact wether the input 
     is consolidated or unconsolidated 
     - remove old lithology only endpoint `src/app/api/v1/endpoints/lithology_classification.py`
- `src/app/api/v1/router.py`, `src/app/main.py `and `src/app/common/schemas.py` : removed lithology endpoint and connected with classify all endpoint  
- `docs/API_and_Docker.md` describe new endpoint
- testing the classification endpoint in `tests/api/v1/test_classify_all.py`

Time advantage from the two step classification: 
Time per inference in ms : 
- examples full model : 1702 , 1977, 1622
- examples two step: 441, 425, 360


src/app/common/schemas.py
```diff
@@ -799,37 +798,63 @@ def from_prediction(
...
+                    "uscs": "SC",
+                    "debris": ["not_specified"],
+                    "color": "beige",
+                    "grain_angularity": ["angular"],
+                    "grain_shape": ["platy", "elongated"],
+                    "organic_components": ["undifferenciated_organic_material", "remains_of_wood"],
+                }
+            }
+        }
+    )
```
christianabbet: Can we add multiple examples ? If so, would it make sens to add have both (consolidated / unconsolidated) answer schemas ?

src/app/common/schemas.py
```diff
@@ -799,37 +798,63 @@ def from_prediction(
...
-    class_name: str = Field(
+    Consolidated rock tasks: lithology, alteration_degree_consolidated, cementation, color, mineral_components,
+        accessory_components.
+    Unconsolidated sediment tasks: en_main, uscs, debris, color, grain_angularity, grain_shape, organic_components.
+    """
+
+    predictions: dict[str, str | list[str]] = Field(
         ...,
-        description="Predicted class name for the input description.",
+        description=(
```
christianabbet: We should adapt the basemodel to also provide typing for all the elements. Ideally, we want the user to know what to expect from the results. For example, the prediction returns None if field not present, otherwise predicted field

```
....
uscs: USCSClasses  | None = None
grain_shape: list[GrainShapeClasses]  | None = None
....
```

src/app/api/v1/endpoints/classify_all.py
```diff
@@ -0,0 +1,121 @@
...
+        if task_name == "lithology":
+            predictions["lithology"] = lithology_class.name
+            continue
+        model = bert_models[task_name]
+        classes = model.predict_from_embedding(shared_hidden_states, extended_mask)
+        predictions[task_name] = (
+            [c.name for c in classes] if model.classification_system.is_multi_label() else classes[0].name
+        )
+
+    return ClassifyResponse(predictions=predictions)
```
christianabbet: Update model creation with actuall tasks (see comment on Pydantic model for prediction)

```
ClassifyResponse(**predictions)
```

src/app/api/v1/endpoints/classify_all.py
```diff
@@ -0,0 +1,121 @@
...
+from __future__ import annotations
+
+import logging
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from app.common.schemas import ClassifyRequest, ClassifyResponse
+
+if TYPE_CHECKING:
+    from classification.models.model import BertModel
```
christianabbet, line 13: Is this linked to circular imports ? if so, can we change the loading logic / file structure  to avoid using this ?

christianabbet (changes requested):

src/app/common/schemas.py
AgathaSchmidt: I added two examples, which can be selected in the drop down

src/app/api/v1/endpoints/classify_all.py
AgathaSchmidt: The TYPE_CHECKING is a guard which is neccessary for running the CI testsm where only the lighter dependencies are installed. 
Importing directly works locally, but crashes in CI. 
The guard delays the import so it only happens when BERT is actually used, not just whenever the file loads

AgathaSchmidt: This is the error I had in the pytest here on GitHub when i removed the TYPE_CHECKING and imported BertModel directly 

```
ImportError while loading conftest '/home/runner/work/swissgeol-boreholes-dataextraction/swissgeol-boreholes-dataextraction/tests/conftest.py'.
tests/conftest.py:10: in <module>
    from app.main import app
src/app/main.py:15: in <module>
    from app.api.v1.endpoints.classify_all import load_models
src/app/api/v1/endpoints/classify_all.py:9: in <module>
    from classification.models.model import BertModel
src/classification/models/model.py:7: in <module>
    import datasets
E   ModuleNotFoundError: No module named 'datasets'
Error: Process completed with exit code 4.
```

src/app/api/v1/endpoints/classify_all.py
```diff
@@ -0,0 +1,119 @@
...
+    for task_name in relevant_tasks:
+        if task_name not in bert_models:
+            logger.warning(f"Task '{task_name}' not loaded, skipping.")
+            continue
+        if task_name == "lithology":
+            predictions["lithology"] = lithology_class
+            continue
+        model = bert_models[task_name]
+        classes = model.predict_from_embedding(shared_hidden_states, extended_mask)
+        predictions[task_name] = classes if model.classification_system.is_multi_label() else classes[0]
```
christianabbet: By merging the other branch with ranking we made `is_multi_label` outdated, we just need to update it  (with `.classification_task()` == ClassificationTask.multi_label) I think. It does not exists on my branch

```python
        predictions[task_name] = (
            classes
            if model.classification_system.classification_task() == ClassificationTask.multi_label
            else classes[0]
        )
```

Also exists in model.py

src/app/common/schemas.py
```diff
@@ -799,37 +810,102 @@ def from_prediction(
...
+        accessory_components.
+    Unconsolidated sediment tasks: en_main, uscs, debris, color, grain_angularity, grain_shape, organic_components.
+    """
+
+    accessory_components: list[AccessoryComponentsSystem.AccessoryComponentsClasses] | None = Field(
+        default=None, description="Predicted accessory component classes (consolidated rock only)."
+    )
+    alteration_degree_consolidated: AlterationDegreeConsolidatedSystem.AlterationDegreeClasses | None = Field(
+        default=None, description="Predicted alteration degree class (consolidated rock only)."
+    )
```
christianabbet: I still don't get the list of classes. It should be "ENMainSystem.ENMainClasses" no ? This is mainly a concern for when we will provide it through assets

Image

christianabbet (changes requested):
Small comments

src/app/common/schemas.py
AgathaSchmidt: I found the source of the problem and a solution. 
Root cause: The classification classes (e.g. `ENMainClasses`) are `IntEnums` with 0-based indexing, required for training/inference (they map directly to the BERT model's output logit positions — see `src/classification/utils/datasets/`). But `IntEnum's` default serialization emits the integer` .value,` not a readable label.

To return human-readable predictions, `_serialize_single_label`/`_serialize_multi_label` override serialization to emit `.name i`nstead. The side effect: FastAPI derives the OpenAPI schema from the serialized type, so once the field serializes to a plain str, the enum's list of valid member names is no longer reflected in the generated schema/docs.

Fix: restore that list explicitly via `json_schema_extra` (see `_enum_names_schema_extra`), so the docs still show the valid class names even though the wire format is a plain string

christianabbet (commented):
The solution looks complex IMO but also I don't have any other obvious alternative to suggest. Would it be possible to replace the IntStr to EnumStr then? I guess this implies also modifying the way index are generated for each str entry. If this is too complex, we can leave it like this for now and open a ticket for the future

AgathaSchmidt:
It might be not very elegant but I don't think changing to StrEnum will be better. 
We define the Classes as `IntEnum` because HuggingFace's `AutoModelForSequenceClassification` needs int ids to size the classifier head. A string value breaks model construction. Therefore many methods depend on int ID labels e.g. `preprocess_entry`
I tried to simplify the logic a bit by centralizing multiple steps in one helper 
--> the serialization (so in the response schema we see the class name e.g. "grey" instead of its ID "5")
--> writing the json schema by using `.name` of each member of the Enum
What do you think @christianabbet ?

christianabbet (approved):
LGTM

Issue #512, opened 2026-07-01
More accurate extension of material description bounding boxes


stijnvermeeren-swisstopo:
In the Thurgau dataset, the stratigraphy is currently incorrectly extracted for this borehole: https://boreholes.swissgeol.ch/54394/stratigraphy/18978#lithology

The corresponding profile is found on page 4 of `5014_Umnutzung_Muehle_Rickenbach_KB4-17_17.pdf` in the TG train dataset.

The elevation is incorrectly extracted as an additional layer at the top, whereas the description of the last layer is not extracted at all. As a consequence, the alignment between the depth intervals and the descriptions is also off-by-one along the entire borehole.

<img width="689" height="270" alt="Image" src="https://github.com/user-attachments/assets/306831f3-23ff-4081-b874-cdf64838e0e7" />

This could be related to the changes from https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/437 / https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/470.

It is also related to the open issue https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/471

Issue #325, opened 2025-10-24
Refactor `AtoBSidebar` to use a proper `SidebarEntry` subclass and simplify Interval types


TicaGit:
### Description

Currently, the sidebar and interval type hierarchy are inconsistent, which makes type handling and future refactoring more complex.

Every Sidebar is supposed to have entries of type SidebarEntry, however, AtoBSidebar uses entries of type AToBInterval, which is not a subclass of SidebarEntry. (extraction/features/stratigraphy/sidebar/classes/a_to_b_sidebar.py)

### Proposed Solution

- Create a new entry type, subclass of SidebarEntry and representing entries for AtoBSidebar.
```
class AtoBEntry(SidebarEntry[tuple[float, float]]):
    ...
```

- Update AtoBSidebar to use AtoBEntry instances instead of AToBInterval.
Adjust any methods that currently depend on AToBInterval accordingly.

- Unify interval handling once this refactor is done:
   - The Interval class will represent a simple (start, end) structure, independent of the type.
   - Delete all other specialized interval types (AAboveBInterval, AToBInterval, etc.).
   - Each Sidebar implementation can define its own depth_intervals() method to extract intervals from its entry type.

PR #519, opened 2026-07-10
Take table structures into account when clustering depth values

stijnvermeeren-swisstopo:
Provides partial solutions for and #325.

- Use consistent sidebar entry types, also for the intervals in the AToBSidebar
  - This provides a partial solution for #325.
  - Also fix errors related to IntervalZone that were reported by Copilot
- (Not just for AAboveBSidebars as previously, but consistently for all sidebar types:) Partition potential sidebar entries by containment in detected table-like structure (if any), and only apply clustering within each such partition (more accurate and more efficient).
  - This also allows for an simpler implementation of the condition that a ProtocolSidebar is only allowed within a table-like structure.
- Only allow horizontally oriented text lines in material descriptions (width > height). Even though there are a few PDF files where descriptions are oriented vertically (e.g. ZH 699262001-bp.pdf), the rest of the pipeline is not able to deal with that correctly anyway, so it is safer to ignore these from the start (avoids false positives).
- Replace the old `detect_misalignment` condition for clusters/columns of sidebar entries with a new, simpler condition based on the vertical lines in the detected table structures (if any). In particular, we only allow a cluster when there is no vertical line that splits the entries in two (some entries to the left of the line, other entries to the right).
  - Having a more robust condition here will help solve #512 in a future PR.
- Fix an incorrect inequality in the condition for vertical lines in the TableStructure class.
  - Also added a simple unit test for the table detection that would have caught this bug.
- Decrease the max angle deviation for vertical/horizontal lines in table-like structures from 10° to 5°. This avoids some false positives (e.g. the diagonal lines in the strip log between 0m and 1.7m in ZH 675255008-bp.pdf) while having no immediate negative effect on other benchmarking inputs. Potentially, we could make these conditions even more precise by detecting and taking into account the overall page skew, but this does not seem to be necessary for the time being.
- Make the regex in the DepthColumnEntryExtractor class stricter (no longer allowing suffixes like "müM" or "mm" here).
  - Also added some unit test cases for this.
- Require that in every entry of an AToBSidebar, the start depth show be strictly above the end depth.

**Effects on ZH benchmarking**
Layer F1 improves from 84.4% to 84.5%. Metadata metrics improve more significantly.

Improvements:
- `268124587-bp.pdf`: gets rid of false positive sidebar & (vertical) material descriptions
- `268124613-bp.pdf`: top of borehole profile is now fully detected (though still with a couple of FP depth values) 

Affected but quality neither clearly better nor clearly worse:
- `681249142-bp.pdf` 
- `684252058-bp.pdf`
- `690251013-bp.pdf` (different second FP borehole detected, groundwater now linked to the correct borehole, but that's mostly a lucky coincidence)
- `691251009-bp.pdf` (detection is more complete, but still far from accurate) 

**Effects on Geoquat benchmarking**
Layer F1 improves from 48.4% to 48.9%. Other metrics improve similarly.

Improvements:
- `13080.pdf`: false positive borehole no longer detected (vertical descriptions)
- `14273.pdf`: false positive borehole no longer detected (vertical descriptions)
- `1506.pdf`: better sidebar
- `5872.pdf`: false positive borehole no longer detected
- `A11395.pdf`: ruler no longer incorrectly included in material descriptions
- `A202.pdf`: last layer now correctly extracted (previously, the value `200` would interfere and cause the `detect_misalignment` condition to fail, leading to the rejection of the entire cluster)
- `A359.pdf`: grouping of material descriptions is more accurate (OCR artifact that interfered is now ignored as vertical text)
- `A435.pdf`: material description of first layer now correctly included (upward expansion of material description rect works better when ignoring text lines that are higher than they are wide)
- `A57.pdf`: incorrect sidebar no longer detected
- `A8561.pdf`: more accurate grouping of material descriptions (bullet points `-` that were interfering are now ignored as text lines that are higher than they are wide)
- `B626.pdf`: arithmetic progression within depth values no longer incorrectly removed (previously, this was a side-effect of an OCR artifact "`0`" that was detected in the strip log, and interfered with the logic; now, since there is a verticaly line from the table structure in between, this interference no longer happens)


src/extraction/features/extract.py
```diff
@@ -353,8 +353,7 @@ def _find_depth_sidebar_pairs(self) -> list[MaterialDescriptionRectWithSidebar]:
         )
         for column in a_to_b_sidebars:
             for entry in column.entries:
-                used_entry_rects.add(entry.start.rect)
-                used_entry_rects.add(entry.end.rect)
+                used_entry_rects.add(entry.rect)
```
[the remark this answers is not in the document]
stijnvermeeren-swisstopo: The new implementation is fine for our use case.

src/extraction/features/stratigraphy/sidebar/utils/cluster.py
```diff
@@ -1,15 +1,64 @@
...
+
+                extended_set.add(entry)
+                if line_y0 <= rect.y1 and rect.y0 <= line_y1:
+                    inside_set.add(entry)
+
+        return VerticalLinePartition(
+            line, left=left, right=right, left_extended=left_extended, right_extended=right_extended
+        )
+
+    def no_conflict(self, partition: "VerticalLinePartition[EntryT]") -> bool:
```
AgathaSchmidt: I can't find where `no_conflict` is used

AgathaSchmidt (commented):
Looks good! 
Great improvement on all those files!

src/extraction/features/stratigraphy/sidebar/utils/cluster.py
stijnvermeeren-swisstopo: Seems to be a relict of some earlier implementation, I've removed it.

AgathaSchmidt (approved):

PR #522, opened 2026-07-15
Classification of borehole type

### LINKED ISSUE #448: Classification of borehole type
The borehole types that are currently available in the Boreholes database schema are `"borehole"`, `"virtual borehole"`, `"penetration test"`, `"trial pit"`, `"outcrop"`, `"other"` and `"not specified"`: [borehole_type_codelist.csv](https://github.com/user-attachments/files/26602283/borehole_type_codelist.csv)

This classification is Prio 1 for 2026.

There is a lot of labelled data available in the Boreholes database across all workgroups. The data is strongly dominated by the type `"borehole"`. It is not clear how trustworthy this data is. It seems likely that `"borehole"` was sometimes used as a "dummy value" even for boreholes for which a different type would be more accurate. (This applies in particular to the canton of Zurich workgroup, where `"borehole"` is the only type that is currently in use.) [borehole_type_stats.csv](https://github.com/user-attachments/files/26602312/borehole_type_stats.csv)

Overall, we have
- 43134x `"borehole"`
- 1325x `"trial pit"`
- 731x `"penetration test"`
- 36x `"other"`

The types `"virtual borehole"` and `"outcrop"` are currently not used in the DB. The type `"not specified"` is never used, but there are 2031 boreholes where the type is not defined at all (`NULL` value in the DB).

```sql
WITH data AS (
SELECT
  id_bho,
  name_wgp,
  (SELECT text_cli_en from bdms.codelist WHERE codelist.id_cli = borehole_type_id) as property,
  (SELECT STRING_AGG(identifier_value, ',') FROM bdms.borehole_identifiers_codelist WHERE identifier_id = 100000006 AND borehole_id = id_bho) as geoquat_id
FROM bdms.borehole
JOIN bdms.workgroups ON id_wgp_fk = id_wgp
)

SELECT property, COUNT(*) as boreholes, COUNT(geoquat_id) as geoquat_boreholes FROM data
GROUP BY property
ORDER BY COUNT(*) DESC
```

AgathaSchmidt:
Closes #448 

## Data 

borehole_type

Note: zuirch_ground_truth_new2 includes 3 hand labeled penetration test boreholes

## Results: two approaches 
**First approach** 
Dataset: n = 816 | n borehole = 432 | n penetration test 51 | n trial pit 334 
Searching for known German drilling-method keywords associated with "trial_pit" and "penetration_test".
Absence of any keyword defaults to "borehole".
97.1% of files in the selected dataset are german (only 1 of the 68 missed  trial_pit/penetration_test cases is a non-German file)
Note: "ambiguous" describes when keywords from two different classes match 

**Accuracy: 714/816 = 89.6%**

| True \\ Predicted | ambiguous | borehole | penetration_test | trial_pit |
|---|---|---|---|---|
| borehole | 1 | 424 |3 | 3 |
| penetration_test | 0 | 23 | 28 | 0 |
| trial_pit | 10 | 45 | 0 | 279 |

Some "penetration test" are impossible to classify keyword-based since there is no text, eg. 

Screenshot 2026-07-20 at 08 57 23

A293, A292, A201, A198, A197, A1216, A1217, A1218, 

Also a problem are long documents where the borehole is e.g. on page 18 (1019_Kreuzlingen_5037_RS_1) -_> since we are not identifying the borehole in this approach but simply read out the text and after some short cleaning use it as input, the tokens are already exceeded from the text of the 17 previous pages. (1020_Kreulingen_5037_RS4, 1080_Kreuzlingen_1361_RKS_RS1)

**Second approach** 
Document level training: 
We trained with one label per pdf. That means that we missed a little bit of data for the training process. 
The main advantage of this is that we don't have to run the full extraction pipeline, detecting boreholes. (e.g. for Thurgau this a long time) 
By just extracting the text we had training data available quickly.

Dataset | Local PDFs | Matched to ground truth | Multi-borehole PDFs
-- | -- | -- | --
zurich | 160 | 160 | 10
thurgau | 6,245 | 6,186 | 0
geoquat/train | 1,159 | 1,159 | 1
deepwells | 111 | 65 | 0

| Class | Last Evaluation F1 | Test F1 (global) |
|---|---|---|
| borehole | 0.993 | 0.996 |
| penetration_test | 0.615 | 0.667 |
| trial_pit | 0.946 | 0.967 |
| **macro F1** | – | **0.8216** |
| **micro F1** | – | **0.9918** |

## Inference 
For inference in the API a borehole level prediction was enabled. 
The model was trained on document level text as input and produced one label as output. 
When getting the PDF from the API (per drag and drop) we run the extraction pipeline to identify every borehole in the pdf and assign an index. 
Then the respective text assigned to the borehole is fed into the trained model and we get per borehole labels 

## Limitations 
Due to the document-level training data approach we are loosing some valuable data. 
First, we loose the labelled data of the other potential boreholes in our document. 
Secondly, for files like 1080_Kreuzlingen_1361_RKS_RS1, where the borehole is on the last page, our current simple text extraction approach never includes the text for the last page since the tokens are used up before by the text form the previous pages.  
Potential improvements: 
- get borehole level training data e.g. by running extraction pipeline as in inference or by running the assets pipeline to identify which pages are actually including a borehole 

# Final summary of changes 
- New "borehole type" classification system plugged into the existing classification pipeline (training, evaluation, and CLI tooling all support it).
- New API endpoint that takes an uploaded PDF and returns the predicted type for each borehole found in it.
- Since this classification works at the whole-document level (not per material-description layer like existing classifiers), the pipeline now knows how to extract and use the relevant header text from a document as input, and can train/evaluate on that.
- The CLI for running classifications can now take multiple input files/datasets at once and merge them into a single run.
- Ground truth data now also stores borehole_type and borehole_purpose per borehole
- New training configuration for the borehole-type BERT model. 
- Added a dependency needed to support file uploads in the API.


stijnvermeeren-swisstopo:
@AgathaSchmidt 
It is not entirely clear how to compare the accuracy (89.6%) from the first approach (keyword-matching) with the numbers that you reported for the second approach.

I would be interesting to look at a few concrete examples of the following:
- Boreholes that are marked with type "borehole" in the ground truth, but are classified by our methods as something different. (Since we suspect that the GT over-uses the generic "borehole" type, it would be interesting to check if the classification might actually be more correct in some cases, even when the GT says that the correct type should be "borehole".)
- Boreholes where the keyword-matching approach does not find any relevant borehole type keywords, but the trained model still classifies the type as something different from the generic "borehole". This would help us understand what else the model might have learned to recognize in the documents in order to differentiate between the different types.

AgathaSchmidt:
@stijnvermeeren-swisstopo  Indeed, the numbers are not completely comparable. Even though with micro F1 we have a number representing accuracy, the data is different. 
In the keyword based approach I manually selected a subset which includes almost all the non-borehole class cases while the BERT numbers are based on the automatic train/val/test split. 
Here are some tables to answer your questions: 

### First question 
<p><strong>Keyword-based approach</strong></p>

filename | predicted | true_label | matched_keywords
-- | -- | -- | --
Positive |   |   |  
267126034-bp.pdf | trial_pit | borehole | Sondierschacht
2910_SL4_Ermatingen_787.pdf | trial_pit | borehole | Baggerschlitz;Baggerschlitz;SL
2951_BS_1-4_Eschlikon_3126.pdf | trial_pit | borehole | Baggerschlitz;Baggerschlitz;BS;BS
3172_RKS_41_35_Kreuzlingen_Weinberg.pdf | penetration_test | borehole | Rammkernbohrung
954_EWS_Lengwil_811_RKB1.pdf | penetration_test | borehole | Rammkernbohrung;Rammkernbohrung
False positive |   |   |  
5237_EWS_Uttwil_27.pdf | penetration_test | borehole | RS




<p><strong>BERT model (test data)</strong></p>

filename | predicted | true_label
-- | -- | --
Positive |   |  
267123094-bp.pdf | trial_pit | borehole (Baggerschlitz)
267125358-bp.pdf | trial_pit | borehole (Baggerschacht)
2948_BS_1-4_Eschlikon_3126.pdf | trial_pit | borehole (Baggerschlitz)
3179_BS_41_134_Kreuzlingen_Weinberg.pdf | trial_pit | borehole (Baggerschlitz)
4408_Neubau_MFH_Ermatingen_BS_1_18.pdf | trial_pit | borehole (Baggerschlitz)
False positive |   |  
5403_Kehrichtdeponie_Rueti_Siegershausen_KB1_20.pdf | trial_pit | borehole (Kernborhung)
  |   |  





### Second question: (missed labels by keyword which were found by BERT ) 

- A1188pdf trail pit  (VERSICKERUNGSSCHAECHTE)
- 698_Lindenacker_Eschlikon_BS1.pdf trail pit (Baggerschlitz handwritten)



filename | true_label | keyword_predicted | bert_predicted | comment
-- | -- | -- | -- | --
1019_Kreuzlingen_5037_RS_1.pdf | penetration_test | borehole | penetration_test | long document
1446_BS_Münchwilen_2104_BS4.pdf | trial_pit | borehole | trial_pit | BAGGERSONDIERSCHÄCHTE
1787_EWS_Romanshorn_2044_BS1.pdf | trial_pit | borehole | trial_pit | 2 Baggerschlitzen
4796_Baugrundabklaerung_St_Margarethen_Muenchwilen_BS2_15.pdf | trial_pit | borehole | trial_pit | long document
4797_Baugrundabklaerung_St_Margarethen_Muenchwilen_BS3_15.pdf | trial_pit | borehole | trial_pit | long document
4835_Neubau_MFH_Stocken_Oberneunforn_BS2_20.pdf | trial_pit | borehole | trial_pit | BAGGERSONDIERSCHÄCHTE
4838_Neubau_2_MFH_Stocken_Oberneunforn_BS2_17.pdf | trial_pit | borehole | trial_pit | BAGGERSONDIERSCHÄCHTE
5403_Kehrichtdeponie_Rueti_Siegershausen_KB1_20.pdf | borehole | borehole | trial_pit | Kernbohrung
816_Nussbaum_Matzingen_AG_Kesswil_RK_T5-2.pdf | penetration_test | borehole | penetration_test | Rammkernsondierungen
979_Kreuzlingen_Ueberbauung_Sallmannpark_Loewenareal_BS2.pdf | trial_pit | borehole | trial_pit | BAGGERSONDIERSCHÄCHTE
A1188.pdf | trial_pit | borehole | trial_pit |  VERSICKERUNGSSCHAECHTE





<p><strong>Comparable accuracy --> Same data (test BERT)</strong></p>

Approach | Correct | Accuracy
-- | -- | --
BERT | 930/943 | 98.6%
Keyword (full text) | 908/943 | 96.3%


class | support | BERT F1 | Keyword F1
-- | -- | -- | --
borehole | 881 | 0.993 | 0.981
trial_pit | 55 | 0.930 | 0.814
penetration_test | 5 | 0.500 | 0.133
other | 2 | 0.000 | 0.000
macro avg |   | 0.606 | 0.482
micro avg | 943 | 0.986 | 0.963

Unfortunately the test data does not include many penetration test samples






All in all: by adding new keywords like BAGGERSONDIERSCHÄCHTE or changing the algorithm to deal better with longer documents we could probably push the performance of the keyword based approach to be better than the BERT based approach. 
Nevertheless, a strong pro argument for the BERT approach is that it works well, without having to investigate so deeply and spending much time on finding the right keywords etc. 

With no doubt, we gain valuable insights from these investigations. Nevertheless, we should evaluate wether the time spend and the control are worth it for all the other tasks coming up, or if we decide for one universal BERT approach which is suitable for many different tasks.

src/app/common/schemas.py
```diff
@@ -926,3 +926,36 @@ class ClassifyResponse(BaseModel):
...
+    """A single borehole's predicted type."""
+
+    borehole_index: int = Field(
+        ...,
+        description="Index of the borehole within the document (0-based).",
+    )
+    class_name: str = Field(
+        ...,
+        description="Predicted borehole type for this borehole.",
+    )
```
christianabbet, line 949: Shouldn't we also have the classes of `BoreholeTypeClasses` (not_specified, borehole, ...) there as for the classify all 

Image

src/classification/utils/datasets/classification.py
```diff
@@ -368,16 +396,72 @@ def process(
...
+            if (cls.reduce_label(layer, borehole) is not None or allow_none) and layer.material_description is not None
         ]
 
+    @classmethod
+    def _process_documents(
+        cls,
+        ground_truth: dict[str, list[GroundTruthBoreholeWithLanguage]],
+        document_text: dict[str, str],
+        allow_none: bool,
+    ) -> list[LayerInformation]:
```
christianabbet, line 373: I would just put the function above `process` for readability

src/extraction/features/classification_text.py
```diff
@@ -0,0 +1,66 @@
...
+    borehole_index: int
+    text: str
+
+
+def _borehole_pages(borehole: BoreholePredictions) -> list[int]:
+    """Return the sorted, deduplicated 1-indexed page numbers a borehole's bounding boxes span."""
+    return sorted({bboxes.page for bboxes in borehole.bounding_boxes})
+
+
+def extract_borehole_texts(file: Path | BytesIO, filename: str, header_only: bool = False) -> list[BoreholeText]:
```
christianabbet: It seems header_only = True is the default behaviour. Should we then set it to True by default ? Or rename the variable like : "include_description: False"

src/app/common/schemas.py
```diff
@@ -926,3 +926,36 @@ class ClassifyResponse(BaseModel):
         default=None,
         description="Predicted USCS class (unconsolidated sediment only).",
     )
+
+    model_config = ConfigDict(json_schema_extra={"example": {"class_name": "Marlstone"}})
```
christianabbet, line 931: shouldn't this be in the BoreholeTypePrediction or below ?

src/app/api/v1/endpoints/classify_borehole_type.py
```diff
@@ -0,0 +1,47 @@
...
+    It runs the extraction pipeline to detect each borehole and its page
+    span, extracts and filters header-like text scoped to that borehole (matching how the model was
+    trained), and runs a single full forward pass through the `borehole_type` model per borehole — no
+    shared-embedding shortcut across models, since `borehole_type`'s fine-tuned head/pooler layers are
+    specific to this model.
+
+    Args:
+        data (bytes): Raw bytes of the uploaded PDF document.
+        filename (str): Name of the uploaded file, used as an identifier.
+        bert_models: Models loaded at startup via the lifespan, keyed by classification system name.
```
christianabbet: type missing after `bert_models (...)`

src/classification/config/bert/bert_config_borehole_type.yml
```diff
@@ -0,0 +1,81 @@
...
+      - zurich_ground_truth_new2.json
+      - thurgau_ground_truth_new.json
+      - geoquat_ground_truth_new.json
+      - deepwells_ground_truth_new.json
+    document_texts:
+      - zurich_filtered_text.json
+      - thurgau_filtered_text.json
+      - geoquat_train_filtered_text.json
+      - deepwells_filtered_text.json
+  zurich_borehole_types:
```
christianabbet: why plural ?

src/classification/runner.py
```diff
@@ -47,54 +53,181 @@ class ClassificationOptions:
...
+    merged: dict[str, list[GroundTruthBorehole]] = {}
+    for json_path in _expand_json_paths(paths):
+        merged.update(GroundTruth(json_path).ground_truth)
+    return merged
+
+
+def _extract_file_text(pdf_path: Path) -> str:
+    """Extract a single file-level text for `pdf_path`, merging all its boreholes (dropping exact repeats).
+
+    Mirrors `scripts.extract_full_text_csv`, the script used to precompute the *_filtered_text.json
```
christianabbet: What is this file ? I cannot find it in `scripts` folder

src/classification/runner.py
```diff
@@ -104,7 +237,8 @@ def run_classification_predictions(
     This is the core prediction logic, decoupled from tracking and evaluation.
 
     Args:
-        file_path (Path | None): Path to the JSON file containing material descriptions to classify.
+        file_paths: One or more JSON files (or directories of such files) containing material
```
christianabbet: Type missing

christianabbet (changes requested):
API / Docker works fine. Test it with few documents (boreholes and trial pit).  It would be nice to have the results in docs/Experiment_Results.md with the metrics.

Few comments from Claude for documentation that seemed coherent:
* Shouldn't we document the approach in the .md files? We already have something for the other endpoint `classify`
* `borehole_type` missing from [README.md:246-247]
* `-dt` new option for main.py not documented in readme

src/classification/runner.py
AgathaSchmidt: Its an artifact 
I will remove the comment

christianabbet (approved):
LGTM

PR #533, opened 2026-08-03
Refactor code without MaterialDescriptionRectWithSidebar class

stijnvermeeren-swisstopo:
Simplify the code by getting rid of the `MaterialDescriptionRectWithSidebar` class.

There is also a slight modification in the scoring of candidate boreholes, where we use a more accurate bounding box that is derived from the actual material description lines (after processing/filtering) instead of using the original rectangle (before any processing/filtering happens). Because this puts the "layer indicator" layouts somewhat at a disadvantage (more filtering happens here), we compensate for this by given a factor 1.5 boost to this sidebar type.

Overall, this means that there are no changes to the extraction results on the Zurich and Geoquat validation datasets.


src/extraction/features/extract.py
```diff
@@ -128,20 +125,25 @@ def process_page(self) -> list[ExtractedBorehole]:
         return [candidate.borehole for candidate in valid_candidates]
 
     def _contained_in_table_index(
```
AgathaSchmidt: This method is never called 
Pre-existing dead code ?

AgathaSchmidt (changes requested):

src/extraction/features/extract.py
stijnvermeeren-swisstopo: @AgathaSchmidt 

I did some digging into the Git history and it seems like the usage of this method was removed in https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/403.

In that PR, I don't see any comments that claim that keeping this method is intentional (e.g. arguing that the method might still be useful in the future). For this reason, I would assume that this method remains in the codebase by accident, and I will now remove it.

src/extraction/features/stratigraphy/layer/page_bounding_boxes.py
```diff
@@ -90,20 +43,20 @@ def from_json(cls, data) -> "PageBoundingBoxes":
         )
 
     @classmethod
-    def from_material_description_rect_with_sidebar(
-        cls, pair: MaterialDescriptionRectWithSidebar, page_number: int
+    def from_sidebar_and_rect(
+        cls, sidebar: None | Sidebar, material_description_rect: pymupdf.Rect, page_number: int
```
AgathaSchmidt: Sidebar | None

AgathaSchmidt (approved):

PR #534, opened 2026-08-03
fix(classification): fall back to argmax logit instead of class 0 for empty predictions

### LINKED ISSUE #514: Compute different predictions for classification, causing inconsistent F1 metrics
## Description
The evaluation metrics produced during training's test-set evaluation ([train.py](src/classification/models/train.py)) do not match the metrics produced by the standalone classification pipeline ([main.py](src/classification/main.py)) for the *same* trained model and the *same* ground-truth data. Example: mineral_components on Azure, deepwells (enhanced) test set — `train.py` reports micro/macro F1 of `0.7799692` / `0.9424132`, while `main.py` reports `0.7772` / `0.9372` for what should be an identical evaluation. This has also been observed on classification systems beyond mineral_components.

## Root cause (identified via code inspection)
The two pipelines use **different fallback rules** for multi-label/rank tasks when no class's logit exceeds the 0 decision threshold:

- `train.py:490-494` (`ConfusionMatrixCallback.compute_metrics`, used for train/test evaluation): if no logit is `> 0`, falls back to the **argmax logit class** — i.e. it credits the model's single best guess.
  ```python
  predictions = (logits > 0).astype(int)
  id_no_prediction = predictions.sum(axis=-1) == 0
  predictions[id_no_prediction, logits[id_no_prediction].argmax(axis=-1)] = 1
  ```
- `model.py:339-340` and `model.py:343-348` (`BertModel.predict_class_batched`, used by `main.py`'s inference pipeline via `runner.py`): if no logit is `> 0`, falls back to a **hardcoded class index 0** (`not_specified`), regardless of what the model actually scored highest.
  ```python
  prediction = [row.nonzero(as_tuple=True)[0].tolist() or [0] for row in (outputs.logits > 0)]
  # rank task: same `or [0]` fallback
  ```

For single-label tasks both paths use plain `argmax` (`train.py:500` vs `model.py:342`) and are identical — consistent with the discrepancy specifically affecting `multi_label`/`rank` systems (e.g. `mineral_components`, `accessory_components`, `organic_components`, `debris`, `en_secondary`).

Tokenization (`max_length=512`, `truncation=True`, `padding="max_length"`) is identical in both paths, ruling that out as a factor.

## Steps to Reproduce
1. Train a model on a `multi_label` or `rank` classification system (e.g. `mineral_components`) via `train.py`.
2. Note the test-set micro/macro F1 reported by `train.py`'s built-in evaluation (`ConfusionMatrixCallback`).
3. Run the same trained model through `main.py`'s classification pipeline (`-c bert`) on the same ground-truth file, then score it with `evaluate()` in `evaluate.py`.
4. Compare micro/macro F1 between the two — they differ.

## Expected Behavior
The same model evaluated on the same data should produce identical metrics regardless of which entry point (`train.py` or `main.py`) computes them.

## Actual Behavior
`main.py`'s pipeline scores consistently lower on both micro and macro F1 for multi-label/rank systems, because ambiguous ("no confident class") samples are always mapped to `not_specified` instead of the model's actual best guess.

## Screenshots / Logs
```
train.py  (Azure, deepwells enhanced, mineral_components): micro F1 = 0.7799692, macro F1 = 0.9424132
main.py   (same model/data):                                micro F1 = 0.7772,    macro F1 = 0.9372
```

## Additional Context
- Also observed on classification systems other than `mineral_components` — not yet exhaustively checked per-system, but not isolated to one.
- Not currently blocking, but affects reliability of model comparison/selection when relying on `train.py`'s reported test metrics as a proxy for production (`main.py`) performance.
- Secondary risk worth flagging: the two pipelines compute metrics via **two independently maintained implementations** — sklearn's `classification_report` in `train.py` vs a custom set-based tp/fp/fn accumulator in `evaluate.py` (`per_class_metric`/`AllClassificationMetrics`). Even after fixing the fallback-rule mismatch, consider unifying these into one shared metrics path to prevent this class of bug from recurring.
- Unverified open question: whether `main.py`'s "deepwells (enhanced)" benchmark run evaluates the exact same test-set sample slice (`split_samples` deterministic hash split, used in `train.py`'s `test_sets`) or the full ground-truth file — worth confirming this isn't a contributing factor alongside the fallback-rule bug.

christianabbet:
Closes #514 

## Summary
In the multi-label and rank classification tasks, when no logit exceeded the classification threshold, the model defaulted to predicting class `0` in `main.py` and `argmax` in `train.py`. This caused inconsistent behavior between training and inference. To keep online runs as ground truth, we align with the logic of training and use argmax as fallback for inference. 

This logic also affected the `en-secondary` ranking pipeline where values were recomputed.


AgathaSchmidt (approved):
Great find!! 
Thanks for pointing that out and fixing it

PR #538, opened 2026-08-04
Exclude sidebar entries from material descriptions

stijnvermeeren-swisstopo:
- Prevent sidebar entries (e.g. depth values) from being included as part of the material descriptions, even when there is an overlap between the detected sidebar and the bounding box of the material descriptions. This has a significant positive impact on the stratigraphy F1 scores, especially in the Geoquat dataset, where there is such an overlap for many borehole profiles.
- Remove the old `LayerIdentifierSidebar._clean_block` method, which removed the layer identifiers from the block, as this is now done more generically (layer identifiers are sidebar entries), making this method redundant. 
- Keep track of a dedicated bounding box for `AToBInterval` objects, instead of implicitly deriving the bounding box as the union of the bounding boxes of the start and end values. This enables cleaner removal of these intervals from the final material description text (see example `10626.pdf` below).
- In order to obtain these bounding boxes more accurately, we now detect `AToBInterval`s from `TextLine` objects instead of from individual `TextWord` objects. This is more robust and accurate anyway.

Example of overlap between sidebar and material descriptions bounding box (Geoquat `A391.pdf`):
image

Example `10626.pdf` with implicit `AToBInterval` bounding box (old):
image

Example `10626.pdf` with explicit `AToBInterval` bounding box (new):
image

Geoquat benchmark:
- Layer F1 improves +0.9% from 48.8% to 49.7%
- Depth interval F1 improves +0.2% from 63.6% to 63.8%
- Material description F1 improves +0.9% from 64.0% to 64.9%
- Improvements on the following files:
  - `10088.pdf`
  - `10626.pdf`
  - `5793.pdf`
  - `A11281.pdf`
  - `A11285.pdf`
  - `A218.pdf`
  - `A318.pdf`
  - `A391.pdf`
  - `A418.pdf`
  - `A430.pdf`
  - `B147.pdf` (better detection of `AToBInterval`s on page 2)
  - `B258.pdf`
  - `B542.pdf` (better detection of `AToBInterval`s)
  - `B80.pdf` 

Zurich benchmark:
- Layer F1 improves +0.2 from 84.5% to 84.7%
- Depth interval F1 remains at 89.6%
- Material description F1 improves +0.3% from 88.4% to 88.7%
- Improvements on the following files:
  - `269124219-bp.pdf` (better detection of `AToBInterval`s)
  - `681249142-bp.pdf`


src/extraction/features/extract.py
```diff
@@ -232,6 +232,20 @@ def _get_interval_block_pairs(
...
+            # remove description words that are already part of the sidebar
+            clean_description_lines = []
+            for text_line in description_lines:
+                clean_words = [
+                    word
+                    for word in text_line.words
+                    if not any(entry.rect.contains(word.rect) for entry in sidebar.entries)
+                ]
+                if len(clean_words) > 0:
+                    clean_description_lines.append(TextLine(words=clean_words))
```
stijnvermeeren-swisstopo: Need to pass `text_angle` here once https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/535 is merged.

stijnvermeeren-swisstopo: Done.

AgathaSchmidt (approved):
Looks good, 
I think the TODO comment in line 53 of `src/extraction/features/stratigraphy/sidebar/extractor/layer_identifier_sidebar_extractor.py` can be removed ? 

"TODO: Similar to AToBSidebarExtractor.find_in_words(). Refactoring may be desired."

stijnvermeeren-swisstopo:
Maybe, these methods seem to have diverged quite a bit already, so they are probably not as similar anymore compared to when this TODO was first written. I think it's ok to remove it.

PR #541, opened 2026-08-06
Qualitative analysis of missing incorrectly extracted groundwater level on zh dataset

### LINKED ISSUE #515: Qualitative analysis of missing / incorrectly extracted groundwater level on ZH dataset
## Description
Go through the Zurich (ZH) dataset and perform a qualitative analysis of groundwater level detection failures — cases where the groundwater level is missing or incorrectly extracted. Identify whether these failures are linked to symbol detection, text detection, or other causes.

## Motivation
Groundwater level detection currently achieves only around 50% accuracy on the ZH dataset. We want to understand the root cause(s) of failure so we can target the right fix and raise detection accuracy to 80% ideally.

## Proposed Solution
1. Review groundwater level extraction results on the ZH dataset and collect cases of missing or incorrect detections.
2. Categorize each failure by root cause: symbol detection, text detection, or other.
3. For each failure category, assess whether existing approaches (e.g. Bedrock, BERT-based classification, or keywords matching) can resolve it.
4. Where existing approaches are insufficient, propose an alternative method to improve groundwater level detection/classification.

AgathaSchmidt:
Closes #515

# Error analysis
Analysis based on `data/zurich/` (160 PDFs) against `data/zurich_ground_truth.json`
(Most information in the tables was gathered with Claude Code)

root_cause | count | meaning
-- | -- | --
text_detection | 123 (64 incorrect + 59 missing) | a groundwater keyword was found somewhere in the document
other | 43 (all missing) | neither keyword nor symbol found anywhere in the document
symbol_detection | 8 (5 incorrect + 3 missing)| a visual groundwater symbol was found somewhere

Only 14 files have a symbol detected:
`267122001`, `267125315`, `267125364`, `267125378`, `268124336`, `268124375`, `268124561`, `268124570`, `268124587`, `269125064`, `269126143`, `677250019`, `679249004`, `690251013`

...even though many other files visually have the same stacked-line symbol. Investigated why, and it splits into two distinct causes:

**Keyword is present but groundwater extraction fails:** 

Missing (59 rows) 

category | count | meaning
-- | -- | --
key_not_near_value | 25 | the ground-truth value isn't near any matched key at all
some_fields_near_key | 34 | only some of date/depth/elevation are near a key — the rest simply aren't printed as text anywhere nearby
(all fields near key but still missed) | 0 | —

**Incorrect (64 rows)**

category | count | meaning
-- | -- | --
wrong_value_near_same_key | 31 | the correct value is **also** inside the same key's window as the wrong one extracted — right neighborhood, wrong pick among several nearby candidates
correct_value_near_different_key | 20 | the correct value sits near a different occurrence of the key elsewhere on the page (e.g. a key repeated in multiple places) — wrong occurrence got matched
correct_value_nowhere_near_any_key | 7 | same positioning problem as the missing bucket
no_real_gt_entry_exists | 6 | pure hallucination — no real groundwater reading exists for that borehole at all; the extractor invented one from unrelated nearby text

- **Shape near-miss** (small population): the symbol *is* found within the current search window, but fails one geometric check by a small margin — e.g. `267122002`/`267123041` fail only on the gap/overlap check, and `683251057`/`683251058` have a textbook-correct symbol (right length, right ratio, right gap) but it's drawn ~10pt beside the depth label instead of directly above/below it, so the "line overlaps the label" check rejects it.
- **Search-radius bottleneck** (the majority of the 146): the symbol sits 40–150pt away from the nearest depth/date-looking text, entirely outside the window the search even looks in — no threshold tuning touches these, since the candidate line pair is discarded before any shape check runs.

**Tried and reverted, no net improvement:**
- Widened the search radius + loosened the shape thresholds (ratio, gap) to catch both buckets at once. Benchmarked against ground truth: recall was **completely flat** (one new correct match, one previously-correct match corrupted into a wrong one) while precision dropped (0.484 → 0.456) from **8 new false-positive** groundwater entries. Reverted.
- Padded-rectangle join in `get_axis_aligned_lines` (`data_extractor.py`) to close the diagonal search-reach gap described above. Implemented and benchmarked across the full corpus — no overall improvement. Reverted.

**Very difficult / out of scope**
- `267125307`, `267125308`, `267125309`, `267125310` → depth only readable from a line on the ruler, and in `681249142` only a line in the striplog
- `677259003`, `695265009` → can't find groundwater at all
- no keyword, no symbol, no color: `695248002`, `697243001`, `700246002`
- Ground truth different from what I see in the document: `267125339`, `267125338`, `267125334`

**Shared-code limitation identified, not merged**
- The key→value proximity search used by *every* extractor (coordinates, elevation, name, groundwater — `get_axis_aligned_lines` in `data_extractor.py`) only searches a cross-shaped region (same row, or same column, extended outward), never the diagonal quadrants. A value offset both sideways *and* up/down from its key can never be reached no matter how the four search factors are tuned. Prototyped a single-padded-rectangle join that covers the diagonal case too (verified against the existing unit tests, all pass), but reverted without a full-corpus benchmark validation — given the symbol-detection lesson above, a change shared across every extractor needs measurement before landing, not just passing unit tests.

# First implemented improvements
### Merge candidates
Fixed a double-counting bug: a candidate with only 1 of 3 fields (e.g. depth+date but no elevation) was reported as-is, so when a second, complementary partial candidate existed elsewhere on the page, the same real reading counted as both a false negative and a false positive. `merge_weak_candidates` now conservatively merges a single-field candidate into the one non-conflicting richer candidate on the page (never merges two multi-field candidates, to avoid corrupting genuinely distinct multi-reading documents).

### Search above
`search_above_factor` was `0` while `search_below_factor` was already `2` — the key→value search only ever looked below a groundwater key, never above it. Raised to `2` to match, recovering readings printed above their key.
- `268125539`
- `683248005`

### Color
Some documents highlight the groundwater reading in a color that differs from the rest of the page. `get_minority_color_lines` (new) flags lines in a non-modal color that also contain a date, as an extra candidate area alongside keyword/symbol detection.
- `266126001`
- `267125439`
- `268123003`
- `268125592`
- `268124391`
- `268125539`

Screenshot 2026-08-06 at 15 17 57

Screenshot 2026-08-06 at 15 18 23
Screenshot 2026-08-06 at 15 18 50
Screenshot 2026-08-06 at 15 19 44
Screenshot 2026-08-06 at 15 20 08
Screenshot 2026-08-06 at 15 20 25

## Results
branch | gw_f1 | gw_precision | gw_recall
-- | -- | -- | --
develop |0.3797 | 0.4409| 0.3333 
515 | 0.4242| 0.4884| 0.375 

# Summary 
1. We have a lot of cases where a keyword is actually present but we still fail to extract the groundwater data. 
I tried tuning the existent search window etc but this did not lead to an overall improvement. 
Still, the searching algorithm based on a keyword has the biggest potential to detect more groundwater data. 
2. Also, there are a lot of stacked line symbols not identified as such. This is the second biggest potential in my opinion. 
I tried tuning some of those parameters in `src/extraction/features/groundwater/groundwater_symbol_detection.py`, but I was not very successful
 
```
    avg_text_size = sum(line.rect.height for line in text_lines) / len(text_lines)
    rel_size_ok = l_top.length * 0.8 > l_bot.length > l_top.length / 4  # bottom one is smaller, but not too small
    global_size_ok = avg_text_size < l_top.length < 3 * avg_text_size  # size comparable to avg text height
    pos_ok = l_top.start.x < l_bot.start.x and l_bot.end.x < l_top.end.x  # the top one fully "spans" the bottom one
    pos_to_bb_ok = hit_line.rect.x0 < l_top.end.x and l_top.start.x < hit_line.rect.x1  # longest line overlaps the bb
    gap_ok = l_bot.start.y - l_top.start.y < l_bot.length / 3  # gap is small, relative to the bottom line
    no_bb_in_gap = not any(
        line.rect.intersects(pymupdf.Rect(l_top.start.x, l_top.start.y, l_top.end.x, l_bot.start.y))
        for line in text_lines
    )  # no text line sits in between the two lines

```


src/extraction/features/groundwater/groundwater_color_detection.py
```diff
@@ -0,0 +1,45 @@
...
+        if "lines" in block
+        for line in block["lines"]
+        for span in line["spans"]
+    ]
+    if not span_colors:
+        return []
+
+    modal_color = Counter(color for _, color in span_colors).most_common(1)[0][0]
+
+    def line_color(text_line: TextLine) -> int | None:
```
christianabbet, line 33: one liner description ?

src/extraction/features/groundwater/groundwater_extraction.py
```diff
@@ -371,13 +372,59 @@ def get_groundwater_from_lines(
...
+
+        Args:
+            found_groundwaters (list[FeatureOnPage[Groundwater]]): The list of found groundwater features.
+
+        Returns:
+            list[FeatureOnPage[Groundwater]]: The list of groundwater features, with weak candidates merged.
+        """
+
+        def num_fields(gw: FeatureOnPage[Groundwater]) -> int:
+            return sum(v is not None for v in (gw.feature.depth, gw.feature.date, gw.feature.elevation))
```
christianabbet: one liner documentation?

src/extraction/features/groundwater/groundwater_extraction.py
```diff
@@ -371,13 +372,59 @@ def get_groundwater_from_lines(
...
+
+        def conflicts(a: FeatureOnPage[Groundwater], b: FeatureOnPage[Groundwater]) -> bool:
+            return any(
```
christianabbet: one liner documentation?

christianabbet (approved):
Good documentation of the problem. The code looks clean and logic self explanatory. Works well on my side. Same metrics when running the pipeline. Small comment for doc but already approved.

src/extraction/features/groundwater/groundwater_color_detection.py
```diff
@@ -0,0 +1,45 @@
...
+    Returns:
+        list[TextLine]: Lines in a minority color that also contain a date.
+    """
+    span_colors: list[tuple[pymupdf.Rect, int]] = [
+        (pymupdf.Rect(span["bbox"]), span.get("color"))
+        for block in page.get_text("rawdict")["blocks"]
+        if "lines" in block
+        for line in block["lines"]
+        for span in line["spans"]
+    ]
```
stijnvermeeren-swisstopo: I would add color as an (optional) attribute on the `TextWord` class, and directly set it in `extract_text.py`. Then we don't need to call `page.get_text` again here anymore. And having the text color information could well be useful in other places in the future as well.

src/extraction/features/groundwater/groundwater_extraction.py
```diff
@@ -371,13 +372,59 @@ def get_groundwater_from_lines(
...
+                getattr(a.feature, field) is not None
+                and getattr(b.feature, field) is not None
+                and getattr(a.feature, field) != getattr(b.feature, field)
+                for field in ("depth", "date", "elevation")
+            )
+
+        weak = [gw for gw in found_groundwaters if num_fields(gw) == 1]
```
stijnvermeeren-swisstopo: Why only do this when only one field is defined? I think it is plausible that some profiles have the same measurement with multiple attributes written in multiple places in the document. In such cases, it would also be beneficial to merge them. Did you test that?

The only case that I can think of where this could go wrong, is when there are non-conflicting measurements for different boreholes on the same page. But that is a risk that also already existing with your proposed implementation.

src/extraction/features/groundwater/groundwater_extraction.py
```diff
@@ -371,13 +372,59 @@ def get_groundwater_from_lines(
...
+        # return anyway, we can infer information later
         return FeatureOnPage(
             feature=Groundwater(depth=depth, date=date, elevation=elevation),
             rect=rect_union,
             page=page_number,
         )
 
+    def merge_weak_candidates(
+        self, found_groundwaters: list[FeatureOnPage[Groundwater]]
+    ) -> list[FeatureOnPage[Groundwater]]:
```
stijnvermeeren-swisstopo: A couple of unit tests for this method would be useful and should not be too difficult to write (probably even Copilot or Claude can help with that).

src/extraction/features/groundwater/groundwater_extraction.py
```diff
@@ -440,6 +492,7 @@ def extract_groundwater(
             if found_groundwater:
                 found_groundwaters.append(found_groundwater)
 
+        found_groundwaters = self.merge_weak_candidates(found_groundwaters)
```
stijnvermeeren-swisstopo: We have a risk here that we merge two measurements together that at this point seem compatible, but later on would become incompatible after we take into account the detected borehole elevation (in `borehole.filter_groundwater_entries()` at the bottom of `extract.py`).

As a hypothetical example, if we have extracted a measurement "4 m" and a second measurement with a "460 müM" elevation, then these would currently be treated as compatible and merged. However, it might well be that the borehole elevation is then detected as "470 müM", meaning that the two measurements are actually different (with the second one having a depth of "10 m"). On the other hand, if the borehole elevation would have been "464 müM", then this would confirm that the measurements are actually compatible.

In order to be able to take all of this into account, it might be better to also move this method to a point later on in the pipeline, possible as part of the `borehole.filter_groundwater_entries()` method?

stijnvermeeren-swisstopo (changes requested):
Thanks for these useful insights. Looking at the font colour is a great idea that I had not yet considered. I also was not aware that we can extract font color directly from the PyMuPDF text extraction (at least for digitally-born PDFs, for OCRd PDF files this will most likely not work).

It's nice to see a significant benchmarking improvement with the new implementation. I think that the changes all make sense, though there are a few implementation details that I would propose to do differently. (See code comments below.)

Regarding the insights from the PR descriptions:
- `268125539-bp.pdf` has a second colour match that is actually a false positive ("Durchlässigkeit bei Pumpversuch"). Maybe we could try to detect that the unit here is not "m" but "m/s"? However, I'd also be happy to just accept this small extraction error for now.
- You mention that there is a "line overlaps the label" check that does not work as expected for `683251057-bp.pdf` and `683251058-bp.pdf`. These two files in particular would not benefit from removing this check though, as the correct value is already extracted from elsewhere on the profile. Still, did you check if we can remove this check without any negative side-effects (or maybe even with some positive side-effects) on other inputs?
- I believe that it is intentional that we look vertically and horizontally but not diagonally from the key. Most likely, we already realized that there are too many false positives that come into play when we also look diagonally. Are you aware of any concrete examples that would benefit from looking diagonally?
- The large number of files in the categories "wrong_value_near_same_key" and "correct_value_near_different_key" suggest to me that in order to achieve further improvements, we might need to become stricter when extracting potential depth/elevation values near a groundwater key or symbol, to avoid false positives already at that stage. Maybe we should define a follow-up issue to look into that idea?

src/extraction/features/groundwater/groundwater_extraction.py
AgathaSchmidt: Yes, you are right, especially since we are trying to extract three values: data, elevation and depth, it makes sense to make merging of more than two fields possible 
I tried to implement this and was able to improve extraction for 267123078 were we now also extract the two elevations

stijnvermeeren-swisstopo:
Regarding the "line overlaps the label" check, this is necessary to avoid false positive groundwater measurements due to false positives in the symbol detection. For example, without this check, we would extract groundwater level like these (`267125047-bp.pdf`, `269125064-bp.pdf`, `684256005-bp.pdf`), because some of the lines in the plot / strip log somehow fulfill the conditions for the groundwater symbol detection.

image
image
image

Without the check, these false positive decrease the overall groundwater F1 score by ca. 4%.

Conclusion: we cannot remove or relax this condition, before we have made the detection of the groundwater symbol more robust. However, this is clearly beyond the scope of this PR, and not a high priority. In the meantime, leaving the existing "line overlaps the label" check without changes is fine.

src/extraction/features/groundwater/groundwater_extraction.py
AgathaSchmidt: **raw predictions (before any merging)** 

267125223-bp.pdf:
candidate 1: date=1982-02-18, depth=3.93, elevation=None    
candidate 2: date=None,       depth=None, elevation=434.9

268124635-bp.pdf:
candidate 1: date=None,       depth=10.0, elevation=None   
candidate 2: date=1970-04-28, depth=None, elevation=600.0 

681249142-bp.pdf:
candidate 1: date=1966-06-23, depth=None, elevation=None
candidate 2: date=None,       depth=1.4,  elevation=400.0  

**267125223** 

ground truth

```json
            "date": "1982-02-18",
            "depth": 3.93,
            "elevation": 434.9,
```

predicted 

```json
        "groundwater": [
          {
            "date": null,
            "depth": 3.28,
            "elevation": 434.9,
          },
          {
            "date": "1982-02-18",
            "depth": 3.93,
            "elevation": 434.25,
```

old branch 

```json
            "date": "1982-02-18",
            "depth": 3.93,
            "elevation": 434.9,
```

**268124635**

- ground truth

```json
                    "date": "1970-04-28",
                    "depth": 6.0,
                    "elevation": 407.62
```

- predicted

```json
            "date": "1970-04-28",
            "depth": -186.38,
            "elevation": 600.0,
          },
          {
            "date": null,
            "depth": 10.0,
            "elevation": 403.62,
```

predicted (old branch) 

```json
            "date": "1970-04-28",
            "depth": 10.0,
            "elevation": 600.0,
```

681249142

- groundtruth

```json
                    "date": "1966-06-07",
                    "depth": 3.0,
                    "elevation": 397.24
```

- predicted borehole 0 (borehole 1, 3, 4) nothing

```json
            "date": null,
            "depth": 1.4,
            "elevation": 400.0
```

- predicted borehole 2

```json
            "date": "1966-06-23",
            "depth": null,
            "elevation": null,
```

- predicted old branch : for borehole 2  (other boreholes hat nothing)
- 

```json
            "date": "1966-06-23",
            "depth": 1.4,
            "elevation": 400.0,
```

Summary

For 267125223 and 268124635 we have a similar root cause.
We have two gw readings. And before, we were merging them into one which delivered the correct prediction for 267125223 and an incorrect, but single prediction for 268124635.

Since the terrain elevation is known now it is computing the implied depth. In the case of 267125223 the terrain elevation used for this is a different elevation the ground truth gw data is based on. We get conflicting depths—> 3.93 as extracted (gt, correct) and 3.28 as implied by terrain elevation (wrong). The mismatch is treated as an conflict. The merge is blocked and both predictions are kept and filled with via infer_infos
Same thing is happening for 268124635

For 681249142 we have a different root cause. The terrain elevation is not known, so the implied depth cannot be the reason for it. Before we implemented the current behavior, the extracted gw data got merged into one entry before borehole-assignment ran, so that single combined entry got assigned to borehole 2.

After implementing the changes: merging is deferred until after borehole-assignment, so each small, still-separate raw candidate gets independently matched to whichever borehole its own rect happens to be closest to —> the date mention lands in borehole[2] (correct-ish), but the depth+elevation mention lands in borehole[0] instead (wrong borehole). By the time `filter_entries` runs per-borehole, they're in different buckets and can never be reunited.

All in all, I think that  267125223 and 268124635 are actually behaving as expected. OCR confusion or OKT OKR confusion ar eleading to those results, but generally the methodology behaves as expected 

@stijnvermeeren-swisstopo what do you think?

stijnvermeeren-swisstopo: @AgathaSchmidt  Thanks for the detailed investigation.

I think that in principle, the new implementation is conceptually more sound, but the disadvantage is that mistakes in elevation detection (e.g. confusing OKT and OKR) are more likely to "propogate" into the extracted groundwater measurements and cause additional mistakes there.

Is the new implementation a lot more complex than the old one? (It's hard to tell, because your last commit contains a lot of commented out code.) If the additional complexity is small, then I would accept it, but if there's a lot of additional complexity, then I'd be happy to revert to your original implementation.

AgathaSchmidt: @stijnvermeeren-swisstopo 
I cleaned up the code a bit 
In my opinion the code is not a lot more complex 

Real added complexity: one helper (`implied_depth`, approx. 6 lines) plus one extra branch in `conflicts()` that compares implied depths (approx. 4 lines), plus a small `_group_by_page` helper needed because the merge/overlap logic moved from page to borehole level. Everything else in the diff is pure relocation —> same methods, moved from `GroundwaterLevelExtractor` to `GroundwatersInBorehole`

If we are willing to accept the propagated OKT/OKR/OCR issues I think it is a good idea conceptually 
And overall still a huge improvement on the groundwater metrics

src/extraction/features/groundwater/groundwater_extraction.py
```diff
@@ -267,6 +366,19 @@ def remove_duplicates(self):
...
+
+        Args:
+            terrain_elevation (float): The elevation of the terrain at the top of the borehole.
+            layers (list[Layer]): The list of layers in the borehole.
+        """
+        self.merge_compatible_candidates(terrain_elevation)
+        self.remove_overlaps()
+        self.remove_duplicates()
+        for entry in self.groundwater_feature_list:
+            entry.feature.infer_infos(terrain_elevation, layers, entry.rect)
```
stijnvermeeren-swisstopo, line 380: Can we do
```
        for entry in self.groundwater_feature_list:
            entry.feature.infer_infos(terrain_elevation, layers, entry.rect)
```
first, and only afterwards `self.merge_compatible_candidates(terrain_elevation)`? Then we might not need the `implied_depth` method (which sort-of duplicates the logic from `infer_infos`) inside `merge_compatible_candidates` anymore. Or is there some side-effect that I'm missing?

stijnvermeeren-swisstopo (changes requested):
I'm happy with the new behaviour. There's just one part of the implementation where I'm wondering if a simplification is possible.

stijnvermeeren-swisstopo:
I implemented the change proposed in my last comment, inferring missing depths already before the methods `merge_compatible_candidates`, `remove_overlaps` and `remove_duplicates`.

Additionally:
- `merge_compatible_candidates` is now simplified, no longer requiring information on the borehole elevation
- `remove_duplicates` is completely redundant and was removed from the code, as it essentially tried to do the same thing as `merge_compatible_candidates`, with the only difference being that duplicates were removed rather than merged.

View just these changes: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/541/changes/2384e73d8ef2054c41be67aa7f734d0eba765b49..6ab143186d4cfc94e70b52c3b916924e694f6051

There is a small change in behaviour, related to the fact that `remove_duplicates` used to be called before inferring missing depths, leading to potentially incorrectly identified duplicates (essentially the same issue that I highlighted earlier on in this PR: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/541#discussion_r3767045258).

This change unfortunately decreases the accuracy a little bit (Zurich groundwater F1 -0.4% from 41.2% to 40.8%), since incorrectly extracted groundwater measurements in `267122001-bp.pdf` and `267126034-bp.pdf` are no longer removed. However, this actually seems more correct, since the extracted values were actually not fully compatible, so to treat the as duplicates was not really conceptually correct anyway. The proper solution for these extraction mistakes would be to avoid these false positive values to be extracted in the first place, rather than trying to deal with them in a deduplication step.

@christianabbet would you be happy to review my changes?

christianabbet (approved):
Checked last two commits, LGTM

stijnvermeeren-swisstopo (approved):

Issue #505, opened 2026-06-26
Improve borehole identification, continuation detection and name extraction for 3384.pdf


stijnvermeeren-swisstopo:
The 259-page PDF file 3384.pdf (https://assets.swissgeol.ch/?assetId=3384) contains many borehole profiles, mostly single-page profiles, but also some that span across several PDF pages. It provides a fascinating test case for our extraction pipeline.

For the purposes of this ticket, I'd like to focus on how we can improve those extracted features that also have an impact on the assets classification pipeline:
- borehole identification
- continuation detection
- borehole name extraction

For the scope of this issue, I would ignore the quality of the extracted layers (depths, descriptions). We should also ignore any borehole profiles that are part of a cross section profile (e.g. p24-25).

I've summarized my evaluation of the current data extraction pipeline output in an Excel file: [3384.xlsx](https://github.com/user-attachments/files/29383538/3384.xlsx)

Borehole identification:
- Quite a few false positives in the early parts of the PDF file.
- Two false negatives later on in the file (p211, p252)
- Three pages where both the correct profile as well as a false positive is detected (p240, p241, p249)

Continuation detection (possibly related to https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/466):
- 11 false positives, especially where no depths are detected on one of the pages
- 14 false negatives
- 10 true positives

Borehole name extraction (possibly related to https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/525)
- Name not extracted on p92, p193 and p197

We should aim to improve on these inaccuracies, while avoiding any negative impact the benchmarking datasets.

stijnvermeeren-swisstopo:
- Continuation detection was improved sufficiently in #466 .
- Borehole name detection will be solved as part of #535 .

So for this issue, only the borehole identification part remains.

PR #542, opened 2026-08-10
More robust continuation detection 

### LINKED ISSUE #466: Boreholes spanning consecutive pages are not detected
#  Current status

After reviewing Geoquat and Zurich examples, the majority of consecutive borehole merges are working correctly. However, some boreholes spanning consecutive pages are not being detected at all. Known affected files:

  * ZH:268124125
  * GQ:B147
  * GQ:5115
  * GQ:5111

# Proposal

Investigate the files above to identify any common pattern causing layers to go undetected. Once detection is fixed, verify that detected layers merge correctly across multiple pages.

AgathaSchmidt:
closes #466

The continuation detection error described in #466 were mostly considered out of scope as described in the comment under the issue description. The error are due to OCR errors and bad geometry scores. 

Nevertheless, I wanted to use this issue to address a part of #505 
--> This PR proposed a more robust continuation detection which resolves almost all detected continuation detection problems in the huge 3348.pdf file, while keeping the side effects on zurich and geoquat/val reasonable. 

## Main changes 
- Borehole-name cross-check: extract borehole names per page and attach the nearest one to each extracted borehole; if two boreholes across a page break both have a confidently-detected but disagreeing name, block the merge —> direct evidence they're different boreholes, overriding the text/depth heuristics.
- Fix false-positive merge when depths are missing: `_is_continuation` previously defaulted to merge when neither side had depth values to compare; now it defaults to not merge without evidence.
- Wider overlap detection window: if the plain boundary match fails (e.g. an OCR-truncated first/last layer blocks alignment), retry after dropping up to 2 layers from either end of the current/previous page's layers, with a guard requiring depth corroboration for single-layer matches.
- New depth-reset fallback: when boundary text is unreadable (e.g. obscured by a ruler), detect overlap purely from repeated depth values shared between the two pages.
- Loosened thresholds: `duplicate_layer_threshold` / `material_threshold` lowered 0.95 → 0.90, and partial-overlap acceptance now needs only 1 (not 2) consecutive depth-matched layers.
- Text normalization for comparison: fold accents and collapse punctuation/whitespace before Levenshtein comparison, so independent OCR passes of the same row (accents, comma placement) compare as equal.
- 
### Errors continuation detection

**False positives** 

- 71-72 unmerged
- 108 - 109  unmerged
- 139-140-141 unmerged
- 147-148 unmerged
- 162-163 unmerged
- 169-170 unmerged
- 177-178 unmerged
- 187-188-189  unmerged
- 191-192 unmerged
- 212-213 unmerged ( wrong sidebar detected though fallback which allows merge when the depth in creases )
- 217-218-219-220-221 unmerged
- 240-241 unmerged
- 246-247 unmerged

**False negative** 

- 132 -133 merged
- 135-136 merged
- 138-139 merged
- 149-150 merged
- 161-162 merged
- 165 -166 / 167 - 168 - 169(166 and 167 are the same thing just that on 167 is no ruler anymore) merged
- 179-180 merged
- 181-182 merged
- 186-187 merged
- 195-196 merged

**True positives** 

- 137-138
- 151-152
- 170-171-172
- 173-174
- 176-177-
- 193-194
- 221-222

### Side effects 
Zurich : 
positive: 
- 684252058-bp.pdf —> continuation correctly detected and merged

negative: 
- 267124180-bp.pdf false borehole name discovered on page 2 which reverts the previous correct merge

Geoquat 
positive:
- 10088, B366—> FP continuation removed

negative:
- B147 —> FP continuation detection (page 1 and 2 falsely merged)

Zurich | layer_f1 | depth_intervall_f1 | Material_description_f1 | elevation_f1 | coordinates_f1 | Borehole_name_f1
-- | -- | -- | -- | -- | -- | --
develop | 0.844 | 0.895 | 0.884 | 0.829 | 0.705 | 0.748
branch | 0.847 | 0.895 | 0.886 | 0.829 | 0.705 | 0.741

Geoquat | layer_f1 | depth_intervall_f1 | Material_description_f1 | elevation_f1 | coordinates_f1 | Borehole_name_f1
-- | -- | -- | -- | -- | -- | --
develop | 0.488 |0.636 | 0.6398 | 0.607 |  0.478 | 0.608
branch | 0.497 | 0.635 | 0.648 | 0.607 | 0.476 | 0.603

### DIFF
--- a/src/extraction/core/extract.py
+++ b/src/extraction/core/extract.py
@@ -14,17 +14,18 @@ from extraction.features.groundwater.groundwater_extraction import (
     GroundwaterInDocument,
     GroundwaterLevelExtractor,
 )
-from extraction.features.metadata.borehole_name_extraction import NameInDocument, extract_borehole_names
+from extraction.features.metadata.borehole_name_extraction import BoreholeName, NameInDocument, extract_borehole_names
 from extraction.features.metadata.metadata import FileMetadata, MetadataInDocument
 from extraction.features.predictions.borehole_predictions import BoreholePredictions
 from extraction.features.predictions.file_predictions import FilePredictions
 from extraction.features.predictions.predictions import BoreholeListBuilder
 from extraction.features.stratigraphy.layer.continuation_detection import merge_boreholes
-from extraction.features.stratigraphy.layer.layer import LayersInDocument
+from extraction.features.stratigraphy.layer.layer import ExtractedBorehole, LayersInDocument
 from swissgeol_doc_processing.geometry.geometry_dataclasses import Line
 from swissgeol_doc_processing.geometry.line_detection import extract_lines
 from swissgeol_doc_processing.text.extract_text import extract_text_lines
 from swissgeol_doc_processing.text.matching_params_analytics import MatchingParamsAnalytics
+from swissgeol_doc_processing.utils.data_extractor import FeatureOnPage
 from swissgeol_doc_processing.utils.file_utils import read_params
 from swissgeol_doc_processing.utils.strip_log_detection import StripLog, detect_strip_logs
 from swissgeol_doc_processing.utils.table_detection import TableStructure, detect_table_structures
@@ -36,6 +37,25 @@ table_detection_params = read_params("table_detection_params.yml")
 striplog_detection_params = read_params("striplog_detection_params.yml")
 
 logger = logging.getLogger(__name__)
+
+
+def _assign_borehole_names(
+    extracted_boreholes: list[ExtractedBorehole], name_entries: list[FeatureOnPage[BoreholeName]]
+) -> None:
+    """Attach the closest name candidate found on this same page to each borehole, if any were found.
+
+    A page essentially always has at most a couple of boreholes and name candidates, so matching by
+    vertical distance to the top of each borehole's column on this page is enough.
+
+    Args:
+        extracted_boreholes (list[ExtractedBorehole]): The boreholes just extracted from this page.
+        name_entries (list[FeatureOnPage[BoreholeName]]): The name candidates found on this same page.
+    """
+    if not name_entries:
+        return
+    for borehole in extracted_boreholes:
+        borehole_top = borehole.bounding_boxes[-1].get_outer_rect().y0
+        borehole.name = min(name_entries, key=lambda entry: abs(entry.rect.y0 - borehole_top))
 
 
 @dataclasses.dataclass
@@ -153,6 +173,7 @@ def extract(
                 analytics,
                 **matching_params,
             ).process_page()
+            _assign_borehole_names(extracted_boreholes, name_entries)
             boreholes_per_page.append(extracted_boreholes)
 
             # Extract the groundwater levels
--- a/src/extraction/features/stratigraphy/layer/continuation_detection.py
+++ b/src/extraction/features/stratigraphy/layer/continuation_detection.py
@@ -15,6 +15,21 @@ DEPTHS_QUANTILE_SLACK = 0.1
 MIN_NAME_CONFIDENCE = 0.5
 
 
+def _confident_names(borehole_a: ExtractedBorehole, borehole_b: ExtractedBorehole) -> tuple[str, str] | None:
+    """Return both boreholes' normalized printed names, if each was confidently detected, else None."""
+    if not (
+        borehole_a.name
+        and borehole_b.name
+        and borehole_a.name.feature.confidence >= MIN_NAME_CONFIDENCE
+        and borehole_b.name.feature.confidence >= MIN_NAME_CONFIDENCE
+    ):
+        return None
+    return (
+        _normalize_for_comparison(borehole_a.name.feature.name),
+        _normalize_for_comparison(borehole_b.name.feature.name),
+    )
+
+
 def _names_conflict(borehole_a: ExtractedBorehole, borehole_b: ExtractedBorehole) -> bool:
     """Check whether both boreholes have their own confidently detected name, and those names disagree.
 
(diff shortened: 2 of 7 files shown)


christianabbet (approved):
Detection on 3384 look really good. works well on my side. Minor comments, already accepted

Small comment, while I was checking the files:
* 137 is merged to 138 AND 139 on my side

AgathaSchmidt:
Thanks Christian!

Yes that is true
137-138 were merged already without any code changes but our pipeline was failing to merge 138-139 which is why you can find 137-138 in "true positives" and 138-139 in "false negatives" with the comment that they have been successfully merged after the changes

src/extraction/features/stratigraphy/layer/overlap_detection.py
```diff
@@ -10,6 +12,10 @@
 
 logger = logging.getLogger(__name__)
 
+MAX_BOUNDARY_LAYERS_TO_DROP = 2
+DEPTH_VALUE_TOLERANCE = 0.05
```
stijnvermeeren-swisstopo: Isn't this a bit large? I would expect that at most, we need some tolerance for floating point arithmetic errors. But if we have values that are 5cm apart, then they should probably be treated as being unequal for real? Or do you have a concrete example where we really need this much tolerance?

src/extraction/core/extract.py
```diff
@@ -38,6 +39,25 @@
 logger = logging.getLogger(__name__)
 
 
+def _assign_borehole_names(
```
stijnvermeeren-swisstopo, line 42: I'm not a huge fan of assigning borehole names for boreholes twice with two different methods.

Do you think that it might make sense to always match the all of the metadata to the extracted stratigraphy already on a page-by-page basis, before doing any continuation detection? I.e. use more or less the existing logic from the `BoreholeListBuilder` class, but call it for each page separately inside the `for page_index, page in enumerate(doc)` loop of the `extract()` method, similar to how you currently call this `_assign_borehole_names` method?

I think that the fact that you needed to implement this method, shows that the current matching between boreholes and metadata is not really done in the right place. But changing this would probably be a bigger refactoring that might be best done as a separate PR.

src/extraction/features/stratigraphy/layer/continuation_detection.py
```diff
@@ -5,10 +5,40 @@
...
 DEPTHS_QUANTILE_SLACK = 0.1
+MIN_NAME_CONFIDENCE = 0.5
+
+
+def _names_conflict(borehole_a: ExtractedBorehole, borehole_b: ExtractedBorehole) -> bool:
+    """Check whether both boreholes have their own confidently detected name, and those names disagree.
+
+    A borehole's own printed name is stronger evidence of identity than the layer text or depth
+    heuristics used elsewhere in this module: if both pages name their borehole and the names don't
+    match, they must be different boreholes, regardless of what those other heuristics conclude.
```
stijnvermeeren-swisstopo, line 38: Would it make sense to also do the inverse logic? I.e. if both borehole have an identified name, and it it different, then we automatically treat it as a continuation, even if other criteria fail? This could potentially allow us to still treat the 3 pages of `268124125-bp.pdf` as the same borehole, since each page mentions the name "RKB 17-3". The same for `A435.pdf`/`A437.pdf`, where both pages mention the name "KB 10"/"KB 12".

stijnvermeeren-swisstopo (changes requested):
I will have to continue the review tomorrow, but here are some intermediate comments.

src/extraction/features/stratigraphy/layer/overlap_detection.py
```diff
@@ -118,6 +124,129 @@ def find_split_by_convolution(
...
+        OverlapResult | None: indices that define the overlapping layers, or None if no overlap.
+    """
+    prev_values = _collect_depth_values(layers_prev)
+    if not prev_values:
+        return None
+    prev_max = max(prev_values)
+
+    lower_id = 0
+    matches = 0
+    for layer in layers_curr:
```
stijnvermeeren-swisstopo: Could you just write
```
for lower_id, layer in enumerate(layers_curr):
```
here, instead of defining and incrementing `lower_id` separately? It would be less verbose and it would make the connection between `layer` and `lower_id` more explicit.

src/extraction/features/stratigraphy/layer/overlap_detection.py
```diff
@@ -118,6 +124,129 @@ def find_split_by_convolution(
...
+        if result is not None and _is_trustworthy_retry_match(result, trimmed_curr):
+            return OverlapResult(upper_id=result.upper_id, lower_id=result.lower_id + drop)
+
+    for drop in range(1, MAX_BOUNDARY_LAYERS_TO_DROP + 1):
+        if len(layers_prev) <= drop:
+            break
+        trimmed_prev = layers_prev[:-drop]
+        result = _find_longest_overlap(trimmed_prev, layers_curr, matching_params)
+        if result is not None and _is_trustworthy_retry_match(result, layers_curr):
+            return result
```
stijnvermeeren-swisstopo, line 163: A small explanatory comment could be beneficial here, in particular to explain why here we return `result`  directly, while in the very similar code at line 152, we add `drop` to the `lower_id`  before returning.

stijnvermeeren-swisstopo (changes requested):
I don't have much to add to my initial review from yesterday.

Overall, I'm very impressed by the improvements on the 3348.pdf file, good job!

One small remark regarding the PR description: page 200-201 does not look like a false positive to me. It appears to me that these are actually two separate scans of the exact same page (probably done by accident). Since the borehole profile is identical, it does make sense to me to only extract it once as a single borehole. In other words: I'm satisfied with the behaviour of this PR also on those pages.

src/extraction/features/stratigraphy/layer/overlap_detection.py
stijnvermeeren-swisstopo: After having thought about this a little bit more, I would propose a short-term fix (for this PR) and a long-term fix (for a new issue).

Short-term fix: unless you see any negative side-effect, I would propose to remove this constant. I hope that it should be sufficient to just use `math.is_close` with the default tolerance, and for the inequality we could use `end > prev_max and not math.isclose(end, prev_max)` instead of `end > prev_max + DEPTH_VALUE_TOLERANCE`.

Long-term fix: I would propose to change the data type of all values that are inherently exact (e.g. extracted depths and elevations) to decimal (https://docs.python.org/3/library/decimal.html). Then (in)equalities should work out-of-the-box without having to worry about precision/tolerance. We should keep using float as a data type for all values that are inherently inexact (e.g. PDF page coordinates).

AgathaSchmidt: yes thats possible, with no negative effect on 3348, zurich and geoquat/val

src/extraction/core/extract.py
AgathaSchmidt: @stijnvermeeren-swisstopo 
If we do page-per-page basis, would it be okay for you to loose the metadata for some cases, where the metadata information is not on the same page as the borehole ? 
i.e. loosing borehole name for zurichs A42 and elevation data for A11452.pdf and groundwater information for 8366.pdf 
(This is what a current alternative approach yields)

src/extraction/features/stratigraphy/layer/continuation_detection.py
AgathaSchmidt: I tried to implement this and it works for `268124125-bp.pdf` with no other side effects (committed, but unpretty code --> due to the comment you mentioned before, about how we are accessing borehole name. Can probably be fixed as soon as we agree if we accept the downsides of the per-page matching)

Aren't `A435.pdf`/`A437.pdf` already merged on the develop branch?

src/extraction/core/extract.py
stijnvermeeren-swisstopo: @AgathaSchmidt good insights. Then maybe we should allow assigning data from adjacent pages to a borehole, as long as there is no other borehole detected on those adjacent pages?

Either way, as I said, I would not include such changes in this PR. I'm quite happy to accept this PR with the two different matching methods, as long as we write down these ideas for future improvements in a follow-up issue.

src/extraction/features/stratigraphy/layer/continuation_detection.py
stijnvermeeren-swisstopo: > Aren't A435.pdf/A437.pdf already merged on the develop branch?

On develop yes, but no longer with the initial implementation of this PR. Indeed, you actually reported these documents in the PR description yourself as having negative side effects.

src/extraction/features/stratigraphy/layer/overlap_detection.py
stijnvermeeren-swisstopo: I created an issue for the long-term fix idea: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/549

src/extraction/features/stratigraphy/layer/continuation_detection.py
AgathaSchmidt: True, I missed that those are the same files I reported above

src/extraction/features/stratigraphy/layer/overlap_detection.py
AgathaSchmidt: I can and I did now. 
Only thing to mention: from how I understand it, a small noqa comment is necessary because `lower_id` is not used directly in the for-loop which causes a warning

src/extraction/features/stratigraphy/layer/overlap_detection.py
AgathaSchmidt: Thanks!

src/extraction/core/extract.py
AgathaSchmidt: created #550  
Feel free to adjust the issue description if needed

AgathaSchmidt:
Updated PR description after the changes based on the comments from @stijnvermeeren-swisstopo 
--> Less negative side effects on zurich and geoquat overall

src/extraction/features/stratigraphy/layer/overlap_detection.py
stijnvermeeren-swisstopo: Ah, I missed that we need to handle the case where `layers_curr` might be empty, and we still need an initialized value of `lower_id` after the for loop. This makes it a little more verbose than what I was hoping for. Still, I have a slight preference for the new implementation. 👍

stijnvermeeren-swisstopo (approved):
Looks good to me!

Issue #525, opened 2026-07-21
Improvements for borehole name extraction on SBB profiles


stijnvermeeren-swisstopo:
Borehole ID 45341: name "S1" not extracted because OCR misreads it as "SI".

> <img width="493" height="140" alt="Image" src="https://github.com/user-attachments/assets/6901020f-b2a3-47e3-b96c-304f31def614" />

Borehole ID 45291: name extracted as "KB3 1 50" instead of "KB3", even though the OCR correctly reads "KB3 1 : 50"

> <img width="583" height="84" alt="Image" src="https://github.com/user-attachments/assets/5996bb9b-3772-47a7-8ba5-7dc3cdd7d4b6" />

Borehole ID 45295: name extracted as "Kb" instead of "Kb 02/4"

> <img width="337" height="79" alt="Image" src="https://github.com/user-attachments/assets/753eac03-0e5c-4b24-b7d5-8f9539d7bdd0" />

Borehole ID 45299: name "RKS3-13" not extracted at all

> <img width="235" height="99" alt="Image" src="https://github.com/user-attachments/assets/1de76654-5da1-4974-8405-062b321d2d11" />

Borehole ID 45331: name extracted as "-" instead of "B2"

> <img width="665" height="94" alt="Image" src="https://github.com/user-attachments/assets/9c6b2ba2-0d0c-4e38-8cc4-8e9e3826f4a7" />

Borehole ID 45410: "Anhang B2" misidentified as borehole name (instead of "KB23")

> <img width="284" height="146" alt="Image" src="https://github.com/user-attachments/assets/46eb4d7c-34d4-45c4-88df-aec7e79637a2" />

Borehole ID 45409 (and many more): extraction fails because PDF rotated by 90°

> <img width="591" height="797" alt="Image" src="https://github.com/user-attachments/assets/d1d4ad37-7552-4576-99e1-fd260fa3f3fe" />

Borehole ID 45456: Borehole name "RB 1/01/P" missed

> <img width="448" height="125" alt="Image" src="https://github.com/user-attachments/assets/1cbffcbf-518f-4b1e-8705-51225173940a" />

Borehole ID 45507: name "KB2" missed

> <img width="569" height="118" alt="Image" src="https://github.com/user-attachments/assets/86a0b358-10cf-4cba-b4e2-c3730889d151" />

Borehole ID 45685: name "KB 14" missed

[...]

PR #543, opened 2026-08-11
Smarter splitting of TextLines based on vertical lines in detected table structures

stijnvermeeren-swisstopo:
The legacy method of splitting `TextLine`s based on the `is_line_start` method was intended to detect where text was extracted as a single text line from a PDF file, even though the correct interpretation would be to treat the words as two or more adjacent lines.

E.g. in this example (ZH `267122002-bp.pdf`), the numbers 85 and 10 should ideally be treated separately, and not considered to be part of the line `"... m.10%Steinen,komp.8510"`.

image

The existing logic was purely based on the positioning of the text lines, and did not work in a very robust way. For example, it splits text lines too aggressively on several Geoquat dataset inputs (see improvements listed below).

Moreover, sometimes this would split the borehole name into two lines, preventing accurate name extraction. For example, this affects two SBB profiles from https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/525, namely `45295.pdf` (where "Kernbohrung Kb 02/4" used to be split into "Kernbohrung Kb" and "02/4") and `45456.pdf` (where "RB 1/01/P" used to be split into "RB" and "1/01/P").

We replace it with a new logic, where text lines are only split later on, based on the detected structure lines from the table-like structures.

## Dealing with grid lines

The main challenge with this approach are documents that contain grid lines, which might lead to excessive splitting. Examples:
- ZH `267125223-bp.pdf`
- ZH `267125334-bp.pdf`
- ZH `267125338-bp.pdf`
- ZH `267125339-bp.pdf`
- ZH `268124125-bp.pdf`
- ZH `268124375-bp.pdf`
- ZH `677259003-bp.pdf`

The proposed implementation solves most of these cases by rejecting vertical lines that clearly intersect words, already when constructing the table-like structures.

There are still many grid lines that become part of `TableStructure` objects, and it would be cleaner to somehow detect and remove all of them. However, that is beyond the scope of this issue. For the time being, I would say that the proposed approach works well enough.

## Impact on runtime

 While we reduce the computation that is necessary in `extract_text.py`, we increase the computation required to construct `TableStructures`, as well as adding a new method `_split_text_lines_by_table_structures` to `extract.py`.

This involves a nested loop to check for intersections between words and vertical lines. I was able to make this sufficiently efficient by using an RTree structure for efficient spacial indexing of text lines in `table_detection.py`. This is not entirely new, it was already used elsewhere (e.g. for computing the sidebar noise count), but I've now created a dedicated `TextLineRTree` wrapper class for it, in order to make usage slightly easier.

Overall, on my machine, this seems to make the entire pipeline slightly faster if anything, thought the change is hardly significant.

## Impact on benchmarking

The proposed implementation has only a small impact (in the region of +/- 0.1%) on the benchmarking metrics.

If it were not for two other issues that interfere on a few files, there would have been a measurable improvement on the Geoquat benchmark.

In detail:

**Zurich**
- Minor decrease of material description F1 on `268124125-bp.pdf` (incorrect split due to grid lines)
- Minor decrease of groundwater F1 on `268124635-bp.pdf` (additional incorrect groundwater measurement)

Line splitting behaviour is preserved without changes on `267122002-bp.pdf`. 

**Geoquat validation**
- Worse on `12327.pdf` (`AAboveBSidebar` incorrectly selected over `LayerIndicatorSidebar` -> https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/545)
- Better on `1506.pdf` (`LayerIndicatorSidebar` correctly selected)
- Slightly worse on `680.pdf` (incorrect `LayerIndicatorSidebar` selected on page 2 -> https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/545)
- Slightly worse on `7317.pdf` (`AToBSidebar` selected instead of `LayerIndicatorSidebar`, both are reasonable)
- Better on `A11363.pdf` (additional good text line split)
- Better on `A218.pdf` (lost a false positive text line split)
- Better on `A337.pdf` (lost a false positive text line split)
- Better on `A356.pdf` (lost a false positive text line split)
- Better on `A430.pdf` (lost a false positive text line split)
- Better on `A8384.pdf` (low quality scan, not sure what is happening)
- Worse on `AKB_137.pdf` (worse `AAboveBSidebar` selected -> https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/545 and https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/546)
- Worse on `B147.pdf` (incorrect text splits due to grid lines on page 2)
- Better on `B156.pdf` (additional good text line splits)
- Worse on `V13.pdf` (actually we get an additional good text line split, but unfortunately this makes space for an entirely new false positive borehole -> https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/546 )
- Slightly worse on `V69.pdf` (a useful text line split is now no longer applied)

Line splitting behaviour is preserved without changes on several additional borehole profiles, including `A11300.pdf`, `A11363.pdf`, `A11394.pdf`, `A11436.pdf`, `A1183.pdf` and `A263.pdf`.

### DIFF
--- a/src/swissgeol_doc_processing/text/textline_rtree.py
+++ b/src/swissgeol_doc_processing/text/textline_rtree.py
@@ -1,4 +1,4 @@
-"""This module provides an efficient spacial index for text lines."""
+"""This module provides an efficient spatial index for text lines."""
 
 import fastquadtree
 import pymupdf
--- a/src/swissgeol_doc_processing/text/textline_rtree.py
+++ b/src/swissgeol_doc_processing/text/textline_rtree.py
@@ -1,0 +1,22 @@
+"""This module provides an efficient spacial index for text lines."""
+
+import fastquadtree
+import pymupdf
+
+from swissgeol_doc_processing.text.textline import TextLine
+
+
+class TextLineRTree:
+    """Small wrapper around fastquadtree.RectQuadTreeObjects for text lines collections."""
+
+    def __init__(self, text_lines: list[TextLine]):
+        min_x = min([line.rect.x0 for line in text_lines])
+        max_x = max([line.rect.x1 for line in text_lines])
+        min_y = min([line.rect.y0 for line in text_lines])
+        max_y = max([line.rect.y1 for line in text_lines])
+        self.text_line_rtree = fastquadtree.RectQuadTreeObjects((min_x, min_y, max_x, max_y), capacity=8)
+        for line in text_lines:
+            self.text_line_rtree.insert((line.rect.x0, line.rect.y0, line.rect.x1, line.rect.y1), obj=line)
+
+    def query(self, rect: pymupdf.Rect) -> list[TextLine]:
+        return [item.obj for item in self.text_line_rtree.query((rect.x0, rect.y0, rect.x1, rect.y1))]
--- a/src/swissgeol_doc_processing/text/textline_rtree.py
+++ b/src/swissgeol_doc_processing/text/textline_rtree.py
@@ -23,4 +23,5 @@ class TextLineRTree:
             self.text_line_rtree.insert((line.rect.x0, line.rect.y0, line.rect.x1, line.rect.y1), obj=line)
 
     def query(self, rect: pymupdf.Rect) -> list[TextLine]:
+        """Efficiently finds all text lines whose bounding box intersects with the given rectangle."""
         return [item.obj for item in self.text_line_rtree.query((rect.x0, rect.y0, rect.x1, rect.y1))]
--- a/src/swissgeol_doc_processing/text/textline_rtree.py
+++ b/src/swissgeol_doc_processing/text/textline_rtree.py
@@ -10,11 +10,15 @@ class TextLineRTree:
     """Small wrapper around fastquadtree.RectQuadTreeObjects for text lines collections."""
 
     def __init__(self, text_lines: list[TextLine]):
-        min_x = min([line.rect.x0 for line in text_lines])
-        max_x = max([line.rect.x1 for line in text_lines])
-        min_y = min([line.rect.y0 for line in text_lines])
-        max_y = max([line.rect.y1 for line in text_lines])
-        self.text_line_rtree = fastquadtree.RectQuadTreeObjects((min_x, min_y, max_x, max_y), capacity=8)
+        if text_lines:
+            min_x = min([line.rect.x0 for line in text_lines])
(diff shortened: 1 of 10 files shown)


src/swissgeol_doc_processing/text/textline_rtree.py
```diff
@@ -0,0 +1,26 @@
...
+            min_y = min([line.rect.y0 for line in text_lines])
+            max_y = max([line.rect.y1 for line in text_lines])
+            bounds = (min_x, min_y, max_x, max_y)
+        else:
+            bounds = (0, 0, 1, 1)
+        self.text_line_rtree = fastquadtree.RectQuadTreeObjects(bounds, capacity=8)
+        for line in text_lines:
+            self.text_line_rtree.insert((line.rect.x0, line.rect.y0, line.rect.x1, line.rect.y1), obj=line)
+
+    def query(self, rect: pymupdf.Rect) -> list[TextLine]:
```
AgathaSchmidt, line 25: Maybe add small docstring

AgathaSchmidt (approved):
I have more differences than the reported ones. 
For zurich, 267122002 differs from develop, the FP number is removed thanks to the better splitting like for 675230002 and 268124635 as well

Also, I found some negative effects which were not reported 

267125334 —> stark sandiger | Kies —> unnecessary split 
267125239 —> include  "Kote OK Terrain" into elevation box 
268124125 —> falsely cutting off “modrig” because there is a line (writer did not respect lines when writing down) --> in my opinion this can be considered out of scope 

But generally speaking, I think the new proposed approach is sound and an improvement

stijnvermeeren-swisstopo:
@AgathaSchmidt 

- I did not report `267122002` and `675230002` because the evaluation already treated all of the descriptions as correct in the old implementation as well, but indeed, the new behaviour is actually better.
- For `268124635`, I did report the change in groundwater extraction in the PR description. I don't immediately see any other changes? 
- I did not report `267125334` because the split has no impact at all on our accuracy metrics. You are of course correct that it would be cleaner not to split the line. For the sake of completeness, I will add this in the to the list of documents with challenging grid lines in the PR description.
- The bounding box for the elevation in `267125239` has changed indeed, but again, this has not impact on the accuracy metrics, and this type of bounding box is common in other files as well, so I would rank this as a "neutral" change.
- I actually did report `268124125` already in the PR description?

PR #544, opened 2026-08-11
Extract line breaks in material descriptions

### LINKED ISSUE #524: Extract line breaks in material descriptions
Example:

Image

In this case, the word "Fremdkomponenten" should ideally be preceded by a line break.

However, wrapped text should not be extracted with line breaks at the end of every line!

It might not always be possible to differentiate between a line break and a wrapped line, but in most cases, we should at least be able to make a good guess.

Example:

Image

The description _"Feinsand, siltig, kiesig, mit wenig org. Beimengungen, mit Betonbruchstücken bis ø 12 cm, oft angerostet, geruchlich unauffällig, mit Kalkausfällungen, braun-grau"_ should be extracted as a single wrapped line, not as two separate lines. Ideally, we could also re-construct the hyphenated word "angerosted" from the parts "ange-" and "rosted", but this could become complicated and may be deferred to a follow-up issue.

AgathaSchmidt:
Closes #524 

## Changes
 The field `text` in `predictions.json` now includes exlicit line breaks `\n`
(I am not sure if the `\n` line break is "renderable" by the application)

**What was already there:** 
- A line break is inserted between two consecutive description lines whenever the earlier line ends in a page change, a :/;, or a . 

**New:** 
- a point is only a line break indicator if that isn't part of a known abbreviation (ca., bzw., etc.) —> these are treated as unconditional, forced breaks. Otherwise, it's inserted if that line's width falls more than 30% short of the widest line seen anywhere in the same borehole, on the assumption that a genuinely full line would come close to that borehole's longest observed line and only a "short" one signals an intentional break rather than a layout wrap.
- a vertical gap between tow lines as big as 0.5x line height also indicates a line break 
The vertical gap catches some examples which would have remain undetected otherwise, e.g. 

**267123080**

"Mittelkies, sandig, schwach feinkiesig, schwach siltig, mit vereinzelten Steinen, grau bis bunt, feucht bis nass (Feinsand-/Siltlinsen: 3.9 m - 4.0 m, 4.45 m - 4.5 m) Wasserzutritte ins Bohrloch ab 3.8 m bis 4.8 m”

Mittelkies, sandig, schwach feinkiesig, schwach siltig, mit vereinzelten Steinen, grau bis bunt, feucht bis nass (Feinsand-/Siltlinsen: 3.9 m - 4.0 m, 4.45 m - 4.5 m)\nWasserzutritte ins Bohrloch ab 3.8 m bis 4.8 m

**267122002**

“Kiessand mit Steinen, z.T. leicht siltig, kompakt\n17.3 - 17.7 siltreiche Lage (0226) 21.0 - 23.0 leicht siltig (0127)”

"Kiessand mit Steinen, z.T. leicht siltig, kompakt\n17.3 - 17.7 siltreiche Lage (0226)\n21.0 - 23.0 leicht siltig (0127)",

**268125447**

"beiger, schwach lehmiger Silt-Feinsand, wenig Kies grauer, schwach siltiger Kies mit reichlich Sand, vereinzelt Steine”

"beiger, schwach lehmiger Silt-Feinsand, wenig Kies\ngrauer, schwach siltiger Kies mit reichlich Sand, vereinzelt Steine”

### DIFF
--- a/src/app/common/schemas.py
+++ b/src/app/common/schemas.py
@@ -531,7 +531,7 @@ class LayerMaterialDescriptionSchema(BaseModel):
         cls, prediction: MaterialDescription, pdf_img_scalings: list[tuple[float]]
     ) -> "LayerMaterialDescriptionSchema":
         return cls(
-            text=prediction.text,
+            text=prediction.text_with_line_breaks,
             bounding_boxes=[
                 BoundingBoxWithPage(
                     page_number=line_feature.page_number,
--- a/src/app/common/schemas.py
+++ b/src/app/common/schemas.py
@@ -531,7 +531,7 @@ class LayerMaterialDescriptionSchema(BaseModel):
         cls, prediction: MaterialDescription, pdf_img_scalings: list[tuple[float]]
     ) -> "LayerMaterialDescriptionSchema":
         return cls(
-            text=prediction.text_with_line_breaks,
+            text=prediction.text,
             bounding_boxes=[
                 BoundingBoxWithPage(
                     page_number=line_feature.page_number,
--- a/src/extraction/annotations/draw.py
+++ b/src/extraction/annotations/draw.py
@@ -250,6 +250,13 @@ class PageDrawer:
                     fill=pymupdf.utils.getColor(color),
                     width=0,
                 )
+                if line.feature.text.endswith("\n"):
+                    self.shape.draw_line(
+                        line.rect.top_right * self.page.derotation_matrix,
+                        line.rect.bottom_right * self.page.derotation_matrix,
+                    )
+                    self.shape.finish(color=pymupdf.utils.getColor("blue"), stroke_opacity=0.5, width=2)
+
                 self._draw_correctness_line(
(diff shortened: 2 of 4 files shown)


stijnvermeeren-swisstopo:
@AgathaSchmidt when you write

> What was already there:
> - A line break is inserted between two consecutive description lines whenever the earlier line ends in a page change, a :/;, or a .

...then where was this implemented? Do you think it is a good idea to do this at all? Most false positives seem to be coming from this (e.g. abbreviations that are not defined in the code yet) and also the color or semicolon doesn't always mean that it should be followed by a line break. Might it be better or ignore punctuation altogether when deciding whether to insert line breaks of not?

AgathaSchmidt:
@stijnvermeeren-swisstopo  Apparently I must have confused something, because you are right, I added `ends_with_break_punct` and `ends_with_abbreviation` in this PR 
And you are also right, that we can delete those and from what I saw this removed some FP and does not play a role in detecting real line breaks

stijnvermeeren-swisstopo (approved):
Nice, it's a simple approach but it already behaves really well in almost all cases. Good job!

PR #557, opened 2026-08-17
smarter matching of extracted metadata/groundwater measurements with specific boreholes.

stijnvermeeren-swisstopo:


stijnvermeeren-swisstopo:
Will be reviewed together with https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/558, since these changes are too inter-dependent to be evaluated separately.

PR #559, opened 2026-08-18
Consider position when evaluating layers

### LINKED ISSUE #528: Incorrect matching between extracted boreholes and ground truth in 681249142-bp.pdf
I observed that during the evaluation/scoring for the the ZH file `681249142-bp.pdf`, the extracted borehole for "104" is actually incorrectly compared with the ground truth for borehole "105". This makes the accuracy metrics for the ZH dataset slightly worse than they should be.

We should check (in the method `LayerEvaluator.match_boreholes_to_ground_truth`) why this happens, and if we can fix it.

AgathaSchmidt:
Closes #528 

## Problem description

LayerEvaluator.match_boreholes_to_ground_truth (layer_evaluator.py:210-274) is purely content-based and position-blind — >it never looks at where a borehole physically sits on the page

```
pred (physically "104") vs GT "104": score = 0.429
pred (physically "104") vs GT "105": score = 0.543   ← wins, wrongly
```

**Extracted (predicted borehole, physically sitting under the printed "104" label):**

```
[0] (0.0–0.2)   'Humus Sand Ikies 22g siltiger Sand'
[1] (0.2–7.0)   'Sand/Kies'
[2] (7.0–12.0)  'Kies'
```

Only 3 layers were extracted — GT "104" actually has 5, so the extraction merged/lost 2 real layers (evident in that garbled first entry, which looks like it swallowed multiple original lines).

**Ground truth "104" (correct match, 5 layers):**

```
[0] (0.0–0.6)   'Humus'
[1] (0.6–1.2)   'Sand / Kies'
[2] (1.2–2.7)   'siltiger Sand'
[3] (2.7–7.0)   'Sand / Kies'
[4] (7.0–12.0)  'Kies'
```

**Ground truth "105" (wrong match, 4 layers):**
```
[0] (0.0–1.4)   'Humus mit Sand und Kies'
[1] (1.4–2.9)   'Sand, sandiger Silt'
[2] (2.9–7.0)   'sandiger Kies'
[3] (7.0–12.0)  'Kies'
```

Since 105 has Humus Sand and Kies in it in has a higher similarity with "Humus Sand Ikies 22g siltiger Sand" and additionally, since it has only 4 layers instead of 5 the denominator is smaller for 105 instead of 104 which leads to a better score of 0.543

## Idea / Solution 
Position could be taken into consideration. 
Fist tries/ implementations assumed a fixed, left to right reading as it is also often in the ground truth, but not always, e.g. 
675245009
    -->  on the page, reading left-to-right, "II" comes first and "3" comes second — but ground truth lists "3" first (index 0) and "II" second (index 1) 

So now, on top of the content similarity, each candidate match gets a small bonus if its position in that reading order lines up with where the ground truth borehole sits in the list. This nudges close, ambiguous calls toward the physically correct pairing.
So the position bonus only acts as a tiebreaker: it can tip a close call, but a genuinely strong content match can still win even if it goes against the expected page order.

This solves the comparison issue in the mentioned file. 
And also the performance increases. Even though `compute_borehole_affinity_and_mapping` was scoring the other borehole to match better, in fact, when choosing the correct borehole we end up with better material description metric.

per file metric 

|metric | before | now |
|---|---|---|
|precision | 0.909 | 0.939 |
|recall | 0.536 | 0.554 |
|f1 | 0.674 | 0.697 |

General metrics - no effects

Zurich | layer_f1 | depth_intervall_f1 | Material_description_f1 | elevation_f1 | coordinates_f1 | Borehole_name_f1 | gw_f1
-- | -- | -- | -- | -- | -- | --| --
develop | 0.846 | 0.894 | 0.885 | 0.829 | 0.699 | 0.746 | 0.372
branch | 0.846 | 0.894 | 0.885 | 0.829 | 0.699 | 0.746 | 0.372

Geoquat | layer_f1 | depth_intervall_f1 | Material_description_f1 | elevation_f1 | coordinates_f1 | Borehole_name_f1 | gw_f1
develop | 0.496 |0.635 | 0.648 | 0.602 |  0.475 | 0.604 | 0.053
branch | 0.496 | 0.635 | 0.648 | 0.602 | 0.475 | 0.604 | 0.053

## Limitations 
Honestly speaking, these changes might be a pretty specific solution for the specific file mentioned --> value `BOREHOLE_ORDER_BONUS_WEIGHT = 0.4 ` was chosen so it works with this file, without harming others

Also, now that we understand the root cause of the problem, we might consider leaving the behavior as it is, which is choosing the borehole which matches best based on text similarity

### DIFF
--- a/src/extraction/evaluation/layer_evaluator.py
+++ b/src/extraction/evaluation/layer_evaluator.py
@@ -9,6 +9,7 @@ import Levenshtein
 from core.benchmark_utils import Metrics
 from core.ground_truth import GroundTruthBorehole, GroundTruthLayer
 from extraction.features.predictions.borehole_predictions import (
+    BoreholePredictions,
     BoreholePredictionsWithGroundTruth,
     FileLayersWithGroundTruth,
 )
@@ -21,6 +22,7 @@ logger = logging.getLogger(__name__)
 
 MATERIAL_DESCRIPTION_SIMILARITY_THRESHOLD = 0.9
 MAX_DEPTH_SCORE = 1.0
+BOREHOLE_ORDER_BONUS_WEIGHT = 0.4  # tune, revisit, if necessary
 
 
 class LayerEvaluator:
@@ -214,7 +216,9 @@ class LayerEvaluator:
         """Match predicted boreholes to ground truth boreholes.
 
         This method compares the predicted boreholes with the ground truth boreholes and establishes a mapping
-            between them based on their similarity.
+            between them based on their similarity, with a bonus for agreeing on page position (reading order)
+            to break close, ambiguous ties --> ground truth is usually, but not always, annotated in reading order
+            (see `BOREHOLE_ORDER_BONUS_WEIGHT`).
 
         Args:
             file_predictions (FilePredictions): all predictions for the file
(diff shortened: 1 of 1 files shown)


stijnvermeeren-swisstopo:
Thanks for the insightful investigation into the issue.

However, I don't feel good about the proposed solution. There is no good reason why the ground truth should list the boreholes in a particular file in any specific order, so I don't like to rely on that.

Instead, I think we should find a better metric for computing the similarity between an extracted borehole and ground truth borehole data. In particular, this metric should be more robust against layers that are incorrectly split and/or merged. I have proposed a possible solution here: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/563. If that PR is accepted, then this one would be redundant and could be closed.

stijnvermeeren-swisstopo:
Closed in favour of the alternative solution from https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/563

Issue #554, opened 2026-08-17
Lithology models have hallucinations for class predictions


christianabbet:
Follow up of #529 

The classification often "guesses" or "hallucinates" values for color or alteration_degree_consolidated (where "moderately_weathered" seems to be treated as some sort of default value), even when the description does not contain any information about these attributes. For example:

   * "Sand", "Silt", "Gravel", "Mergel", "Silt gelblich, mit wenig Feinsand und wenig Kies" get misclassified as "moderately_weathered"
   * "Granit" gets misclassified as "grey"
   * "Lehm" gets misclassified as "brown"
   * "Mergel" gets misclassified as "brown"

Can be related to (#553). See if these are majority classes. Possible solutions are:
* Try to investigate if calibration of outputs can help. 
* Try to balance data before training

christianabbet:
@stijnvermeeren-swisstopo Here is the current status of the classification tasks:

* When #553 is validated, all models will be retrained to include true negatives (out-of-context samples) and keywords (short text definitions linked to each class). This will hopefully removes a large part of the bias in the classification tasks.

* For the "hallucinations", the most likely explanation is a bias in the data itself: "Mergel" frequently appears in the same sentence as "brown". Confirming this would require a more in-depth investigation of the model. One solution would be to curate unbiased examples for each task, but that represents a significant amount of data to prepare. An alternative would be to build mixed samples as augmentation (sentence A + sentence B = label A + label B) in order to decorrelate the values.

My suggestion would be to move this issue to the next phase of the project rather than attempting a quick fix with custom samples.

Issue #553, opened 2026-08-17
Lithology models have poor performance for short texts


christianabbet:
Follow up of #529 

Quite often, an unconsolidated description is misclassified as consolidated. Especially with short inputs, the lithology classification seems rather unreliable. For example:

   * "Gravel" gets misclassified as "marlstone"
   * "Kies" gets misclassified as "conglomerate"
   * "Sand" gets misclassified as "sandstone"
   * "Lehm" gets misclassified as "sandstone"
   * "Silt" gets misclassified as "sandstone"
   * "Granit" gets misclassified as "evaporite"
   * "Silt gelblich, mit wenig Feinsand und wenig Kies" gets misclassified as "siltstone"
   
Following comment on #564, we should pay attention to not returning any consolidation type value at all for empty / irrelevant inputs

See if these are majority classes. Possible solutions are:
* Try to investigate if calibration of outputs can help. 
* Try to balance data before training

PR #564, opened 2026-08-26
feat(api): add classification_tasks field to classify response

### LINKED ISSUE #551: API attribute for lithology type
Follow up of #529 

It would be good to have an attribute that explicitly distinguishes between "consolidated" and "unconsolidated", rather than having to implicitly deduce this from the other attributes that are present in the response.

Add an attribute `type` (or similar) in the answer that says either : `consolidated`, `unconsolidated` or `unknown`

christianabbet:
Closes #551 

## Summary

Adds a new `classification_tasks` field to the classify API response that indicates whether the material was classified as consolidated rock or unconsolidated sediment. This field is determined by the lithology head and helps clients understand which classification fields are populated in the response.

## API

Screenshot 2026-08-26 at 14 15 16

### DIFF
--- a/src/app/api/v1/endpoints/classify_all.py
+++ b/src/app/api/v1/endpoints/classify_all.py
@@ -6,7 +6,7 @@ import logging
 from pathlib import Path
 from typing import TYPE_CHECKING
 
-from app.common.schemas import ClassificationTasksClasses, ClassifyRequest, ClassifyResponse
+from app.common.schemas import ClassificationConsolidationClasses, ClassifyRequest, ClassifyResponse
 from classification.utils.datasets.classification import ClassificationTask
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


stijnvermeeren-swisstopo (changes requested):
`classification_tasks` is not a very intuitive name to me.

I would suggest `consolidation`. So we would have `"consolidation": "consolidated"` or `"consolidation": "unconsolidated"`.

And shouldn't we also allow some _unknown_ value such as `"consolidation": "unknown"` or `"consolidation": null`, which can be used in case the input is e.g. empty of does not contain any geological information that can be analysed (e.g. an input such as "hi my name is Stijn")?

christianabbet:
@stijnvermeeren-swisstopo Right now `consolidation` can be None, but we do not have a system to check confidence if the input is to be trusted. We would need true negatives or calibration for that. This goes more with the model training. Do you want to already implement this logic in this ticket?

I would say that maybe this is something we can tackle with the  #553 or #554 as this is more "model-related"

stijnvermeeren-swisstopo (approved):
@christianabbet yes, let's look at not returning any consolidation type value at all for empty / irrelevant inputs in the context of #553 / #554 .

As far as this issue is concerned, all looks good to me.

PR #565, opened 2026-08-27
feat(classify): remove main class from secondary predictions

### LINKED ISSUE #555: API Primary / secondary components overlap
Follow up of #529 

For input "Siltiger-toniger Kies, mit wechselndem Sandgehalt, braun-braungrau, bindig, trocken", I get "gr" both as en_main and as the first element of en_secondary. This is contrary to the requirements of https://github.com/swisstopo/swissgeol-boreholes-dataextraction/issues/447. ("The possible values for main component are different from the possible values from the secondary component. Still, the classification of main and secondary component is not independent at all. For example, is the main component is already "sand", then "sandy" should not be included among the secondary components.")

Remove primary prediction from list of secondary to avoid conflict. Can be done at the API level

christianabbet:
Closes #555 

Avoid redundancy of main component in secondary

## Before
Screenshot 2026-08-27 at 15 14 11
## After
Screenshot 2026-08-27 at 15 12 44

### DIFF
--- a/src/app/api/v1/endpoints/classify_all.py
+++ b/src/app/api/v1/endpoints/classify_all.py
@@ -7,7 +7,9 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 from app.common.schemas import ClassifyRequest, ClassifyResponse
+from classification.utils.datasets import ExistingClassificationSystems
 from classification.utils.datasets.classification import ClassificationTask
+from classification.utils.datasets.en_main import ENMainSystem
 
 if TYPE_CHECKING:
     from classification.models.model import BertModel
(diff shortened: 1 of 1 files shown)


src/app/api/v1/endpoints/classify_all.py
```diff
@@ -79,6 +81,21 @@ def load_models() -> dict[str, BertModel]:
...
+    """Drop the main class from the secondary predictions to avoid redundancy.
+
+    Args:
+        main: Predicted main lithology class.
+        secondary: Predicted secondary lithology classes (ranked).
+
+    Returns:
+        The secondary classes with the entry matching `main` removed.
+    """
+    return [sec for sec in secondary if sec != main]
```
stijnvermeeren-swisstopo, line 96: With the current implementation, I don't really see the benefit of having a dedicated method for this. The implementation is so simple, that it would be easier to read if it is just inlined in the "Step 3" code below. However, if this method has been created in anticipation of a more complex logic in the future, then we can leave it like this.

stijnvermeeren-swisstopo (approved):
Works as expected.

I have one comment about the implementation, but I'd also be ok with merging the PR as-is, so I'm already approving.

src/app/api/v1/endpoints/classify_all.py
christianabbet: Updated code accordingly

PR #566, opened 2026-09-04
insert single whitespace between wrapped lines (where no newline is inserted)

stijnvermeeren-swisstopo:
Fixes a small mistake in https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/544 : when there is no line break detected, the text lines should be joined with a single whitespace in between, rather that joining the last word of the previous line and the first word of the next line in to a single mangled word.


christianabbet (approved):
LGTM

PR #567, opened 2026-09-07
fix(classify): omit not_specified fields from classify response

### LINKED ISSUE #552: Remove not specified answers from API response
Follow up of #529 

It would be cleaner if we omit attributes that have a value "not_specified" or ["not_specified"] from the response.

* If field is completely empty, do not even return the field at all. Better than putting to None the argument

christianabbet:
Closes #552

# Description
* Fields with only a `not_specified` prediction are no longer returned by `/classify`. They're omitted instead of showing `"not_specified"` or `["not_specified"]`.

* Examples for output prediction were outdated (e.g. "uscs": "not_specified" -> "uscs": "ML"). Fix with current model output.

# Example
Screenshot 2026-09-07 at 15 47 14


stijnvermeeren-swisstopo (commented):
@christianabbet My tests work well, but I wanted to ask Copilot to check the code as well. Could you check the AI's comments (including the surpressed one) to see if they are relevant or can be ignored?

PR #573, opened 2026-09-16
Feat/issue 550/assign borehole name page by page

### LINKED ISSUE #550: Assign borehole name "page-by-page"
From #542 (comment) 

The matching between boreholes and metadata is currently not in the best place, which became clear when we had to implement `_assign_borehole_names` and `_confident_names` for using borehole names in the continuation detection.

It might make sense to always match all of the metadata to the extracted stratigraphy already on a page-by-page basis, before doing any continuation detection, i.e. use more or less the existing logic from the `BoreholeListBuilder` class, but call it for each page separately inside the for `page_index`, `page in enumerate(doc)` loop of the `extract()` method. 

Since we have cases where the metadata is not on the same page as the borehole 
e.g.  borehole name for zurichs A42 and elevation data for A11452.pdf and groundwater information for 8366.pdf 
we could consider a more flexible approach  which allows assigning data from adjacent pages to a borehole, as long as there is no other borehole detected on those adjacent pages.

AgathaSchmidt:
Closes #550 

# Changes 
Metadata matching moved from a single whole-document pass after continuation merge to a per-page pass before the continuation merge
The main goal was to change the existing structure where we had to assigning borehole names for boreholes twice with two different methods. (`_assign_borehole_names` and `_confident_names` )

**extract.py**
- Removed the whole-document collection lists `all_name_entries`/`all_groundwater_entries` (`NameInDocument`/`GroundwaterInDocument`).
- metadata.elevations/metadata.coordinates (still extracted for whole document) are now sliced per page inside the loop.
- Groundwater extraction moved earlier in the loop, before metadata assignment, so the elevation/groundwater overlap guard can use it.
- Calls `assign_page_metadata` per page and collects `boreholeless_pages`, then calls `resolve_boreholeless_pages` once after the loop, before merge_boreholes.
- BoreholeListBuilder(...).build() replaced with `build_borehole_predictions`(...).

**extract_stratigraphy.py** 
- Restructuring: per-page `assign_page_metadata`/`resolve_boreholeless_pages`
- `build_borehole_predictions` wrapper (which called `BoreholeListBuilder`) was deleted in favor of the shared function defined in predictions.py

**layer.py**
 - `ExtractedBorehole` gained elevation, coordinates, and groundwater fields

**continuation_detection.py** 
-  `_merge_boreholes` now also carries elevation, coordinates, and groundwater forward across a merge 

**predictions.py**
- `BoreholeListBuilder` class deleted entirely — > its methods live on as free functions as described below, and `build()` is replaced by the standalone `build_borehole_predictions`.
- `extend_list` —> Moved from a `@staticmethod` on `BoreholeListBuilder` to a module-level function with the exact same body
- delete  `remove_elevations_overlaping_groundwater` and replace with `_remove_elevations_matching_groundwater` —> Same logic, just scoped to one page instead of the whole document

- `_one_to_one_match_element_to_borehole`, `_many_to_one_match_element_to_borehole`: de-classed into a free function taking `boreholes` as an explicit parameter instead of `self`, with the matching algorithm itself unchanged: 
- `compute_distance`: de-classing, plus it now takes the whole `borehole` instead of a pre-extracted `bounding_boxes` list — same distance calculation either way.
- `build_borehole_predictions`: now that matching already happened per page beforehand, it just reshapes each already-matched borehole straight into a `BoreholePredictions`.
- `_assign_borehole_names` : deleted -  Its job (attach the nearest name to each borehole on a page) is now done by `assign_page_metadata`, which reuses the real `one_to_one_match_element_to_borehole` matcher instead of `_assign_borehole_names`'s naive per-borehole `min()`

## new functions
- `resolve_boreholeless_pages`: for pages that had metadata but no borehole to attach it to, tries to give it to a neighboring page's borehole —> only if exactly one of the two neighbors has exactly one borehole. Otherwise leaves it unassigned.
- `assign_page_metadata`: matches one page's name/elevation/coordinate/groundwater candidates to that page's boreholes (one-to-one for name/elevation/coordinate, many-to-one for groundwater) and writes the results onto each `ExtractedBorehole`.

# side effects 

zurich 
691251009.pdf --> one redundant title candidate less 
geouat/val : 
A11452 : both wrong name, but now that a name can also be on page 2 we find one which has at least a better keyword ("Rotationskernbohrung" instead of “Neigung der Bohrung”)


stijnvermeeren-swisstopo (approved):
@AgathaSchmidt Nice work. The borehole name matching now is more intuitive, and the overall code complexity is reduced.

Inspired by your work, I proposed some additional refactoring / simplification in a separate PR: https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/575 . That also removes the need for the `if TYPE_CHECKING:` workaround by moving the groundwater dataclass definition to a separate Python file from the groundwater extraction logic, which seems a cleaner solution to me. The additional changes in my PR can either by merged into this PR already, or reviewed and merged separately, as you wish.

I'm not entirely happy yet with the almost identical / duplicated data structures  ExtractedBorehole and BoreholePredictions. I'd also prefer to have a cleaner way to represent an extracted stratigraphy (with just sidebar and material descriptions, but without any associated metadata/groundwater yet), rather than immediately creating an ExtractedBorehole object with empty values for metadata and groundwater. However, that's already at a pretty high level of nitpicking, and beyond the scope of the issue. So I'm also happy to accept this implementation as a pragmatic solution for now.

PR #574, opened 2026-09-16
Feat/issue 556/memory usage documentation

### LINKED ISSUE #556: Memory usage documentation
Follow up of #529 

Finally, it would be good to document the memory requirements for running the API to the Markdown documentation. In particular, the following estimates would be useful:

   * Memory usage when idle and classification models not loaded?
   * Memory usage when idle and classification models are loaded?
   * How much does the memory usage increase when processing large documents?

AgathaSchmidt:
Closes #566 

Tested memory usage for classification endpoint and classify_borehole_type endpoint 
Ran everything three times and reported median and standard deviation


christianabbet (approved):
LGTM

PR #575, opened 2026-09-17
Assign borehole name per page: additional refactoring

stijnvermeeren-swisstopo:
Additional changes based on https://github.com/swisstopo/swissgeol-boreholes-dataextraction/pull/573

- separate files for dataclass definitions, removing the need to the "TYPE_CHECKING" workaround
- remove redundant LayersInDocument class
- remove redundant borehole_predictions_list param for create_response_object, replacing it with a simple boolean
- remove unused filename from BoreholePredictions class
- combined borehole post-processing method that deals with both inserting line breaks as well as processing groundwater measurements
  - adds line break logic also to the API, where it was previously missing


AgathaSchmidt (approved):
Thanks for the refactor! 
It looks good and I think it should be definitely be merged with my branch before merging it to develop

